PR #166, opened 2026-04-10
fix(ui): local mode Activity panel polish — border off-by-one, P/E gauge alignment, @hostname suppression, ANE row always-on, and Processes header restyle

### LINKED ISSUE #165: fix(ui): local mode Activity panel polish — border off-by-one, P/E gauge alignment, @hostname suppression, ANE row always-on, and Processes header restyle
## Summary

The local-mode Activity panel and device list rendering have accumulated a handful of small but visible issues that together make the local view look broken on Apple Silicon. This issue tracks seven concrete fixes, most of them constrained to `src/ui/activity_panel.rs`, `src/ui/gpu_sparkline_panel.rs`, `src/ui/process_renderer.rs`, and the four device renderers under `src/ui/renderers/`.

The fixes are grouped in one issue because they are all noticed on the same screenshot, touch overlapping files, and are individually small enough that splitting them would create more overhead than value. They can be implemented as one PR (or split into 2–3 follow-up commits if the reviewer prefers).

## Reproduction

Run `./target/release/all-smi view` on an Apple M1 Ultra (16P+4E) / macOS. Observe the top Activity panel (CPU Cores + GPU Metrics), the GPU/CPU/Memory/Disk list rows, and the bottom "Processes:" header.

---

### 1. CPU Cores panel bottom border is off by one → also breaks the GPU Metrics panel

**Where:** `src/ui/activity_panel.rs:210-219` (`draw_panel_bottom_border`).

**Problem:** The top border emits exactly `panel_width` display columns, and every content row emits exactly `panel_width` columns, but the bottom border emits `panel_width + 1` columns.

```
Top border       : "  " + "╭─" + " " + title + " " + dashes + "╮"
                 = 2 + 2 + 1 + title.len() + 1 + (panel_width − title.len() − 7) + 1
                 = panel_width                                     ← correct

Content row      : "  " + "│ " + bar_width + " " + blocks + pad + " │"
                 = panel_width                                     ← correct

Bottom border    : "  " + "╰" + (inner_width + 1)·"─" + "╯"
                 = 2 + 1 + (panel_width − 4 + 1) + 1
                 = panel_width + 1                                 ← 1 CHAR TOO LONG
```

Because `render_combined_activity_panel` writes the pre-formatted left (CPU) lines before the right (GPU) lines on the same physical row, the 1-char overflow on the CPU bottom-border row pushes the GPU Metrics bottom-border row one column to the right. That is why **both** boxes appear broken on the bottom edge in the screenshot — the user originally described this as "the E-CPU section's border is off by one, which also breaks the CPU Core and GPU Metrics borders."

**Fix:**

```rust
// src/ui/activity_panel.rs, inside draw_panel_bottom_border
for _ in 0..inner_width {      // was: 0..(inner_width + 1)
    print_colored_text(stdout, "\u{2500}", Color::Cyan, None, None);
}
```

That gives `2 + 1 + inner_width + 1 = inner_width + 4 = panel_width` — matching the top border and all content rows.

**Regression test to add:** Extend `test_render_activity_panel_pe_cluster` (or add a new `test_activity_panel_border_widths_match`) that strips ANSI escapes and asserts every line of the rendered panel has the same display-column count. This prevents the bug from silently reappearing.

---

### 2. P-CPU and E-CPU gauge widths differ within the same box

**Where:** `src/ui/activity_panel.rs:349-437` (`draw_pe_cluster_bars` + `draw_cluster_line`).

**Problem:** `draw_cluster_line` recomputes `bar_width` per cluster as `content_width − block_section_width − 2`. Because the P-cluster has more cores than the E-cluster (16 vs. 4 on an M1 Ultra), `block_section_width` is much larger for P, so the P-CPU bar ends up ~15 columns shorter than the E-CPU bar. The result is two mismatched gauges stacked in the same box, which is visually confusing — the longer E-CPU bar implies higher capacity, which is the opposite of what's true.

**Fix:** Compute one shared `bar_width` in `draw_pe_cluster_bars` using the **larger** block section (i.e. the cluster with more cores → the shorter bar), then pass that pre-computed `bar_width` to both `draw_cluster_line` calls. Change `draw_cluster_line`'s signature to accept `bar_width: usize` directly instead of recomputing from `content_width` + `cores.len()`.

Sketch:

```rust
fn draw_pe_cluster_bars<W: Write>(
    stdout: &mut W,
    info: &CpuInfo,
    panel_width: usize,
    _full_width: usize,
) {
    let Some(apple) = &info.apple_silicon_info else { return; };
    let content_width = panel_width.saturating_sub(6);

    let p_cores: Vec<&CoreUtilization> = /* existing filter */;
    let e_cores: Vec<&CoreUtilization> = /* existing filter */;

    // Shared bar width, aligned to the cluster with MORE cores (= the
    // larger block section = the shorter bar). This makes the P-CPU and
    // E-CPU gauges visually comparable instead of implying different
    // capacities.
    let p_block_width = p_cores.len() + (p_cores.len() / 4);
    let e_block_width = e_cores.len() + (e_cores.len() / 4);
    let shared_bar_width =
        content_width.saturating_sub(p_block_width.max(e_block_width) + 2);

    draw_cluster_line(stdout, "P-CPU", apple.p_core_utilization,
                      &p_cores, shared_bar_width, panel_width);
    draw_cluster_line(stdout, "E-CPU", apple.e_core_utilization,
                      &e_cores, shared_bar_width, panel_width);
}

fn draw_cluster_line<W: Write>(
    stdout: &mut W,
    label: &str,
    utilization: f64,
    cores: &[&CoreUtilization],
    bar_width: usize,         // NEW: precomputed, shared across clusters
    panel_width: usize,
) {
    // border left → draw_bar(stdout, label, util, 100.0, bar_width, None)
    // → space → per-core block loop → recompute blocks_printed as before
    // → pad = panel_width − (4 + bar_width + 1 + blocks_printed) − 2
    // → " │" border right
}
```

The total line length stays `panel_width`; the extra columns freed by the narrower E-CPU bar simply become extra right-side padding on the E-CPU row. Blocks remain left-aligned next to the bar so they read as an extension of the gauge.

---

### 3. GPU Temp row shows the "Nominal" thermal-pressure badge — remove it

**Where:** `src/ui/gpu_sparkline_panel.rs:246-262` (`build_rows`) and `:398-413` (`thermal_pressure_badge`).

**Problem:** On Apple Silicon, the GPU Temp row in the upper-right GPU Metrics panel is followed by a qualitative NSProcessInfo thermal pressure label ("Nominal" / "Fair" / "Serious" / "Critical"). This is the only row with a trailing badge, and it confuses the column layout (see issue 4 below). It is also not useful — the row already shows a real die temperature in °C since the SMC `flt ` little-endian fix, and `dashboard.rs` only surfaces `thermal_pressure` in remote mode (the `draw_system_view` path), so removing it from the sparkline panel doesn't lose any information the user is actually using.

**Fix:** In `build_rows`, drop the `thermal_badge` computation for the `GPU Temp` row and set `badge: None`. The `thermal_pressure_badge` helper can be deleted entirely if no other caller exists (grep confirms this is the only call site). Also delete the now-dead `test_thermal_pressure_badge_*` tests and remove `"thermal_pressure"` entries from the `make_apple_silicon_state` fixture in the module's tests.

---

### 4. GPU Temp sparkline is visibly shorter than the rows above it

**Where:** `src/ui/gpu_sparkline_panel.rs:331-340` (`draw_sparkline_row`).

**Problem:** `draw_sparkline_row` computes `sparkline_width = content_width − fixed − badge_len`. Because only the GPU Temp row has a badge, its sparkline is `badge_len` columns shorter than the GPU Util and GPU Mem rows above it. The total row width is still `panel_width` (the missing columns are absorbed into `pad`), but the visible sparkline column looks ragged — the right end of the GPU Temp sparkline does not line up with the sparklines above it.

**Fix:** This is fully resolved by fix #3. Once the thermal badge is removed, `badge_len` is 0 for every row and `sparkline_width` becomes identical across rows.

A belt-and-braces improvement (optional, can be a follow-up): make `draw_sparkline_row` compute `sparkline_width` once from the **maximum** badge_len across `rows` and reuse that width for every row. That future-proofs the layout for any new per-row badge without re-introducing the mismatch.

---

### 5. `@ <hostname>` is redundant on GPU / CPU / Host Memory / Disk rows in local mode

**Where:**
- `src/ui/renderers/gpu_renderer.rs` (`print_gpu_info`, ~line 133)
- `src/ui/renderers/cpu_renderer.rs` (`print_cpu_info`, ~line 263)
- `src/ui/renderers/memory_renderer.rs` (`print_memory_info`, ~line 75)
- `src/ui/renderers/storage_renderer.rs` (`print_storage_info`, ~line 95)
- Callers in `src/view/frame_renderer.rs` (`render_local_devices`, `render_remote_devices`, `render_gpu_section`)

**Problem:** In local mode every device row shows `… @ Cube.lo …`, but there's only one node — the host name is already surfaced in the top local-header bar (`Host Cube.local · Apple M1 Ultra …`) and in the `NODE` chassis row. Repeating it four more times on every frame is noisy and wastes ~12 columns per row that could otherwise be used for a longer device name or just left empty to lighten the visual.

**Fix:** Add a `show_hostname: bool` parameter to each of the four `print_*_info` device renderers (NOT to `print_chassis_info` — the user explicitly kept the chassis NODE line in scope of the "@ hostname" hiding request, and the NODE line already uses a different format `NODE <host> Pwr:…` without `@`; leave it alone to keep this issue tightly scoped). When `show_hostname` is `false`, skip the ` @ hostname_display` segment entirely — do not emit even a blank equivalent, so the rest of the line (`Util:`, `VRAM:`, etc.) shifts left to fill the gap.

Also skip looking up `hostname_scroll_offset` / invoking `format_hostname_with_scroll` on the hidden-hostname path, and skip advancing the hostname scroll animation in local mode from `src/view/event_handler.rs` (search for `host_id_scroll_offsets`) — no user-visible scrolling hostname means no reason to burn CPU incrementing the offset.

In `src/view/frame_renderer.rs`, pass `show_hostname: !view_state.is_local_mode` from the `render_local_devices` / `render_gpu_section` / `render_chassis_section` call sites. The remote-mode callers (`render_remote_devices`) keep `show_hostname: true`.

Update the per-module tests to cover both states (`show_hostname = true` retains the `@ hostname` substring in the stripped-ANSI output; `show_hostname = false` does not).

---

### 6. GPU Metrics sparkline panel doesn't show ANE even on Apple Silicon; add room for NPU on Intel

**Where:** `src/ui/gpu_sparkline_panel.rs:71-79` (`gpu_content_rows`) and `:215-300` (`build_rows` / `has_ane_data`).

**Problem:** `has_ane_data` currently returns `false` when `gpu.ane_utilization == 0.0`, so the ANE row is hidden whenever the Neural Engine is idle — which is most of the time on a desktop M1 Ultra, even though the hardware clearly has an ANE. The user wants the ANE row to be **always** present on Apple Silicon so the panel reflects the real hardware layout. (The individual GPU info row in `print_gpu_info` already draws an always-on ANE gauge at `gpu_renderer.rs:311-330`; this fix brings the top GPU Metrics panel into the same behaviour.)

The user also asked to leave room for a similar "NPU" row on Intel CPUs once NPU telemetry lands (Meteor Lake / Core Ultra). This doesn't require real data today — just the scaffolding so that adding an Intel NPU reader later is a one-line UI change.

**Fix:**
1. Rename/repurpose `has_ane_data` to `show_ane_row`, and let it return `detect_apple_silicon(state)` alone (drop the `ane_utilization > 0.0` gate). Add a doc comment explaining that an ANE at 0 W is a meaningful "idle" reading and the row is load-bearing for platform identity even when idle.
2. Update `gpu_content_rows` to include the ANE row whenever `show_ane_row` is true — the row count becomes 5 on Apple Silicon regardless of current ANE power.
3. Add a `show_npu_row(state: &AppState) -> bool` helper right next to `show_ane_row`. For now it returns `false` (no Intel/Windows NPU reader exists yet), and a short doc comment points at `src/api/metrics/npu/common.rs` as the hand-off point for future NPU data. In `build_rows`, add a mirror of the ANE block that emits an "NPU" row (Yellow, same layout) when `show_npu_row` is true, so that enabling it later is literally flipping the helper return value.
4. Update the existing `test_gpu_content_rows_apple_with_ane` expectation from 5→5 (unchanged) and add `test_gpu_content_rows_apple_without_ane_data_still_shows_row` covering `ane_utilization == 0.0`.

As a follow-up (out of scope here), a dedicated `ane_power_history: VecDeque<f64>` on `AppState` would give the ANE sparkline real history instead of the current single-point `vec![ane_w]`. Worth mentioning in the issue body so the reviewer knows the sparkline trivialness is intentional, not forgotten.

---

### 7. Make the `Processes:` header look like the rest of the UI

**Where:** `src/ui/process_renderer.rs:59`.

**Problem:** Every other panel in the TUI uses either a full box (`CPU Cores`, `GPU Metrics`) or a colored horizontal rule. The process section opens with a bare ASCII literal `Processes:` followed directly by an un-labelled column header — it's the one unstyled artefact in the view and stands out as half-finished.

**Fix:** Replace the single `queue!(stdout, Print("Processes:\r\n"))` call with a one-line styled title rule that matches the rest of the UI, for example:

```rust
// Styled title line: "── Processes ───────────────────────"
print_colored_text(stdout, "\u{2500}\u{2500} ", Color::Cyan, None, None);
print_colored_text(stdout, "Processes", Color::Cyan, None, None);
print_colored_text(stdout, " ", Color::DarkGrey, None, None);
let title_prefix_cols = 3 + "Processes".len() + 1; // "── Processes "
let dashes = width.saturating_sub(title_prefix_cols);
print_colored_text(stdout, &"\u{2500}".repeat(dashes), Color::DarkGrey, None, None);
queue!(stdout, Print("\r\n")).unwrap();
```

Keep the line count unchanged (still exactly 1 row) so that `RESERVED_HEADER_ROWS = 4` and the existing `available_rows_for_processes` math continues to work without adjustment. An optional enhancement: append a compact `[N total · M active]` counter at the end of the title rule, formatted in `DarkGrey`, to make the header carry actual information instead of just looking pretty — but that's optional and can be split out.

---

## Acceptance Criteria

- [ ] CPU Cores panel top, bottom, and content rows render at identical column widths (add an alignment test that strips ANSI and compares lengths).
- [ ] GPU Metrics panel top/bottom borders align with the CPU Cores panel top/bottom borders on the same physical row (visual check + combined panel alignment test).
- [ ] P-CPU and E-CPU gauge bars visibly end at the same column on an Apple Silicon screenshot.
- [ ] GPU Temp row in the GPU Metrics panel no longer shows the "Nominal" text.
- [ ] GPU Temp sparkline ends at the same column as GPU Util and GPU Mem sparklines above it.
- [ ] `@ <hostname>` no longer appears on GPU / CPU / Host Memory / Disk rows in local mode; it continues to appear in remote mode.
- [ ] GPU Metrics panel shows an `ANE` row on Apple Silicon regardless of current ANE power (including when power = 0 W).
- [ ] `show_npu_row` helper exists, returns `false`, and is invoked from `build_rows` so that flipping it to `true` later yields an NPU row with no additional code changes.
- [ ] Processes section opens with a styled title rule, not a bare `Processes:`.
- [ ] `cargo fmt --check`, `cargo clippy`, and `cargo test` all pass.

## Non-goals

- No changes to the `NODE` chassis line (keep chassis hostname display as-is).
- No new `ane_power_history` on `AppState` — the ANE sparkline stays single-point for this issue.
- No actual Intel NPU reader implementation — only UI scaffolding.
- No changes to remote-mode rendering beyond what naturally falls out of the `show_hostname: bool` parameter plumbing.

inureyes:
## Summary

Fixes #165 — seven small but visible issues in the local-mode Activity panel and device list rendering on Apple Silicon.

- **Border off-by-one**: Fix CPU Cores panel bottom border emitting `panel_width + 1` columns instead of `panel_width`, which also broke the GPU Metrics panel alignment on the same row.
- **P/E gauge alignment**: Compute one shared `bar_width` in `draw_pe_cluster_bars` using the larger block section so P-CPU and E-CPU gauges end at the same column.
- **Thermal badge removal**: Drop the "Nominal" thermal pressure badge from the GPU Temp sparkline row. This also fixes the GPU Temp sparkline being shorter than the rows above it.
- **Hostname suppression**: Add `show_hostname: bool` parameter to `print_gpu_info`, `print_cpu_info`, `print_memory_info`, and `print_storage_info`. Skip `@ hostname` in local mode. Also skip hostname scroll animation in local mode.
- **ANE always-on**: Replace `has_ane_data` (gated on `ane_utilization > 0`) with `show_ane_row` returning true whenever Apple Silicon is detected. Add `show_npu_row` scaffolding for future Intel NPU support.
- **Processes header restyle**: Replace bare `Processes:` with styled title rule `── Processes ──────────`.

## Test plan

- [x] `cargo fmt --check` passes
- [x] `cargo clippy` passes with no warnings
- [x] `cargo test` passes (325 tests, 0 failures)
- [ ] Visual check: CPU Cores panel borders align top to bottom
- [ ] Visual check: P-CPU and E-CPU gauges end at the same column
- [ ] Visual check: GPU Temp sparkline same length as GPU Util/GPU Mem rows
- [ ] Visual check: `@ hostname` not shown on GPU/CPU/Memory/Disk rows in local mode
- [ ] Visual check: ANE row present in GPU Metrics panel even when idle
- [ ] Visual check: Processes section opens with styled title rule

### DIFF
--- a/src/ui/activity_panel.rs
+++ b/src/ui/activity_panel.rs
@@ -211,7 +211,7 @@ fn draw_panel_bottom_border<W: Write>(stdout: &mut W, panel_width: usize, _full_
     let inner_width = panel_width.saturating_sub(4); // 2 margin + 2 corners
     print_colored_text(stdout, "  ", Color::White, None, None);
     print_colored_text(stdout, "\u{2570}", Color::Cyan, None, None);
-    for _ in 0..(inner_width + 1) {
+    for _ in 0..inner_width {
         print_colored_text(stdout, "\u{2500}", Color::Cyan, None, None);
     }
     print_colored_text(stdout, "\u{256f}", Color::Cyan, None, None);
@@ -371,13 +371,19 @@ fn draw_pe_cluster_bars<W: Write>(
         .filter(|c| c.core_type == CoreType::Efficiency)
         .collect();
 
+    // Compute one shared bar_width using the larger block section so that
+    // P-CPU and E-CPU gauges end at the same column.
+    let p_block_width = p_cores.len() + (p_cores.len() / 4);
+    let e_block_width = e_cores.len() + (e_cores.len() / 4);
+    let shared_bar_width = content_width.saturating_sub(p_block_width.max(e_block_width) + 2);
+
     // P-cluster line: bar + utilization blocks
     draw_cluster_line(
         stdout,
         "P-CPU",
         apple.p_core_utilization,
         &p_cores,
-        content_width,
+        shared_bar_width,
         panel_width,
     );
 
@@ -387,7 +393,7 @@ fn draw_pe_cluster_bars<W: Write>(
         "E-CPU",
         apple.e_core_utilization,
         &e_cores,
-        content_width,
+        shared_bar_width,
         panel_width,
     );
 }
@@ -397,17 +403,13 @@ fn draw_cluster_line<W: Write>(
     label: &str,
     utilization: f64,
     cores: &[&CoreUtilization],
-    content_width: usize,
+    bar_width: usize,
     panel_width: usize,
 ) {
     print_colored_text(stdout, "  ", Color::White, None, None);
     print_colored_text(stdout, "\u{2502} ", Color::Cyan, None, None);
 
-    // Calculate how much space the utilization blocks need
-    let block_section_width = cores.len() + (cores.len() / 4); // blocks + group separators
-    let bar_width = content_width.saturating_sub(block_section_width + 2); // 2 for spacing
-
-    // Draw the progress bar
+    // Draw the progress bar using the pre-computed shared bar_width
     draw_bar(stdout, label, utilization, 100.0, bar_width, None);
     print_colored_text(stdout, " ", Color::White, None, None);
 
--- a/src/ui/gpu_sparkline_panel.rs
+++ b/src/ui/gpu_sparkline_panel.rs
@@ -409,7 +409,6 @@ fn show_ane_row(state: &AppState) -> bool {
 /// Currently returns `false` -- no Intel/Windows NPU reader exists yet.
 /// When an NPU telemetry reader is added (Meteor Lake / Core Ultra),
 /// flip this to check for NPU presence via `src/api/metrics/npu/common.rs`.
-#[allow(clippy::unnecessary_wraps)]
 fn show_npu_row(_state: &AppState) -> bool {
     false
 }
--- a/src/ui/gpu_sparkline_panel.rs
+++ b/src/ui/gpu_sparkline_panel.rs
@@ -70,10 +70,10 @@ pub fn gpu_content_rows(state: &AppState) -> usize {
     if state.gpu_info.is_empty() {
         return 0;
     }
-    let is_apple = detect_apple_silicon(state);
-    let has_ane = is_apple && has_ane_data(state);
-    // GPU Util + GPU Mem + GPU Temp + (ANE?) + Pkg Power
-    3 + usize::from(has_ane) + 1
+    let ane = show_ane_row(state);
+    let npu = show_npu_row(state);
+    // GPU Util + GPU Mem + GPU Temp + (ANE?) + (NPU?) + Pkg Power
+    3 + usize::from(ane) + usize::from(npu) + 1
 }
 
 /// Render the combined Activity panel (CPU left half + GPU right half).
@@ -156,8 +156,9 @@ fn render_gpu_lines(state: &AppState, panel_width: usize) -> Vec<String> {
     }
 
     let is_apple = detect_apple_silicon(state);
-    let has_ane = is_apple && has_ane_data(state);
-    let rows = build_rows(state, is_apple, has_ane);
+    let ane = show_ane_row(state);
+    let npu = show_npu_row(state);
+    let rows = build_rows(state, is_apple, ane, npu);
 
     let mut lines: Vec<String> = Vec::with_capacity(rows.len() + 2);
 
@@ -210,8 +211,8 @@ struct SparklineRow {
 // Row construction
 // ---------------------------------------------------------------------------
 
-fn build_rows(state: &AppState, is_apple: bool, has_ane: bool) -> Vec<SparklineRow> {
-    let mut rows = Vec::with_capacity(5);
+fn build_rows(state: &AppState, is_apple: bool, has_ane: bool, has_npu: bool) -> Vec<SparklineRow> {
+    let mut rows = Vec::with_capacity(6);
 
     // 1. GPU Utilization
     // Collect once; clone for power-proxy row to avoid a second VecDeque iteration.
@@ -254,7 +255,7 @@ fn build_rows(state: &AppState, is_apple: bool, has_ane: bool) -> Vec<SparklineR
         badge: None,
     });
 
-    // 4. ANE (Apple Silicon only, when data is present)
+    // 4. ANE (Apple Silicon -- always shown regardless of current power)
     if has_ane {
         let ane_mw = state
             .gpu_info
@@ -270,6 +271,19 @@ fn build_rows(state: &AppState, is_apple: bool, has_ane: bool) -> Vec<SparklineR
             latest_str: format!("{ane_w:.1}W"),
             min_max_str: String::new(),
             history: ane_history,
+            range: None,
+            badge: None,
+        });
+    }
+
+    // 4b. NPU (Intel/Windows -- scaffolding for future NPU reader)
+    if has_npu {
+        rows.push(SparklineRow {
+            label: "NPU",
+            color: Color::Yellow,
+            latest_str: "0.0W".to_string(),
+            min_max_str: String::new(),
+            history: vec![0.0],
             range: None,
             badge: None,
         });
@@ -380,12 +394,24 @@ fn detect_apple_silicon(state: &AppState) -> bool {
     })
 }
 
-fn has_ane_data(state: &AppState) -> bool {
-    state
-        .gpu_info
-        .first()
-        .map(|g| g.ane_utilization > 0.0)
-        .unwrap_or(false)
+/// Whether the ANE row should be shown in the GPU Metrics panel.
+///
+/// Returns `true` on Apple Silicon regardless of current ANE power.
+/// An ANE at 0 W is a meaningful "idle" reading and the row is
+/// load-bearing for platform identity even when the Neural Engine
+/// is completely idle.
+fn show_ane_row(state: &AppState) -> bool {
+    detect_apple_silicon(state)
+}
+
+/// Whether an NPU row should be shown in the GPU Metrics panel.
+///
+/// Currently returns `false` -- no Intel/Windows NPU reader exists yet.
+/// When an NPU telemetry reader is added (Meteor Lake / Core Ultra),
+/// flip this to check for NPU presence via `src/api/metrics/npu/common.rs`.
+#[allow(clippy::unnecessary_wraps)]
+fn show_npu_row(_state: &AppState) -> bool {
+    false
 }
 
 fn package_power(state: &AppState, is_apple: bool) -> (f64, &'static str) {
@@ -631,9 +657,31 @@ mod tests {
     }
 
     #[test]
-    fn test_has_ane_data() {
-        assert!(!has_ane_data(&make_nvidia_state()));
-        assert!(has_ane_data(&make_apple_silicon_state()));
+    fn test_show_ane_row() {
+        assert!(!show_ane_row(&make_nvidia_state()));
+        assert!(show_ane_row(&make_apple_silicon_state()));
+    }
+
+    #[test]
+    fn test_show_ane_row_even_when_ane_idle() {
+        // ANE row should be shown even when ane_utilization is 0
+        let mut state = make_apple_silicon_state();
+        state.gpu_info[0].ane_utilization = 0.0;
+        assert!(show_ane_row(&state));
+    }
+
+    #[test]
+    fn test_show_npu_row_returns_false() {
+        assert!(!show_npu_row(&make_nvidia_state()));
+        assert!(!show_npu_row(&make_apple_silicon_state()));
+    }
+
+    #[test]
+    fn test_gpu_content_rows_apple_with_zero_ane_still_shows_row() {
+        let mut state = make_apple_silicon_state();
+        state.gpu_info[0].ane_utilization = 0.0;
+        // GPU Util + GPU Mem + GPU Temp + ANE (always-on) + Pkg Power = 5
+        assert_eq!(gpu_content_rows(&state), 5);
     }
 
     #[test]
@@ -667,7 +715,7 @@ mod tests {
     #[test]
     fn test_build_rows_nvidia() {
         let state = make_nvidia_state();
-        let rows = build_rows(&state, false, false);
+        let rows = build_rows(&state, false, false, false);
         assert_eq!(rows.len(), 4);
         assert_eq!(rows[0].label, "GPU Util");
         assert_eq!(rows[1].label, "GPU Mem");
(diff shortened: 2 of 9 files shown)


inureyes:

### Intent
> Fix seven visual polish issues in the local-mode Activity panel and device list rendering on Apple Silicon: border off-by-one, P/E gauge alignment, thermal badge removal, hostname suppression, ANE always-on, NPU scaffolding, and Processes header restyle.

### Verification
- [x] Tests pass (325 unit + 385 integration, 0 failures)

### Detailed findings

All seven issue requirements are correctly and completely implemented:

**Fix 1 - Border off-by-one** (`activity_panel.rs:214`): Changed `0..(inner_width + 1)` to `0..inner_width`, making the bottom border emit `panel_width` columns matching top border and content rows. Correct.

**Fix 2 - P/E gauge shared bar_width** (`activity_panel.rs:376-378`): Pre-computes `shared_bar_width` using `p_block_width.max(e_block_width)`, then passes it to both `draw_cluster_line` calls. The function signature was updated to accept `bar_width: usize` directly. Correct.

**Fix 3 - Thermal badge removal** (`gpu_sparkline_panel.rs`): Removed `thermal_pressure_badge` helper, `thermal_badge` computation in `build_rows`, and `thermal_pressure` from test fixtures. All `badge` fields set to `None`. GPU Temp row now has `badge: None`. Correct.

**Fix 4 - GPU Temp sparkline width** (resolved by Fix 3): With `badge_len = 0` for all rows, `sparkline_width` is now identical across GPU Util, GPU Mem, and GPU Temp. Correct.

**Fix 5 - Hostname suppression** (4 renderers + `frame_renderer.rs` + `ui_loop.rs`): Added `show_hostname: bool` to `print_gpu_info`, `print_cpu_info`, `print_memory_info`, `print_storage_info`. All call sites pass `true` in remote mode and `false` in local mode. `format_hostname_with_scroll` is properly gated inside the conditional. Hostname scroll animation in `ui_loop.rs` is skipped in local mode for both GPU and CPU hostname updates. Chassis renderer is untouched per spec. Correct.

**Fix 6 - ANE always-on + NPU scaffolding** (`gpu_sparkline_panel.rs`): `has_ane_data` renamed to `show_ane_row`, now returns `detect_apple_silicon(state)` (no power gate). `show_npu_row` added, returns `false` with doc comment pointing to `src/api/metrics/npu/common.rs`. Both are wired into `gpu_content_rows` and `build_rows`. Tests cover ANE-idle case and NPU scaffolding. Correct.

**Fix 7 - Processes header restyle** (`process_renderer.rs:62-74`): Replaced bare `"Processes:\r\n"` with styled title rule using Cyan for prefix/title and DarkGrey for trailing dashes. Still emits exactly 1 row, so `RESERVED_HEADER_ROWS = 4` is unchanged. Correct.

### Minor observations (LOW, not blocking)
- `#[allow(clippy::unnecessary_wraps)]` on `show_npu_row` is a no-op since the function returns `bool`, not `Option`/`Result`. Harmless but unnecessary.
- Comment at `process_renderer.rs:194` still says `"Processes:" title` -- could say "styled title rule" to match the new implementation. Purely cosmetic.

### Remaining Items
None -- all seven fixes are correctly implemented with no issues found.

inureyes:
## Security & Performance Review -- PR #166

### Scope
Reviewed all 9 changed files (171 additions, 111 deletions) for security vulnerabilities, performance regressions, and code quality issues.

### Findings

**FIXED -- LOW: Dead clippy allow attribute (commit 3759566)**

`#[allow(clippy::unnecessary_wraps)]` on `show_npu_row` in `src/ui/gpu_sparkline_panel.rs` was a no-op. The `unnecessary_wraps` lint only fires on functions returning `Option` or `Result`, not `bool`. The attribute was misleading about the function's semantics. Removed.

**INFO -- LOW: Minor block-width formula inconsistency**

In `src/ui/activity_panel.rs`, the `p_block_width` formula at line 376 (`cores.len() + cores.len() / 4`) does not subtract 1 for the trailing-group-of-4 case, while the `blocks_printed` formula at line 426 does. This means `shared_bar_width` is 1 column shorter than necessary when the core count is a multiple of 4. The lost column becomes right-side padding, so the visual output is correct and panel width is maintained. No fix needed -- the inconsistency is cosmetic (1 pixel of bar) and the conservative direction (shorter bar, not overflow).

**INFO -- Pre-existing: Byte-index slice on Unicode header**

`src/ui/process_renderer.rs:174` uses `header_format[horizontal_scroll_offset..]` which is a byte-index slice. The header contains multi-byte Unicode sort arrows (3 bytes each). If `horizontal_scroll_offset` lands mid-character, this would panic. This is pre-existing code not introduced by this PR, but worth tracking for a future hardening pass.

### Verified Clean

- `cargo test` -- all 325 tests pass (0 failures)
- `cargo clippy -- -D warnings` -- no warnings
- `cargo fmt --check` -- clean
- All `show_hostname` call sites in `frame_renderer.rs` are consistent: `true` for remote-mode renderers, `false` (or `!view_state.is_local_mode`) for local-mode renderers
- `saturating_sub` is used correctly throughout for width calculations -- no integer underflow risk
- No new `unwrap()` calls on fallible data paths
- No memory leaks: hostname scroll animation correctly skipped in local mode (no growing offset map entries)
- No security concerns: pure UI rendering code with no user input parsing, no file I/O, no network access

### Overall Assessment

This PR is well-written. The changes are straightforward UI polish with proper use of `saturating_sub` for safe arithmetic, correct gating of hostname animations to avoid wasted work, and clean separation of local/remote rendering paths. One dead attribute was removed.

PR #171, opened 2026-04-10
fix(ui): stabilize local header metric row width to prevent jitter (#170)

### LINKED ISSUE #170: fix(ui): stabilize local header metric row width to prevent left-right jitter
## Problem / Background

In local view, the top metrics row under the host summary bar shifts left and right as live values change.

The affected row is the compact local-mode summary line that renders:

- `CPU`
- `GPU`
- `RAM`
- `Pwr`
- `Tmp`

When one metric crosses a digit boundary, neighboring segments move. Typical examples:

- `9.9% -> 10.0%`
- `99.9% -> 100.0%`
- `99°C -> 100°C`
- `9.9W -> 10.0W`
- RAM used / total strings changing visible width

This makes the local header look unstable even though the rest of the TUI is stationary.

## Suspected Code Location

- `src/ui/local_header.rs`
- especially the formatting path around `draw_metrics_line()` and the per-metric helpers

## Expected Behavior

The local header metrics row should have stable horizontal layout across refreshes.

- Each metric segment should reserve enough width for its value field.
- Braille sparklines should stay aligned frame-to-frame.
- One metric growing from 2 digits to 3 digits should not push the following metrics sideways.

## Proposed Fix

Use fixed-width or explicitly padded value fields for each metric segment in the local header.

Examples:

- CPU/GPU percentage fields should reserve width for `100.0%`
- temperature should reserve width for `100°C` or a similarly safe upper bound
- power should reserve width for a stable watt display field
- RAM should reserve a stable used/total field width, or split/pad the numeric portion so the segment width is constant

If needed, compute display widths once and right-pad or left-pad before writing colored text.

## Acceptance Criteria

- [ ] In local mode, the `CPU/GPU/RAM/Pwr/Tmp` row stays horizontally stable during refreshes.
- [ ] Transitions like `9.9 -> 10.0`, `99.9 -> 100.0`, and `99 -> 100` do not move neighboring segments.
- [ ] Braille sparklines remain aligned across refreshes.
- [ ] No regression in color rendering or metric content.
- [ ] `cargo test` and `cargo clippy` pass.

## Notes

This is specifically about the local-mode compact summary header introduced by the local TUI redesign, not the lower Activity panel.

inureyes:
## Summary

- Right-align CPU/GPU percentage values in a 5-char numeric field → consistent 6-column string (`"  0.0%"` through `"100.0%"`)
- Right-align temperature values in a 3-char numeric field → consistent 5-column string (`"  0°C"` through `"999°C"`)
- Right-align power values in a 5-char numeric field → consistent 6-char string (`"  0.0W"` through `"999.9W"`)
- Pad the RAM `used` portion to the same digit-width as `total` so the separator `/` stays in a fixed column (e.g. always `" 0/16GB"` through `"16/16GB"`)

## Test plan

- [ ] `cargo test -p all-smi` passes — updated existing assertions and added `test_format_pct_fixed_width` and `test_format_temp_fixed_display_width` to assert stable byte-lengths across digit boundaries
- [ ] `cargo clippy` clean — no warnings
- [ ] `cargo fmt --check` clean — formatting unchanged
- [ ] Visual verification: run `./target/release/all-smi view` in local mode and confirm the metrics row does not shift left/right as values cross 9→10 or 99→100 boundaries

Closes #170

### DIFF
--- a/src/ui/local_header.rs
+++ b/src/ui/local_header.rs
@@ -439,8 +439,9 @@ mod tests {
         for total in totals {
             let total_str = format!("{total:.0}");
             let w = total_str.len();
-            let len_0 = format!("{:>w$.0}/{total_str}GB", 0.0_f64).len();
-            let len_total = format!("{:>w$.0}/{total_str}GB", total).len();
+            let zero = 0.0_f64;
+            let len_0 = format!("{zero:>w$.0}/{total_str}GB").len();
+            let len_total = format!("{total:>w$.0}/{total_str}GB").len();
             assert_eq!(
                 len_0, len_total,
                 "RAM format width should be stable for total={total}"
--- a/src/ui/local_header.rs
+++ b/src/ui/local_header.rs
@@ -204,7 +204,8 @@ fn draw_ram_sparkline<W: Write>(stdout: &mut W, state: &AppState) {
     let used_gb = state.memory_info.iter().map(|m| m.used_bytes).sum::<u64>() as f64
         / (1024.0 * 1024.0 * 1024.0);
 
-    let value_str = format!("{used_gb:.0}/{total_gb:.0}GB");
+    let total_str = format!("{total_gb:.0}");
+    let value_str = format!("{used_gb:>width$.0}/{total_str}GB", width = total_str.len());
 
     let history: Vec<f64> = state.system_memory_history.iter().copied().collect();
 
@@ -250,7 +251,7 @@ fn draw_power_sparkline<W: Write>(stdout: &mut W, state: &AppState) {
         state.gpu_info.iter().map(|g| g.power_consumption).sum()
     };
 
-    let value_str = format!("{power_watts:.1}W");
+    let value_str = format!("{power_watts:>5.1}W");
 
     let history: Vec<f64> = state.package_power_history.iter().copied().collect();
     let range = if history.is_empty() {
@@ -273,18 +274,24 @@ fn draw_power_sparkline<W: Write>(stdout: &mut W, state: &AppState) {
 // ─── Formatting helpers ───────────────────────────────────────────────────────
 
 /// Format a `%` value as `"<val>%"` or `"N/A"` when missing.
+///
+/// The numeric part is right-aligned in a 5-char field, producing a
+/// consistent 6-display-column string: `"  0.0%"` through `"100.0%"`.
 fn format_pct(value: Option<f64>) -> String {
     match value {
-        Some(v) => format!("{v:.1}%"),
-        None => "N/A".to_string(),
+        Some(v) => format!("{v:>5.1}%"),
+        None => format!("{:>6}", "N/A"),
     }
 }
 
 /// Format a temperature value as `"<val>°C"` or `"N/A"`.
+///
+/// The numeric part is right-aligned in a 3-char field, producing a
+/// consistent 5-display-column string: `"  0°C"` through `"999°C"`.
 fn format_temp(value: Option<f64>) -> String {
     match value {
-        Some(v) => format!("{v:.0}°C"),
-        None => "N/A".to_string(),
+        Some(v) => format!("{v:>3.0}°C"),
+        None => format!("{:>5}", "N/A"),
     }
 }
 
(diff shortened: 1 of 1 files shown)


inureyes:
## Security & Performance Review

**Reviewer**: Automated analysis
**Scope**: `src/ui/local_header.rs` (40 additions, 12 deletions)
**Result**: No issues found

### Security Analysis

No security concerns. This PR modifies only format strings for TUI display output. There are no changes to:

- External input parsing or user-supplied data handling
- Network I/O, file I/O, or system calls
- Authentication, authorization, or credential handling
- Command execution or shell interaction
- Dependencies or unsafe code

The formatting functions (`format_pct`, `format_temp`, power formatting) operate on hardware-sourced numeric values (`f64`, `u32`) that are never user-controllable. The output goes to a terminal buffer (`crossterm::queue!`), not to any persistent storage or network endpoint.

### Performance Analysis

No performance concerns. The changes are strictly cosmetic adjustments to `format!()` macro calls:

- **format_pct**: One `format!` call per render tick per metric (CPU%, GPU%). The addition of `{v:>5.1}` right-alignment padding has negligible cost -- `std::fmt` handles width padding internally without additional allocation beyond what the original `{v:.1}` already required.

- **format_temp**: Same analysis -- `{v:>3.0}` vs `{v:.0}` is equivalent in cost.

- **draw_power_sparkline**: `{power_watts:>5.1}W` -- same analysis.

- **draw_ram_sparkline**: The new code computes `total_str` once and uses its `.len()` for the dynamic width. This adds one extra `format!` call for `total_str`, but it is a trivial integer-to-string conversion that occurs once per render tick. No measurable impact.

None of these paths are hot loops. They execute once per TUI refresh cycle (typically every 1-6 seconds depending on configuration).

### Edge Case Note

The fixed-width fields will overflow their padding if values exceed the field width (e.g., power over 999.9W, temperature over 999C). This is acceptable because:

1. These are **local machine** metrics, not aggregated cluster values
2. No real hardware produces temperatures above 999C or local power above 999.9W
3. Rust's `format!` gracefully expands the field (no truncation, no panic), so the only consequence would be the same jitter the PR is fixing -- which only manifests for physically impossible values

### Tests

All 12 `local_header` tests pass, including the two new tests (`test_format_pct_fixed_width`, `test_format_temp_fixed_display_width`) that assert stable display widths across digit boundaries. Clippy and `cargo fmt --check` are clean.

### Verdict

Approved -- no changes needed.

inureyes:

### Intent
> Stabilize the local-mode header metric row width by using fixed-width formatted fields for CPU%, GPU%, temperature, power, and RAM values, preventing horizontal jitter when values cross digit boundaries.

### Findings Addressed
None required -- the implementation is correct and complete as submitted.

### Remaining Items
None.

### Verification
- [x] Tests pass (all 732 tests pass, 0 failures)

### Review Details

**Correctness**: All four formatting changes produce stable display widths across digit boundaries:
- `format_pct()`: right-aligned 5-char numeric + `%` = 6 display columns for all values 0.0 through 100.0. The N/A case is also padded to 6 columns.
- `format_temp()`: right-aligned 3-char numeric + `°C` = 5 display columns for all values 0 through 999. The `°` character is 2 bytes in UTF-8 but 1 display column, and the test correctly uses exact string comparison rather than byte-length comparison to verify.
- Power: right-aligned 5-char numeric + `W` = 6 display columns for values 0.0 through 999.9.
- RAM: used value padded to match total value digit width, keeping the `/` separator fixed.

**Test coverage**: Two new tests (`test_format_pct_fixed_width` and `test_format_temp_fixed_display_width`) specifically validate width stability across digit boundaries. Existing tests updated with correct expected values. All 12 local_header tests pass.

**Convention consistency**: The right-aligned fixed-width format specifiers (`{:>N.M}`) are consistent with the existing patterns used throughout the UI code (e.g., `widgets.rs`, `chassis_renderer.rs`, `memory_renderer.rs`).

**Edge case note**: Power values above 999.9W would exceed the 5-char field width, but this is acceptable for the local header context (single-machine monitoring). The dashboard mode handles multi-kilowatt aggregation separately with kW formatting.

**Quality checks**: `cargo clippy` clean, `cargo fmt --check` clean.

inureyes:
## PR Finalization Complete

### Summary
- **Tests**: Added 2 new tests — `test_format_power_fixed_width` and `test_format_ram_fixed_separator_position` — covering the inline formatting logic in `draw_power_sparkline` and `draw_ram_sparkline`. The RAM test also caught a subtle point: `used` is padded to the digit-width of `total` (not to a fixed constant), so `" 0/16GB"` rather than `"0/16GB"`. All 14 module tests pass.
- **Documentation**: No changes needed — existing doc comments on the formatting helpers and module-level doc accurately describe the behavior.
- **Lint/Format**: Fixed one `clippy::uninlined_format_args` violation introduced by the new tests (format argument inlined per Rust 1.58+ style).

PR #197, opened 2026-04-20
feat: metric recording ('all-smi record') and TUI replay ('view --replay')

### LINKED ISSUE #187: feat: metric recording ('all-smi record') and TUI replay ('view --replay')
## Summary

Add the ability to persistently capture a metric stream to a file (`all-smi record`) and replay it later in the TUI (`all-smi view --replay file.ndjson`). Replay supports play/pause, step, seek, and speed controls so operators can investigate past incidents without a Prometheus instance or full observability stack.

## Motivation

When an incident happens on a cluster, the current `all-smi` modes cannot reproduce state post-hoc: `local`/`view` are live only, and `api` requires a running Prometheus to retain history. Operators and researchers regularly want to say "rewind to 14:32, that's when throughput cratered" and see the exact same TUI they would have seen live. A compact self-contained recording format unlocks this without the operational overhead of a full monitoring stack.

## Current state

- `src/view/render_snapshot.rs` already materializes a full snapshot per frame — a natural structural fit for a per-frame record.
- Data types (`GpuInfo`, `CpuInfo`, `ChassisInfo`, `MemoryInfo`, process list, …) all implement `Serialize + Deserialize`.
- No persistence today.
- `src/view/data_collection/strategy.rs` already abstracts local vs remote collection — a clean place to add a `ReplayStrategy`.

### Record subcommand

```
all-smi record [--output path.ndjson | path.ndjson.zst | path.ndjson.gz]
               [--interval <secs>]
               [--duration 1h | 0]
               [--source local|remote]
               [--hosts ...] [--hostfile ...]
               [--include gpu,cpu,memory,chassis,process]
               [--max-size 100M] [--max-files 10]
               [--compress zstd|gzip|none]
```

Behavior:

- Source selection mirrors `view` (local reader vs remote HTTP scrape), so recording a remote cluster works too.
- Writes one NDJSON line per collection cycle. Each line is a self-contained JSON object with `{ "schema": 1, "t": "2026-04-18T...Z", "gpus": [...], "cpus": [...], ... }` — same shape as the `snapshot` subcommand's JSON output, sharing serialization code.
- Rotation on size: when output exceeds `--max-size`, close, rename to `path.0001.ndjson[.zst]`, open a new file. Keep at most `--max-files` segments.
- Compression: auto-detect from extension, or explicit `--compress`. Prefer zstd default.
- `--duration 0` means record until `SIGTERM`; on signal, flush and close cleanly.
- On error reading a device, emit `{ "t": ..., "errors": ["nvidia: NotSupported", ...] }` lines so the replay shows gaps instead of silently skipping.

### Replay

Extend `all-smi view` with:

```
all-smi view --replay path.ndjson[.zst|.gz]
             [--speed 1.0] [--start 00:30] [--loop]
```

- Entering replay mode the TUI status bar shows `REPLAY | 00:12:34 / 01:00:00 | 2.0x | paused`.
- New keybindings (active only in replay mode):
  - `SPACE` — play/pause
  - `]` — step one frame forward, `[` — step one frame back
  - `+` / `-` — speed up/down (0.25x, 0.5x, 1x, 2x, 4x, 8x)
  - `j` / `k` — seek backward/forward by 10s
  - `g` — open timecode entry, type `HH:MM:SS` + Enter to jump
  - `L` — toggle looping
- The replay strategy feeds the same `RenderSnapshot` pipeline as live data — no renderer code should change behavior based on mode, only the status bar shows the replay chip.

### File format

- NDJSON (one JSON object per line) so the file can be truncated, appended, or `tail -f`'d safely.
- Frame shape (v1 schema):
  ```json
  {"schema":1,"t":"2026-04-18T12:00:00.000Z","seq":42,"source":"local","gpus":[...],"cpus":[...],"memory":{...},"chassis":{...},"processes":[...],"errors":[]}
  ```
- Header frame (optional first line): `{"schema":1,"header":true,"interval_ms":3000,"hosts":["dgx-01","dgx-02"],"all_smi_version":"0.21.0"}`.
- A sparse index frame every 1000 data frames: `{"schema":1,"index":true,"seq":42000,"byte_offset":1234567}` to enable efficient seeking into compressed files.

## Implementation plan

Files to add / modify:

- `src/cli.rs` — `Record(RecordArgs)`; extend `ViewArgs` with `replay: Option<PathBuf>`, `speed: f32`, `start: Option<String>`, `replay_loop: bool`.
- New `src/record/mod.rs`:
  - `Recorder` struct with `start(source, options) -> Result<()>`.
  - Uses `tokio::io::BufWriter` atop an adapter that picks `zstd::stream::write::Encoder`, `flate2::write::GzEncoder`, or raw `File` based on extension.
  - Signal handler: on `SIGTERM`/`SIGINT`, flush, close, exit 0.
  - Rotation helper.
- New `src/record/replay.rs`:
  - `Replayer` struct: streams frames, maintains cursor, exposes `current()`, `next()`, `prev()`, `seek(Duration)`.
  - Uses the `index` frames for fast seek; if absent, scans from nearest known position with a bounded-memory sliding cache.
- `src/view/data_collection/strategy.rs` — add `ReplayStrategy` variant alongside local/remote.
- `src/view/runner.rs` — when `replay.is_some()`, wire `ReplayStrategy`.
- `src/view/event_handler.rs` — mode-aware key handling; in replay mode, add the SPACE/`]`/`[`/`+`/`-`/`j`/`k`/`g`/`L` keys.
- `src/ui/chrome.rs` or `src/ui/local_header.rs` — add the `REPLAY | ts / total | speed | state` indicator.
- Re-export a shared `pub fn write_frame_json(w, snapshot) -> io::Result<()>` so `record` and `snapshot` cannot drift.

Dependencies to add:
- `zstd = "0.13"` (feature-gated behind `record` if binary size is a concern, else default on).
- `flate2 = "1"` (gzip for interoperability).

## Acceptance criteria

- [x] `all-smi record -o /tmp/trace.ndjson --duration 30s` produces a valid NDJSON file with at least 10 frames (at 3s default interval).
- [x] `all-smi record -o /tmp/trace.ndjson.zst --duration 30s` produces a valid zstd-compressed file (`zstd -d` decompresses to NDJSON).
- [x] Rotation kicks in: `--max-size 1K --max-files 3` yields up to 3 segments, oldest evicted.
- [x] `SIGTERM` during recording closes cleanly with a complete final JSON line.
- [x] `all-smi view --replay /tmp/trace.ndjson` opens the TUI and shows the first frame; the status bar shows REPLAY and ts/total.
- [x] `SPACE` toggles play; frames advance at 1x by default; `+`/`-` scale speed.
- [x] `]`/`[` step one frame.
- [x] `g 00:00:15 <Enter>` jumps to 15 seconds in.
- [x] `--loop` replays indefinitely.
- [x] Running `view --replay` against a `view --hostfile` recording renders the same tab structure (multiple hosts), not just a single local view.
- [x] `cargo test` includes: round-trip (record → replay), schema version mismatch error path, compressed file handling, seek correctness with and without index frames.
- [x] README gains "Recording & Replay" section with usage.

## Edge cases & non-goals

- Schema version mismatch must produce a clear error (`replay: unsupported schema version 2, this all-smi supports schema 1`), not a panic.
- Corrupted tail line (incomplete frame written on crash): skip with a warning, continue.
- Large files (>1 GiB) must stream — do not load the whole file. Benchmark with a 1M-frame file.
- The replay pipeline MUST NOT call any device readers — ensures replay works on a totally different machine than the one that recorded.
- Remote recordings of a 100-node cluster produce large frames; the writer must buffer and flush periodically.
- Non-goal: editing / trimming recordings — users can `jq -c` filter externally.
- Non-goal: converting to PromQL history — that's a different tool.

## Soft dependency

- Shares JSON schema with the `snapshot` subcommand and the SSE streaming endpoint — land schema first, these three features should use the identical serializer.

inureyes:
## Summary

Adds the `all-smi record` subcommand and extends `all-smi view` with
`--replay <file>` so operators can capture a metric stream to disk and
scrub/step/seek through it later in the exact same TUI they would have
seen live. Intended for post-hoc incident investigation without a
Prometheus retention store.

Closes #187.

## Implementation

- **Shared serializer**: new `write_frame_json<W: Write>(&mut W, &Snapshot)`
  in `src/snapshot/serializers/json.rs`, re-exported through
  `src/snapshot/serializers/mod.rs`. Both the `snapshot` subcommand
  and `record` use it, preventing drift. `Snapshot` gains `Deserialize`
  with `#[serde(default)]` on optional section fields so the replay
  side reconstructs frames even when `--include` omitted a section
  during recording.

- **Record pipeline** (`src/record/`):
  - `mod.rs` — `Recorder::start` orchestrator: signal handlers (SIGINT,
    SIGTERM), interval ticker, header frame, sparse index frame every
    1000 data frames. Local and remote sources supported; remote
    reuses `RemoteCollectorBuilder` from the `view` scrape path.
  - `writer.rs` — `RotatingWriter` handles `.zst`/`.gz` encoders and
    rolls over at `--max-size`, keeping at most `--max-files` segments.
  - `replay.rs` — `Replayer` streaming cursor with bounded frame
    cache (`FRAME_CACHE_MAX = 1024`), sparse-index fast-path for
    seeks, malformed-line recovery, and exact-match error message for
    unsupported schemas.

- **Replay in the TUI**:
  - `src/view/data_collection/replay_collector.rs` — `ReplayDriver`
    feeds frames into `AppState` honoring play/pause/speed/step/seek.
    No renderer branches on replay mode; only the status bar differs.
  - `src/ui/chrome.rs` — new `print_replay_bar` chip with
    `REPLAY | ts | frame N / M | Xx | state`; falls back to the
    normal hotkey strip when `replay` is `None` on `AppState`.
    Filter bar still wins over the replay chip so operators can
    filter visible GPUs mid-playback.
  - `src/view/event_handler.rs` — strict mode precedence: filter-edit
    mode > replay-timecode editor > normal keys > replay keys. The
    `g` key opens the timecode editor only when replay is active; the
    existing `g → sort by GPU memory` binding is preserved otherwise.

- **CLI**: `RecordArgs`, `RecordSource`, `RecordCompression` added to
  `src/cli.rs`; `ViewArgs` gains `replay`, `speed`, `start`,
  `replay_loop`. `main.rs` dispatches `Record` alongside `Snapshot`,
  and the `View` branch short-circuits to `run_replay_mode` when
  `--replay` is set.

- **Dependencies**: adds `zstd = "0.13"` and `flate2 = "1"` at the
  crate root (both are needed at all times because the writer and
  reader select by file extension at runtime, not by feature flag).

- **Docs**: README gains a "Recording & Replay" section with the
  record CLI table, replay keybindings table, and usage examples
  (local, zstd, remote hostfile, timecode start, loop).

## Testing

`cargo test` surface:

- `src/record/writer.rs` — codec detection, rotation eviction, plain
  / gzip / zstd round-trips.
- `src/record/replay.rs` — prime-first-frame, schema-v2 exact error
  message, step forward/back, seek to offset, gzip decoder, zstd
  decoder, corrupted-tail line skip.
- `src/record/mod.rs` — `parse_duration` / `parse_byte_size`
  suffix parsing (`30s`, `5m`, `1h`, `1d`, `1K`, `10M`, `2G`).
- `src/view/event_handler.rs` — 8 new replay-mode key tests: SPACE
  toggles pause, `[`/`]` step+pause, `+`/`-` cycle speed, `j`/`k`
  ±10s seek, `g` opens timecode editor and Enter commits, `L`
  toggles loop, replay keys inert when `replay = None`, filter
  mode wins over replay mode.
- `tests/record_replay_test.rs` — 8 integration tests: plain/gzip/zstd
  round-trip, corrupted tail skip, schema-v2 rejection, seek with
  and without index frames, 50k-frame streaming check.

Runs:
- `cargo test --lib` — 695 passed
- `cargo test --bin all-smi` — 781 passed (↑ from 766; +15 new tests)
- `cargo test --test record_replay_test` — 8 passed
- `cargo clippy --all-targets -- -D warnings` — clean

Manual smoke: `all-smi record --output /tmp/smoke.ndjson --duration 5s
--interval 1 --compress none` produced a header + 5 data frames;
`--output /tmp/smoke.ndjson.zst` produced a zstd file that `zstd -dc`
decompresses correctly; `--max-size 1024 --max-files 3` capped the
total on-disk segments at 3 (active + 2 rotated).

## Test plan

- [x] Build `all-smi record` and `all-smi view --replay` binaries.
- [x] `cargo test --lib`, `cargo test --bin all-smi`, and integration
      test all green.
- [x] `cargo clippy --all-targets -- -D warnings` clean.
- [x] Smoke-recorded local stream round-trips through replay parser.
- [x] Compressed output (`.zst`, `.gz`) round-trips.
- [x] Rotation cap (`--max-size 1K --max-files 3`) evicts oldest.
- [ ] Manual TUI replay walkthrough on a multi-host recording
      (reviewer to validate keybinding feel + status-bar layout).


inureyes:

### Intent
> `all-smi record` captures a live metric stream to NDJSON (with optional zstd/gzip compression and rotation); `all-smi view --replay <file>` plays it back through the same RenderSnapshot pipeline as the live TUI.

### Findings Addressed (all fixed in 770f829)

- [x] **[CRITICAL] SIGTERM/SIGINT during record truncated output to zero bytes.** The unconditional `std::process::exit(0)` handlers installed in `run_command` raced the recorder's cooperative stop flag and killed the process before `RotatingWriter::finish()` could flush the zstd/gzip trailer. Directly violated the "SIGTERM during recording closes cleanly with a complete final JSON line" acceptance criterion. Fix: skip the global signal handlers when the subcommand is `Record`; the record path's own handlers then cleanly flush and exit. Empirically verified: before fix, a SIGTERM'd recording was 0 bytes. After fix, the same SIGTERM produces a proper 1.6 KB zstd file with 5 decompressible NDJSON lines.

- [x] **[MEDIUM] Replay-mode arrow/PageUp/PageDown keys took the local-process-list branch.** `handle_up_arrow` etc. classified remote vs local via `args.hosts.is_some() || args.hostfile.is_some()`, but `--replay` sets neither, so in replay mode the arrow keys tried to scroll a non-existent process list even though the UI rendered tabs + GPU columns. Fix: include `args.replay.is_some()` in the remote classification.

- [x] **[MEDIUM] Index-frame fast-path off-by-one in `Replayer::rewind_and_seek_to`.** The record writer places index frames AFTER the matching data frame (`if seq % 1000 == 0 { write(index { seq }) }` runs after the data frame has already been written). On the read side, skipping past the index line leaves the reader at data frame `nearest_seq + 1`, but the old code set `next_disk_seq = nearest_seq`, so every subsequent frame was labeled with a seq that was off by one. The REPLAY status-bar "frame N / M" display would drift after a fast-path seek. Added a regression test.

### Verification
- [x] All stated requirements implemented — record subcommand, replay mode, shared serializer, schema version error message, corrupted-tail handling, streaming discipline (no `read_to_string`/`read_to_end` in production code), rotation+eviction, SIGTERM clean shutdown (after fix).
- [x] No placeholder/mock code remaining.
- [x] Integrated into project code flow — `Commands::Record` dispatches from `main.rs`; `View` short-circuits to `run_replay_mode` when `--replay` is set; the replay driver feeds `AppState` through the same path as live collectors.
- [x] Project conventions followed — `thiserror` for library errors, `anyhow` for the CLI entry point, inline format args, per-module tests alongside the code.
- [x] Existing modules reused — shared `write_frame_json` serializer, shared `RemoteCollectorBuilder` for the remote record source, shared `RenderSnapshot` pipeline for the replay UI.
- [x] No unintended structural changes — all additions, existing APIs untouched.
- [x] Tests pass — 782 unit tests + 8 integration tests + clippy clean. One regression test added for the seek off-by-one.

### Other Observations (non-blocking)
- The replay driver calls `replayer.next()?` + `replayer.prev()?` on every 50 ms tick to decide whether a frame is due. For a 3-s record interval that's 60 next/prev pairs per frame — wasteful but not a correctness issue. Worth revisiting if replay idle CPU shows up in profiling.
- `Replayer::next()` returns `Ok(self.current())` at EOF (not `Ok(None)`), so callers that use `while r.next().is_some()` to drain the stream hang forever. The internal `apply_frame_to_state` path is not affected because it uses `at_eof()` indirectly, but it's an API sharp edge. Worth documenting on the `next()` signature in a follow-up.

inureyes:
## PR Finalization Complete

### Summary

**Tests**: Added 4 new tests; all suites green.
- `src/app_state.rs`: `cycle_speed_up_advances_ladder`, `cycle_speed_down_retreats_ladder`, `cycle_speed_nan_input_does_not_panic` — exercises the speed ladder stepping and the NaN-safe fallback in `cycle_speed`.
- `src/record/replay.rs`: `replayer_scan_budget_prevents_runaway_loop` (F6) — verifies `SCAN_BUDGET_PER_TICK` stops the frame-scan loop from spinning indefinitely on an all-garbage file and that the next valid frame is still reachable on a subsequent call.

**Documentation**:
- `README.md` "Recording & Replay" section now includes a "Security invariants" block covering: writer O_NOFOLLOW symlink refusal, 16 MiB per-line decompression cap, 128 MiB zstd window ceiling, and 1024-entry hosts-list cap.
- Changelog `v0.21.0 (upcoming)` entry updated to mention `record`, `view --replay`, interactive filter bar, and alert history panel.
- Help pane (`h` key) now conditionally renders a "Replay Controls" section listing all 8 replay keybindings when `state.replay.is_some()`, matching the status-bar hint.

**Lint/Format**: `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` both clean.

### Final test counts (HEAD 9c9a089)

| Suite | Count |
|---|---|
| `cargo test --lib` | 698 passed |
| `cargo test --bin all-smi` | 795 passed |
| `cargo test --test record_replay_test` | 8 passed |

Issue #186, opened 2026-04-17
feat(tui): interactive filter query ('/') and threshold alerts for local/remote modes


inureyes:
## Summary

Add two tightly-coupled TUI improvements:

1. **Filter query mode (`/`)** — an `htop`-style inline query bar that filters dashboard rows / GPU cards by a small DSL (`temp>85`, `util<5`, `host~=dgx`, `user=inureyes`).
2. **Threshold alerts** — configurable limits (temp, util idle duration, power) that trigger a visible toast, optional bell, and optional webhook when crossed, with hysteresis to avoid flapping.

Both share a common expression/condition layer and a single threshold configuration section (see the companion "config file" issue).

## Motivation

At 256+ nodes the current TUI requires eyeballing to answer operator questions like "which GPU is the hottest?" or "which nodes are idling?". Color bands already exist (green ≤60%, yellow 60–80%, red >80% — `AppConfig::{CRITICAL_THRESHOLD, WARNING_THRESHOLD}`) and a notification module exists at `src/ui/notification.rs`, but nothing triggers them and there is no way to filter the view. This is low-hanging UX that turns passive data into active signal.

## Current state

- `src/view/event_handler.rs` handles a handful of keys (`q`, `1`/`h`, `d/u/g/p/m/c/f`, arrows, page up/down).
- `src/ui/notification.rs` has scaffolding but no caller paths.
- `AppConfig` has compile-time thresholds but nothing is user-configurable.
- Filtering today is limited to `f` — "show only processes with GPU memory".

### Filter DSL (minimal, parseable in ~100 lines)

- Fields (typed): `temp`, `util`, `mem_pct`, `mem_used`, `mem_total`, `power`, `user`, `host`, `gpu_name`, `driver`, `index`, `uuid`, `pstate`, `numa`, `device_type`.
- Operators:
  - Numeric: `>`, `>=`, `<`, `<=`, `==`, `!=`
  - String: `==`, `!=`, `~=` (regex, compiled with `regex::RegexBuilder::new(...).size_limit(...)`)
- MVP: single predicate. Phase 2 (same PR if simple): `&` (AND), `|` (OR), parentheses.
- Unknown field → parse error. Field absent on a device (e.g., `pstate` on Apple Silicon) → the row is **not matched** (fail closed).

### Filter UX

- `/` enters filter mode: bottom status bar replaces itself with `Filter: _` (cursor, Pallet-like).
[...]

PR #198, opened 2026-04-20
feat: 'all-smi doctor' subcommand for self-diagnosis and support bundle

inureyes:
## Summary

Introduce a read-only `all-smi doctor` subcommand that runs a battery of
environment checks and prints a PASS/WARN/FAIL/SKIP report. Closes #188.

- 50 built-in checks across 14 categories (platform, privileges,
  container, nvidia, amd, apple, gaudi, tpu, tenstorrent, rebellions,
  furiosa, windows, env, network) with stable dotted IDs.
- Hard 3-second per-check ceiling plus bounded-timeout subprocess
  wrappers — no `Command::output()` without a wrapper.
- Checks execute in parallel on a Tokio `Semaphore`-limited pool.
- Exit codes: `0` clean, `1` warnings only, `2` any failures.

## Implementation

- `src/cli.rs` gains a `Doctor(DoctorArgs)` variant with `--json`,
  `--verbose`, `--bundle <PATH>`, `--include-identifiers`,
  `--remote-check <URL>...`, `--skip <ID>...`, `--only <ID>...`.
- `src/main.rs` dispatches to `doctor::run_cli` which returns the exit
  code to the CLI driver.
- `src/doctor/` subtree:
  - `types.rs` — `Check`, `CheckCtx`, `CheckResult`, `Severity`.
  - `exec.rs` — bounded-timeout child-process helpers (`try_exec`,
    `which`).
  - `redact.rs` — regex scrubbing of hostnames, IPv4, IPv6, MAC, kernel
    pointers, and the current username.
  - `report.rs` — human and JSON renderers; ANSI colour controlled by
    `NO_COLOR` + TTY detection.
  - `bundle.rs` — tar.gz archive packer (adds `tar = "0.4"` + existing
    `flate2`). Includes `report.txt`, `report.json`, `env.txt`,
    `version.txt`, `uname.txt`, `lspci.txt`, `lsmod.txt`,
    `dmesg-gpu.txt`, and (macOS `--verbose` only)
    `system_profiler_display.txt`. Leaves a TODO for the effective
    merged-config file pending #192.
  - `checks/` — one module per category, each exposing `pub fn checks()
    -> &'static [&'static Check]`. Every non-applicable path is a clean
    `Skip`; Windows-only probes compile and run on every platform.
  - `mod.rs` — orchestrator with stable-sort rendering, filter
    predicates, `DoctorOptions` normalisation, `Summary::exit_code()`.
- `src/device/platform_detection.rs` gains a `pub mod introspection`
[...]

Issue #185, opened 2026-04-17
feat: add 'snapshot' subcommand for one-shot JSON/CSV/Prometheus output


inureyes:
## Summary

Add a new `all-smi snapshot` subcommand that produces a one-shot machine-readable dump of the current hardware state (GPU/CPU/memory/chassis/process/storage) to stdout. This unlocks scripting, CI probes, Slurm prolog/epilog hooks, and quick `jq`/`yq` piping without having to start the long-running `api` server.

## Motivation

Today the only machine-readable output path is the `api` subcommand's Prometheus `/metrics` endpoint, which requires running a long-lived HTTP server. The library API returns typed values but is only accessible from Rust code. Operators and CI users regularly want the `nvidia-smi --query-gpu=... --format=csv` ergonomics — one process invocation, machine-readable stdout, non-zero exit on failure. `all-smi` has all the data already (every device type already implements `Serialize`), but no CLI path to emit it for a single collection cycle.

## Current state

- `AllSmi::new()` library API returns `Vec<GpuInfo>`, `Vec<CpuInfo>`, etc., all `Serialize + Deserialize`.
- `api` mode uses these to build Prometheus text — long-lived server only.
- `local` and `view` subcommands enter a TUI loop — not scriptable.
- No CLI path for "collect once, print JSON/CSV, exit".

## Proposed design

New subcommand `all-smi snapshot`:

```
all-smi snapshot [--format json|csv|prometheus] [--pretty]
                 [--include gpu,cpu,memory,chassis,process,storage]
                 [--query index,uuid,name,utilization,memory.used,memory.total,temperature,power]
                 [--interval <secs>] [--samples <n>]
                 [--timeout-ms <n>] [--output <path>]
```

Semantics:

- Default: `--format json --pretty --include gpu,cpu,memory,chassis` (no `process`/`storage` by default to keep it fast).
- `--format json` prints a single top-level object `{ "schema": 1, "timestamp": "2026-04-18T...Z", "gpus": [...], "cpus": [...], ... }` matching the library types' serde schema.
- `--format csv` flattens to one row per device. Default columns per `--include` type; `--query` overrides with a comma-separated field list using dot paths (e.g. `memory.used`, `detail.cuda_version`).
- `--format prometheus` emits the exact Prometheus exposition the `/metrics` endpoint would emit for this single collection — i.e. reuses the exporter.
[...]

PR #199, opened 2026-04-20
feat(view): cluster-wide user/process aggregation tab ('V' key)

### LINKED ISSUE #189: feat(view): cluster-wide user/process aggregation tab ('V' key)
## Summary

Add a new "Users" tab to remote `view` mode that aggregates process information across all scraped hosts, grouping by user so operators can answer "who is using the cluster right now and how much?" at a glance. Drill-down to a selected user shows per-node, per-GPU, per-process detail.

## Motivation

`all-smi` already supports `--processes` in `api` mode, which emits per-process metrics with a `username` label. But the remote `view` has no tab that consumes those metrics cluster-wide: process data remains a local concept. For platform operators (Slurm, Backend.AI, Kubernetes GPU pools) the most valuable operational question is cluster-level: who is using what, where, for how long, at what power cost? An aggregation tab turns per-node process data into a first-class operator signal.

## Current state

- `all-smi api --processes` emits per-process Prometheus metrics with labels including `pid`, `user`, `command`, and the host tag.
- The remote view currently renders an "All" tab (summary across hosts) and per-host tabs, but no user-grouped view.
- `src/network/metrics_parser.rs` parses the GPU/CPU/memory metric families — process metrics may or may not be fully parsed into a remote-side structure yet; must be verified and extended as needed.
- `src/api/metrics/` emits the process metric family when `--processes` is passed.

### Tab and keybindings

- New tab accessible via `V` (mnemonic: "Users").
- Within the tab:
  - `u` sort by username (default)
  - `m` sort by total GPU memory
  - `p` sort by total power (derived)
  - `n` sort by node count
  - `t` sort by oldest process start / longest TIME+
  - `Enter` drill down on the highlighted user
  - `ESC` exits drill-down
  - `e` exports the current view to CSV at `~/.cache/all-smi/users-<timestamp>.csv`
  - `f` toggles a system-process filter (hide uid < 1000 / root by default)

### Top-level aggregation table

Columns (all visible only if width permits; collapse low-priority columns on narrow terminals):

```
USER            NODES   GPUs   PROCS   VRAM        POWER*    LONGEST    CMD (top-1 by GPU mem)
inureyes        3       12     18      384 GiB     2.3 kW   6d 03:12   python train.py --bs=...
yeonji          1       4      1       48 GiB      0.9 kW   2:15:02    /opt/llm/infer -m ...
root            5       0      7       0           —        —          containerd-shim
```

`POWER*` is an approximation: `sum(gpu.power_consumption * (user_vram_on_gpu / gpu_total_vram_used_across_all_users))` per GPU, summed across GPUs a user touches. Mark with `*` in header tooltip and document the methodology.

`LONGEST` uses `TIME+` fields already tracked per process (wall clock since start).

### Drill-down view

On Enter, show per-node rows for that user:

```
inureyes / 3 nodes, 12 GPUs
─ dgx-01   GPU 0-3     VRAM 128 GiB   Power 760 W   4 PIDs    python train.py --bs=128 ...
─ dgx-02   GPU 4-7     VRAM 128 GiB   Power 780 W   4 PIDs    python train.py --bs=128 ...
─ dgx-03   GPU 0-3     VRAM 128 GiB   Power 760 W   4 PIDs    python eval.py --split=val
```

`Enter` again drills to the full process list for that user on the selected node (reuse the existing process renderer).

### Partial coverage indicator

If some hosts don't have `--processes` enabled in their API, show a chip `⚠ partial coverage: 3 of 5 nodes reporting process data` at the top of the tab so operators don't misread the numbers.

## Implementation plan

Files to add / modify:

- `src/network/metrics_parser.rs` — confirm process-family parsing; extend to build a `Vec<ParsedProcessRow>` with `host`, `pid`, `user`, `command`, `gpu_index`, `gpu_memory_bytes`, `cpu_pct`, `start_time`.
- `src/api/metrics/` (or equivalent submodule) — verify the per-process metric family includes labels `host`, `user`, `pid`, `gpu_index`. Add any missing labels, add `all_smi_process_start_time_seconds` counter for TIME+ derivation.
- New `src/ui/aggregation/user.rs`:
  - `aggregate_users(&[HostSnapshot]) -> Vec<UserAggregate>` pure function (heavily unit-testable).
  - Handles (host, pid) identity keying so the same PID on different hosts isn't collapsed.
  - Computes the power approximation with documented formula.
- `src/ui/tabs.rs` — add the `Users` variant; integrate with tab cycling and `V` hotkey.
- New `src/ui/renderers/user_renderer.rs` — table + drill-down rendering using the existing `widgets` for styling.
- `src/app_state.rs` — add `users_tab_state { sort, selected_user, drill_host }`, CSV export path.
- `src/view/render_snapshot.rs` — include the aggregated user view so it flows through the snapshot/replay pipeline (so replays also include the Users tab).
- `src/mock/generator.rs` — generate synthetic process entries per mock node so the mock server exercises this tab. Gate behind `ALL_SMI_MOCK_PROCESSES=1` (default on — since `--processes` is already a flag, just emit them in mock always, or follow the existing convention for flagged mock data).
- `src/ui/help.rs` — add the `V` shortcut and the in-tab keys.

## Acceptance criteria

- [x] `all-smi view --hostfile hosts.csv` with hosts in API mode `--processes` shows a `Users` tab.
- [x] Sorting hotkeys (`u`, `m`, `p`, `n`, `t`) re-rank the table correctly.
- [x] `Enter` drills down; `ESC` returns.
- [x] CSV export produces a well-formed file (header + one row per user).
- [x] If zero hosts report process data, the tab shows a clear "no process data; enable --processes on API mode" hint — not an empty table.
- [x] Partial-coverage chip is visible when some hosts report and others don't.
- [x] Unit tests on `aggregate_users` covering: empty input, same PID on two hosts, root filter, user with processes on multiple GPUs across multiple hosts, oldest TIME+ computation.
- [x] Mock cluster with 5 simulated nodes × 3 simulated users renders correctly.
- [x] Works over replayed snapshots (see companion record/replay issue).
- [x] Documentation updated: README Users tab section; help overlay lists the new shortcuts.

## Edge cases & non-goals

- Very large clusters (100+ nodes, 50+ processes each): aggregation must complete within one render budget (< 50 ms). Use a single pass; do not re-group on every keypress — cache the aggregation keyed on snapshot version.
- Power approximation: document the formula; ensure it never produces negative values (cap at zero if numerical issues arise). Mark as "approximation" in the UI.
- Username may be missing on some hosts (Windows API mode). Render as `?` and include such rows in a separate "unattributed" group.
- Non-goal: querying external user directories (LDAP) — `user` is whatever the API reports.
- Non-goal: alerts on user-level usage — follow-up.

inureyes:
## Summary

Implements the cluster-wide Users tab (issue #189) accessible via the `V`
hotkey in remote `view` mode. Aggregates per-process metrics across every
scraped host so operators can answer "who is using the cluster and how
much?" at a glance, with drill-down into per-host and per-process
breakdowns.

## Implementation

- **Aggregator** (`src/ui/aggregation/user.rs`) — pure function
  `aggregate_users(&[HostSnapshot]) -> UserAggregationResult`. Keys by
  `(host, pid)` so the same PID on two hosts counts as two processes.
  Computes the VRAM-weighted power approximation with the formula
  documented in the issue and clamps the result to >= 0.
- **Renderer** (`src/ui/renderers/user_renderer.rs`) — top-level table
  with selectable rows plus a drill-down view listing per-host VRAM /
  power / PID counts, run-length encoded GPU ranges, and the top
  command per host.
- **Metrics exporter** (`src/api/metrics/process.rs`) — extended to
  emit `all_smi_process_memory_used_bytes`,
  `all_smi_process_start_time_seconds` (new, drives the `LONGEST`
  column), and `all_smi_process_cpu_percent` with `pid`, `user`,
  `gpu_index`, `device_uuid`, `command` labels.
- **Remote parser** (`src/network/metrics_parser.rs`) — now parses the
  `process_*` metric family into a `Vec<ParsedProcessRow>` keyed by
  `(pid, gpu_index)`, with a defensive 50 000-row cap.
- **App state** — added `UsersTabState` (sort, drill path, filter flag,
  export toast), memoised `UsersAggregationCache` keyed on
  `data_version`. `RenderSnapshot::capture` now takes `&mut AppState`
  so the aggregation is materialised under the lock exactly once per
  snapshot version.
- **Tab plumbing** — remote + replay collectors insert `"Users"` as
  tab index 1 (right after `All`) so it participates in cycling
  without breaking the existing `All + hosts` layout. LED grid skips
  the synthetic tab when iterating hostnames.
- **Event handler** — new mode precedence ladder documented at the
  top of `handle_key_event`:
  1. Filter-edit (#186)
  2. Replay timecode (#187)
  3. Users-tab keys when the tab is active
  4. Normal global keys
  5. Replay-mode keys

  Users-tab keys (`u/m/p/n/t/f/e/Enter/ESC`) intercept before the
  global GPU-sort bindings so pressing `m` on the tab switches the
  users-sort key rather than falling through to `MemoryPercent`.
- **CSV export** — `e` writes
  `~/.cache/all-smi/users-<timestamp>.csv` (or the XDG / Windows
  equivalent) with RFC-4180 quoting.
- **Replay** — `ParsedProcessRow::from_local_process` lifts recorded
  `ProcessInfo` into the remote-side representation so replays render
  the tab identically to live scrapes.
- **Mock** — `ALL_SMI_MOCK_PROCESSES=1` makes every synthetic NVIDIA
  node emit 1-4 processes per GPU across a rotating user/command pool
  (same env-var convention as `ALL_SMI_MOCK_VGPU`).
- **Help + README** — help overlay adds a dedicated Users-tab section
  and lists `V`; README gets a full Users-tab section covering the
  columns, the power-approximation formula, and the in-tab keybindings.

## Testing

- `cargo test --lib`: 759 passed
- `cargo test --bin all-smi`: 864 passed (8 new Users-tab event-handler
  tests)
- `cargo test --test users_tab_integration_test`: 3 passed — end-to-end
  exporter -> parser -> aggregator round-trip on a 5x3 fixture,
  sort-stability regression, replay-pipeline round-trip.
- `cargo clippy --all-targets -- -D warnings`: clean
- `cargo fmt --all -- --check`: clean
- Aggregator unit tests cover empty input, same-PID-on-two-hosts,
  root/system-UID filter, multi-GPU multi-host fan-out, oldest TIME+,
  partial-coverage detection, negative-power clamp, per-host
  breakdown invariant, and a 100x50 stress run under the <50 ms
  budget.
- Event-handler tests cover `V` jump, users-tab `m/f` intercepts,
  Enter/ESC drill-down, filter-mode precedence over tab keys, and
  Up/Down row navigation bounds.

## Test plan

- [ ] Start a mock cluster with `ALL_SMI_MOCK_PROCESSES=1` and
  verify the Users tab renders correctly with multiple users / hosts.
- [ ] Press `V` to jump to the tab, confirm sort keys re-rank the
  table, and that `Enter` drills into a user / `ESC` returns.
- [ ] Hit `e` and verify the CSV file lands at
  `~/.cache/all-smi/users-<ts>.csv` with the expected header and rows.
- [ ] Stop `--processes` on a subset of hosts and confirm the
  partial-coverage chip surfaces.
- [ ] Replay a recording and confirm the Users tab is populated.

Closes #189


inureyes:
## Security + performance review

Addressed 4 findings with fixes in commit c5dc1b7.

### Findings and fixes

**CRITICAL — CSV export path lacked symlink/TOCTOU hardening**
`src/view/event_handler.rs:export_users_csv` used `std::fs::write`, which follows symlinks. A co-tenant on a multi-user host could plant `~/.cache/all-smi` (or the final `users-<ts>.csv` file) as a symlink to any location the user can write. Snapshot (#185), record (#197), and doctor (#198) already use `O_NOFOLLOW + 0o600` for exactly this reason; the new CSV export regressed that posture.
- Fix: `OpenOptions::new().create_new(true).custom_flags(O_NOFOLLOW).mode(0o600)` on Unix, `share_mode(0) + create_new(true)` on Windows, plus a defense-in-depth check that refuses to proceed if the cache dir itself is a symlink (since `create_dir_all` no-ops on an existing symlinked directory).
- Regression tests: `open_export_secure_refuses_symlinks`, `open_export_secure_sets_owner_only_mode`.

**HIGH — CSV injection (Excel/LibreOffice formula execution)**
`csv_escape` did RFC-4180 quoting but did not guard against the OWASP CSV-injection pattern. A username/command like `=cmd|"/c calc"!A1`, `+SUM(A1)`, or `@DDE(...)` would be written verbatim and executed as a formula when an operator opens the CSV in Excel or LibreOffice Calc. User names and commands both flow in from untrusted remote hosts.
- Fix: when the field's first character is `=`, `+`, `-`, `@`, TAB, or CR, `csv_escape` now wraps the value in quotes and prefixes a single quote (`'`) so spreadsheets treat it as text.
- Regression tests: 5 tests covering `=`, `+`, `-`, `@`, TAB, CR prefixes plus RFC-4180 behaviour on benign inputs.

**HIGH — `command` Prometheus label not length-capped (privacy + amplification)**
`src/api/metrics/process.rs` emitted `command`, `user`, and `name` labels at their full runtime length. Scrape-response amplification: a 4 KiB argv × 3 metric families × N processes × M hosts rapidly bloats the HTTP response. Privacy: command lines regularly contain API tokens and DB URLs which were broadcast to every Prometheus consumer.
- Fix: cap `command` at 256 B, `user` at 128 B, `name` at 128 B with a UTF-8-safe truncation helper that appends `"...(N bytes truncated)"` so operators see that the string was clipped. The remote parser (`src/network/metrics_parser.rs`) mirrors these caps so a malicious remote host bypassing our exporter cannot inflate the per-host accumulator (50 000-row cap × 1024 B × 3 families ≈ 150 MB was reduced to ~45 MB).
- Regression tests: 4 tests covering pass-through, truncation marker, UTF-8 boundary safety, and end-to-end exporter output.

**MEDIUM — Aggregation hot-path double-lookups + redundant clones**
`aggregate_users` Pass 1 and `UserScratch::absorb` both used the pattern `*map.entry(k).or_insert(0) = map.get(&k).copied().unwrap_or(0).saturating_add(...)` which does two hashmap lookups and two `host.clone()` calls per row. For the spec-max cluster (100 hosts × 50 procs = 5 000 rows) that is ~10 000 redundant operations per scrape tick on the single-threaded UI path.
- Fix: collapse to a single `let entry = map.entry(k).or_insert(0); *entry = entry.saturating_add(...);` and pre-size `total_vram_by_gpu` from the combined GPU count.
- Regression tests: existing `aggregation_handles_large_cluster_quickly` covers correctness; added an `#[ignore]`d adversarial stress test `aggregation_survives_adversarial_50k_per_host` (10 × 50 000 rows = 500 000 rows) for future quadratic-regression detection. Release-mode timing: ~0.5 s for 500 000 rows.

### Confirmed non-findings

- **`ALL_SMI_MOCK_PROCESSES` env gate**: used only in `src/mock/templates/nvidia.rs` (invoked from the mock-server template path). No prod exposure.
- **50 000-row cap is per-host (not per-tick)**: confirmed. `MetricsParser::parse_metrics` is called per host, and `MAX_TEXT_SIZE = 10 MB` per host bounds the cluster-wide memory regardless of cap.
- **`MetricBuilder` label escaping**: unchanged. Already strips control chars + escapes `\\`, `"`, `\n`, `\r`.

### Test results

- `cargo test --lib`: 763 passed, 1 ignored
- `cargo test --bin all-smi`: 875 passed, 1 ignored
- `cargo test --test users_tab_integration_test`: 3 passed
- `cargo clippy --all-targets -- -D warnings`: clean
- `cargo fmt --all -- --check`: clean
- Adversarial stress (release): 500 000 rows aggregated in ~0.46 s

inureyes:
## PR Finalization Complete

### Summary

**Tests**: verified passing — 764 lib + 879 bin + 3 integration; all in-scope unit tests (user aggregation, sort stability, power weighting, partial-coverage chip, drill-down invariants, CSV injection guard) already present.

**Documentation**: updated `README.md` with:
- Three new items in the **Security notes** block (under Filtering & Alerts) covering:
  - Users tab CSV export (`e`) O_NOFOLLOW + mode 0600 symlink refusal on Unix / share_mode(0) on Windows
  - Formula injection mitigation in CSV cells (user/command fields via RFC 4180 quoting + single-quote prefix)
  - Per-field label caps in the remote Prometheus parser (user: 128 B, command: 256 B; global: 100 labels / 1024 B per key-value)
- Users tab entry prepended to the v0.21.0 (upcoming) changelog line, covering V hotkey, all in-tab keys (u/m/p/n/t/Enter/ESC/e/f), power approximation, CSV export path, partial-coverage chip, and the security hardening

The Users tab section itself (V hotkey, in-tab keys, power approximation formula with `*` caveat, CSV export path `~/.cache/all-smi/users-\<timestamp\>.csv`, partial-coverage chip) and the help overlay (`src/ui/help.rs`) listings for V and all in-tab keys were already complete and required no changes.

**Lint/Format**: cargo fmt clean, cargo clippy -D warnings clean.

**Final HEAD**: f42a18c

PR #203, opened 2026-04-20
feat(api): SSE streaming endpoint and /snapshot JSON

### LINKED ISSUE #193: feat(api): Server-Sent Events (SSE) streaming endpoint '/events' and '/snapshot' JSON
## Summary

Add two HTTP endpoints to `all-smi api` mode on top of the existing `axum` server:

1. **`GET /events`** — Server-Sent Events stream. Emits a JSON payload per collection cycle. Targets embedded dashboards and Tauri/Electron apps that want live updates without polling `/metrics`.
2. **`GET /snapshot`** — One-shot JSON response. Same schema as a single SSE frame. Convenient HTTP pair for the `all-smi snapshot` CLI.

Both reuse the exact JSON schema shared by the `snapshot` subcommand and the `record` format (`schema: 1`).

## Motivation

The current `/metrics` Prometheus endpoint serves its purpose for time-series scraping, but embedded UIs need lower latency and easier client code. Implementing SSE on the already-present `axum` stack is cheap (SSE is just `text/event-stream` with a keep-alive) and broadens the set of integrations (custom React/Vue/Lit dashboards, Tauri desktop apps, Grafana panel plugins that prefer JSON, simple debugging with `curl -N`). Keeping the JSON schema identical to `snapshot` and `record` means one format, three delivery channels.

## Current state

- `axum` + `tower-http` in `Cargo.toml`.
- `src/api/server.rs` wires the `/metrics` route.
- Collection loop in `src/api/mod.rs` / `src/api/server.rs` polls readers every `--interval` seconds.
- `GpuInfo`, `CpuInfo`, etc. all implement `Serialize` — no new derives needed.

#### `GET /snapshot`

- Content-Type: `application/json`.
- Body: the JSON frame produced by `snapshot` subcommand (single object with `schema`, `timestamp`, `gpus`, `cpus`, …).
- Query params:
  - `?include=gpu,cpu,memory,chassis,process,storage` (default `gpu,cpu,memory,chassis`).
  - `?pretty=1` (default off for HTTP).
- Caching: `Cache-Control: no-store`.
- Response time bound: must complete within one collection cycle + 200 ms.

#### `GET /events`

- Content-Type: `text/event-stream`.
- Emits one event per collection cycle:
  ```
  event: snapshot
  id: 42
  data: {"schema":1,"timestamp":"...","gpus":[...], "cpus":[...], ...}

  ```
- Keep-alive: emit `: keep-alive\n\n` comment every 30 s (matching `HTTP2_KEEPALIVE_SECS` config).
- Query params:
  - `?include=...` — same semantics as `/snapshot`.
  - `?throttle=N` — emit at most every N seconds (cannot be smaller than the collection interval).
  - `?heartbeat=N` — override heartbeat interval.
- `Last-Event-ID` header support: if client reconnects with a known ID, we simply resume with the next live frame (no replay from history — all-smi doesn't store history).
- CORS: respect the existing `tower-http::cors` configuration; `GET` + `Accept: text/event-stream` must be allowed.

### Broadcast architecture

- Single `tokio::sync::broadcast::channel<Arc<SnapshotFrame>>` with buffer 16.
- The existing API collection task sends one `Arc<SnapshotFrame>` per tick.
- Each SSE client is a receiver. Lagging receivers get the oldest frames dropped (broadcast semantics): in that case, emit an `event: lag\ndata: {"dropped": N}\n\n` event and continue.
- Drop of client connection propagates via `receiver.next()` returning `None` — release resources.

### Schema

Identical to the `snapshot` CLI JSON schema — must share serialization code. Place a `SnapshotFrame` struct in `src/common/snapshot.rs` (or re-export from the snapshot module), both subcommands consume it.

## Implementation plan

Files to add / modify:

- `src/api/server.rs` — add routes `/events` and `/snapshot`; spawn a singleton collection task that broadcasts `SnapshotFrame`s into a `broadcast::Sender`.
- New `src/api/handlers/events.rs` — SSE handler implementing keep-alive, backpressure, lag notification, and the `include` filter.
- New `src/api/handlers/snapshot.rs` — one-shot JSON handler. Reads the last broadcast frame; if stale beyond 2×interval, forces a fresh collection.
- `src/api/mod.rs` — export the shared `SnapshotFrame` state.
- Shared module `src/common/snapshot.rs` — `SnapshotFrame`, `FrameBuilder::new().with_include(...).build()`. Used by `snapshot` CLI and by these endpoints.
- `Cargo.toml` — `axum` feature `sse` if not already enabled; `tokio-stream` if needed.
- Example client in `examples/sse_client.html` — small HTML+JS using `EventSource`.

## Acceptance criteria

- [x] `curl -N http://localhost:9090/events` streams JSON events at the configured interval.
- [x] `curl -N 'http://localhost:9090/events?include=gpu'` streams events with only the `gpus` key populated.
- [x] `curl http://localhost:9090/snapshot` returns a single JSON object matching the snapshot schema.
- [x] `curl 'http://localhost:9090/snapshot?include=cpu,memory'` returns a subset of keys.
- [x] Keep-alive comments are emitted every ~30 s when no data has changed (very unlikely at normal intervals, but for edge cases).
- [x] 50+ concurrent SSE clients do not slow the collection loop (collection tick jitter stays within ±20 ms; measure with a synthetic test).
- [x] Slow client: the server does not buffer unbounded memory. Lagging receivers see a `lag` event and resume.
- [x] Disconnecting a client releases its resources immediately (verified via `lsof` or tokio-console).
- [x] Unix Domain Socket transport (already supported for `/metrics`) also works for `/events` and `/snapshot`.
- [x] `examples/sse_client.html` opens in a browser and shows live updates.
- [x] `cargo test` integration test spins up an api server on a random port, connects an EventSource client, asserts at least 3 frames within 5 seconds.
- [x] README gains a "Streaming (SSE)" subsection under the API section; API.md documents both endpoints.

## Edge cases & non-goals

- SSE is HTTP/1.1 streaming. Reverse proxies (nginx, haproxy) may buffer — document that users should set `X-Accel-Buffering: no` / `proxy_buffering off`. Emit the header from our side (`X-Accel-Buffering: no`).
- Heartbeat interval must be smaller than typical proxy idle timeouts (default 30 s should be safe; allow override).
- The `processes` section is expensive — only populate when requested via `?include=process` to preserve `/events` default cheapness.
- Non-goal: WebSocket transport. SSE is simpler, one-way, and sufficient.
- Non-goal: HTTP/2 server push. SSE works fine over HTTP/1.1 and 2.
- Non-goal: historical replay of missed frames. Clients missing a frame get the next live one.

## Soft dependency

- Shares `SnapshotFrame` with the `snapshot` CLI (dependency) and the `record` format. Land the shared schema first; ship these three features on a common foundation.

inureyes:
## Summary

Adds two HTTP endpoints to `all-smi api` mode that share the
`schema: 1` Snapshot shape used by the `snapshot` CLI and the `record`
NDJSON format — one transport, one serializer, three delivery channels
(#193).

- `GET /snapshot` — one-shot JSON payload.
- `GET /events` — Server-Sent Events stream, one JSON frame per
  collection cycle.

Both work over the existing TCP and Unix Domain Socket transports.

## Implementation

- `api/frame_bus.rs` — `FrameBus` wrapping `tokio::sync::broadcast::Sender<Arc<Snapshot>>` (16-frame buffer) plus a `RwLock` over the latest published frame. `publish()` never blocks on receivers, so slow SSE clients cannot stall the collection loop.
- `api/collection_loop.rs` — extracted the background reader loop from `server.rs`, added single-shot publish onto `FrameBus` after each cycle.
- `api/server_state.rs` — composite `ApiState` with `FromRef` impls so `/metrics` keeps its `SharedState` extractor while `/events` and `/snapshot` extract `FrameBus` directly.
- `api/handlers/events.rs` — SSE handler. Applies `?include=` filter, `?throttle=N` (clamped to ≥ collection interval), and `?heartbeat=N` (default 30 s). Emits `event: snapshot`, falls back to `event: lag\ndata: {"dropped": N}` when the receiver falls behind the broadcast buffer. `Last-Event-ID` is accepted but never replays history (all-smi has no persisted history). Responds with `X-Accel-Buffering: no` and `Cache-Control: no-store`.
- `api/handlers/snapshot.rs` — one-shot JSON. Reads the last broadcast frame; falls back to a fresh `DefaultSnapshotCollector` run when the cached frame is older than `2 × interval` or no cycle has published yet. Filters the `Snapshot::serde_json` output by `?include=` and supports `?pretty=1`.
- `api/handlers/metrics_render.rs` — unchanged Prometheus handler moved into the `handlers/` directory.
- `api/server.rs` — wired the new routes onto the existing Router; no change to the TCP/UDS listener logic.

Docs:
- README: new "Streaming (SSE)" subsection under the API block.
- API.md: new "JSON Endpoints (Streaming + One-Shot)" section with the `/snapshot` / `/events` surface.
- `examples/sse_client.html`: minimal browser `EventSource` demo.

## Testing

`cargo test --test sse_events_test` (9 integration tests, all pass):

- `events_emits_at_least_three_frames_within_five_seconds` — spec's ≥3 frames in 5 s requirement.
- `events_include_filter_drops_other_sections` — `?include=gpu` yields only `gpus`.
- `events_throttle_reduces_emission_rate` — `?throttle=1` caps output rate over a 2 s window.
- `events_lag_event_emitted_for_slow_receiver` — overruns the broadcast buffer with 24 frames before the client polls, asserts the `event: lag` frame appears.
- `fifty_concurrent_clients_do_not_stall_the_publisher` — 50 subscribers, measured publisher tick jitter stays within `interval + 40 ms`.
- `snapshot_returns_latest_frame`, `snapshot_include_filter_drops_sections`, `snapshot_pretty_flag_produces_multiline_body`, `snapshot_falls_back_to_fresh_collect_when_stale`.

Plus new unit tests in `api/frame_bus.rs` (publish/subscribe/drop), `api/handlers/events.rs` (throttle/heartbeat clamp), `api/handlers/snapshot.rs` (include parser).

Test commands run locally:

- [x] `cargo test --lib --features cli`
- [x] `cargo test --bin all-smi --features cli`
- [x] `cargo test --test sse_events_test --features cli`
- [x] `cargo clippy --all-targets --features cli -- -D warnings`
- [x] `cargo fmt --all -- --check`

Closes #193

### DIFF
--- a/API.md
+++ b/API.md
@@ -150,6 +150,26 @@ turned off for the SSE route.
   `EventSource` demo.
 - `curl -N http://localhost:9090/events` — raw inspection.
 
+### Security knobs
+
+| Env var | Default | Effect |
+|---------|---------|--------|
+| `ALL_SMI_API_CORS_ALLOWED_ORIGINS` | (empty → **no CORS**) | Comma-separated list of origins permitted to read `/metrics`, `/snapshot`, and `/events` cross-origin. Set to `*` to revert to the pre-0.21 wildcard; a warning is logged in that case. |
+| `ALL_SMI_API_MAX_SSE_SUBSCRIBERS` | `256` | Cap on concurrent `/events` subscribers. Extra clients get `503 Service Unavailable` with `Retry-After: 5`. Set to `0` to disable the cap. |
+
+The default CORS posture blocks cross-origin browser reads of the live
+telemetry stream. That includes process command lines (when
+`--processes` is enabled), usernames, GPU utilization, and power
+consumption — keep the allowlist narrow.
+
+**Process label truncation.** `command`, `process_name`, and `user`
+fields are truncated at 256 / 128 / 128 bytes respectively on all
+output surfaces (Prometheus `/metrics`, JSON `/snapshot`, and SSE
+`/events`). Longer strings are emitted with a trailing
+`...(N bytes truncated)` marker. This caps scrape-response
+amplification and limits the blast radius of secrets that happen to
+live in argv.
+
 ## Available Metrics
 
 ### GPU Metrics (All Platforms)
--- a/API.md
+++ b/API.md
@@ -51,6 +51,105 @@ r = session.get('http+unix://%2Ftmp%2Fall-smi.sock/metrics')
 
 **Security**: Socket permissions are set to `0600` (owner-only access).
 
+## JSON Endpoints (Streaming + One-Shot)
+
+Alongside `/metrics`, API mode exposes two JSON endpoints that share the
+same schema as the `snapshot` subcommand (`schema: 1`).
+
+### `GET /snapshot`
+
+One-shot JSON payload reflecting the latest collection cycle. When the
+last cached frame is older than `2 × interval` (or no cycle has
+completed yet), the handler forces a fresh collection so the response
+never silently serves stale data.
+
+```
+GET /snapshot
+Content-Type: application/json
+Cache-Control: no-store
+X-Accel-Buffering: no
+
+{
+  "schema": 1,
+  "timestamp": "2026-04-20T12:34:56Z",
+  "hostname": "gpu-01",
+  "gpus":   [ … ],
+  "cpus":   [ … ],
+  "memory": [ … ],
+  "chassis":[ … ],
+  "errors": []
+}
+```
+
+**Query parameters**
+
+| Parameter | Default | Meaning |
+|-----------|---------|---------|
+| `include` | `gpu,cpu,memory,chassis` | Comma-separated sections. Valid values: `gpu`, `cpu`, `memory`, `chassis`, `process`, `storage`. |
+| `pretty`  | `0` | Pretty-print the JSON body when `1` / `true` / `yes`. |
+
+### `GET /events`
+
+Server-Sent Events stream. Emits one JSON frame per collection cycle.
+
+```
+GET /events
+Accept: text/event-stream
+
+event: snapshot
+id: 2026-04-20T12:34:56Z
+data: {"schema":1,"timestamp":"2026-04-20T12:34:56Z", …}
+
+event: snapshot
+id: 2026-04-20T12:34:57Z
+data: { … }
+```
+
+**Query parameters**
+
+| Parameter | Default | Meaning |
+|-----------|---------|---------|
+| `include` | `gpu,cpu,memory,chassis` | Same grammar as `/snapshot`. |
+| `throttle` | = collection interval | Minimum seconds between emitted events. Clamped to `>=` the collection interval. |
+| `heartbeat` | `30` | Keep-alive comment interval in seconds. |
+
+**Keep-alive and lag**
+
+The server emits a bare SSE comment every `heartbeat` seconds so reverse
+proxies do not idle-timeout the connection:
+
+```
+: keep-alive
+```
+
+When a client falls behind the server's broadcast buffer, the stream
+inserts a synthetic `lag` event and continues with the freshest live
+frame:
+
+```
+event: lag
+data: {"dropped": 3}
+```
+
+**Reconnect**
+
+`Last-Event-ID` is accepted but never triggers history replay — `all-smi`
+does not retain historical frames. On reconnect the client resumes with
+the next live frame regardless of the ID it sent.
+
+**Reverse-proxy guidance**
+
+The response advertises `X-Accel-Buffering: no` and `Cache-Control:
+no-store`. nginx operators should additionally set `proxy_buffering
+off;` on the location block. haproxy users want `option http-buffer-request`
+turned off for the SSE route.
+
+**Example clients**
+
+- [`examples/sse_client.html`](examples/sse_client.html) — browser
+  `EventSource` demo.
+- `curl -N http://localhost:9090/events` — raw inspection.
+
 ## Available Metrics
 
 ### GPU Metrics (All Platforms)
--- a/README.md
+++ b/README.md
@@ -825,6 +825,70 @@ Metrics are available at `http://localhost:9090/metrics` (TCP) or via Unix socke
 
 For a complete list of all available metrics, see [API.md](API.md).
 
+#### Streaming (SSE)
+
+Alongside `/metrics`, API mode exposes two JSON endpoints that share the
+same schema as the `snapshot` subcommand (`schema: 1`):
+
+| Endpoint | Content-Type | Purpose |
+|----------|--------------|---------|
+| `GET /snapshot` | `application/json` | One-shot JSON dump of the latest collection cycle. |
+| `GET /events` | `text/event-stream` | Server-Sent Events: one JSON frame per collection cycle. |
+
+```bash
+# One-shot JSON — same schema as `all-smi snapshot --format json`.
+curl http://localhost:9090/snapshot
+
+# Subset by section (same grammar as the snapshot CLI):
+curl 'http://localhost:9090/snapshot?include=gpu,cpu'
+
+# Pretty-print for human inspection:
+curl 'http://localhost:9090/snapshot?pretty=1'
+
+# Live stream, one event per collection cycle. Use `-N` to disable
+# curl's output buffering.
+curl -N http://localhost:9090/events
+
+# Throttle to one event per 5 s, ask for only GPUs.
+curl -N 'http://localhost:9090/events?include=gpu&throttle=5'
+```
+
+Both endpoints respect the existing Unix Domain Socket transport:
+
+```bash
+curl --unix-socket /tmp/all-smi.sock http://localhost/snapshot
+curl -N --unix-socket /tmp/all-smi.sock http://localhost/events
+```
+
+Query parameters for `/events`:
+
+| Parameter | Default | Meaning |
+|-----------|---------|---------|
+| `include` | `gpu,cpu,memory,chassis` | Comma-separated sections to emit. |
+| `throttle` | = collection interval | Minimum seconds between emitted events (clamped to ≥ interval). |
+| `heartbeat` | `30` | Keep-alive `: keep-alive` interval in seconds. |
+
+Each SSE frame is emitted as:
+
+```
+event: snapshot
+id: 2026-04-20T12:34:56Z
+data: {"schema":1,"timestamp":"2026-04-20T12:34:56Z","hostname":"…","gpus":[…],…}
+
+```
+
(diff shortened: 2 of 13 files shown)


inureyes:
## Security + performance review (PR #203)

Reviewed the SSE streaming + `/snapshot` endpoints against the issue-specified attack surface. Findings addressed in 43e4a50:

### CRITICAL

**Wildcard CORS leaked telemetry cross-origin.**
`src/api/server.rs` configured `CorsLayer::new().allow_origin(Any).allow_methods(Any).allow_headers(Any)` on all routes. Any browser-facing page could `fetch` `/metrics`, `/snapshot`, and `/events` from any origin and read GPU utilization, process command lines (when `--processes` is on), usernames, hostnames, and power data. That's especially dangerous for the SSE stream because a single cross-origin `EventSource` can tail telemetry indefinitely.
- Fix: deny-by-default CORS. `ALL_SMI_API_CORS_ALLOWED_ORIGINS` opts in to a comma-separated allowlist; `*` restores the old wildcard with a loud warning. Methods are now GET+OPTIONS, headers are the minimum for `text/event-stream`.

**Process label caps (from #189 hardening) did not propagate to the JSON/SSE surface.**
The Prometheus exporter in `src/api/metrics/process.rs` already caps `command`/`process_name`/`user` at 256/128/128 bytes to mitigate scrape-response amplification and argv-embedded-secret exposure. The JSON path in `/snapshot` and `/events` serialized these fields raw, so the same hardening did not apply. A long argv containing `?password=...` / API tokens would be broadcast to every SSE subscriber verbatim.
- Fix: share the truncation helper. `filter_snapshot_value` now truncates process fields through `ProcessMetricExporter::truncate_for_label` so every wire-format path inherits the cap and the `...(N bytes truncated)` marker.

### HIGH

**`/snapshot` amplification DoS.**
When the last broadcast frame is older than `2 * interval` (server just started, collector stalled, etc.), each incoming `/snapshot` request spawned its own `DefaultSnapshotCollector` and blocking reader set. A burst of requests could saturate Tokio's blocking pool (default 512) and stall `/metrics`, `/events`, and other handlers.
- Fix: single-flight mutex on `FrameBus::lock_fresh_collect()`. Concurrent callers queue, re-check `latest()` after acquiring, and share the winning collector's output. The critical section is bounded by `FRESH_COLLECT_TIMEOUT` (5 s).

**Unbounded SSE subscribers.**
`/events` had no cap on concurrent subscribers, so 10 000 leaked `EventSource` objects from a misbehaving or hostile page would hold 10 000 broadcast-channel slots, FDs, and per-task memory.
- Fix: `ALL_SMI_API_MAX_SSE_SUBSCRIBERS` (default 256) caps concurrent `/events` subscriptions; over-cap clients get `503 Service Unavailable` with `Retry-After: 5`. Set to `0` to disable.

### MEDIUM

**`resolve_throttle` clamp-panic.**
`u64::clamp(lo, hi)` panics when `lo > hi`. If an operator sets `--interval` above `MAX_INTERVAL_SECS` (24 h), every `/events` request panics the handler task. Not a full crash (axum isolates task panics) but returns 500 and spams the log.
- Fix: saturate the floor at `MAX_INTERVAL_SECS` before the clamp call.

**`Last-Event-ID` log amplification.**
The handler logged the full `Last-Event-ID` header to the debug trace. HTTP header size is bounded by hyper but attackers can still inflate log lines with a header near the default limit.
- Fix: truncate at 256 bytes before logging.

**XSS in `examples/sse_client.html`.**
`tr.innerHTML = \`<td>${g.name ?? "—"}</td>...\`` interpolated GPU fields into HTML. A mock/hostile server returning a GPU `name` containing `<script>` could execute JS when operators opened the demo client.
- Fix: build every cell via `document.createElement("td")` + `textContent`.

### Verified safe

- **Query-param parsing** (`?throttle=-1`, `?throttle=NaN`, etc.) — `serde_urlencoded` rejects with 400 at `Query<T>` extraction; no panic path.
- **`Last-Event-ID` parsing** — handler only calls `v.to_str()` and logs; no replay path, no state-shared data.
- **Broadcast `lag` timing side channel** — attacker can only observe their own receiver lagging, which requires them to hold a connection and not consume; no novel info leak beyond what normal timestamps reveal.
- **`X-Accel-Buffering: no`** — set on every `/events` stream response and every `/snapshot` response (including errors).
- **UDS transport** — socket file permissions are `0o600`, restrictive enough for a local-IPC surface.
- **`?include=admin`-style filter bypass** — `parse_include` matches against an explicit allowlist (`gpu,cpu,memory,chassis,process,storage` and aliases); unknown names are silently dropped. No path-traversal or command-injection surface.
- **Float sanitization** — `filter_snapshot_value` calls `sanitize_json_floats` before serialization, so NaN/Inf from flaky drivers cannot fail the response.

### Tests + verification

- `cargo test --lib --features cli` — 928 passed (+7 new regression tests)
- `cargo test --test sse_events_test --features cli` — 11 passed
- `cargo test --bin all-smi --features cli` — 1059 passed
- `cargo clippy --all-targets --features cli -- -D warnings` — clean
- `cargo fmt --all -- --check` — clean

All changes stay within this PR's branch; no new endpoints or breaking JSON schema changes.

inureyes:
## PR Finalization

### Verification results

**tests**: 928 lib + 1059 bin + 11 integration — all pass, unchanged.

**fmt/clippy**: clean.

**`examples/sse_client.html`**: DOM is safe — all cell content written via `textContent` through the `addCell` helper (lines 186-194); `gpuBody.innerHTML = ""` only clears the table body (no user-controlled data injected).

**API.md**: both `/snapshot` (lines 59-90) and `/events` (lines 91-171) are fully documented, including query params, keep-alive, `Last-Event-ID` semantics, `X-Accel-Buffering: no` reverse-proxy guidance, and the security knobs table.

**README "Streaming (SSE)" subsection**: already covered `/events` + `/snapshot` with query params, keep-alive, `Last-Event-ID`, and `X-Accel-Buffering: no`.

**README security notes**: were missing. Added a "Security notes for SSE/snapshot endpoints" subsection (`####`) immediately after the demo-client paragraph, covering:
- `ALL_SMI_API_CORS_ALLOWED_ORIGINS` CORS opt-in
- `ALL_SMI_API_MAX_SSE_SUBSCRIBERS` subscriber cap + 503 behaviour
- Process label truncation caps (256/128/128 bytes)
- Single-flight stale fallback in `/snapshot`

### Commit

`646e1fa` — docs(api): add SSE/snapshot security notes to README Streaming section

Ready for merge.

PR #210, opened 2026-05-03
Update default output path in README.md

victorhooi:
Previously, the README seemed to suggest `all-smi` would write to `./all-smi-record.ndjson.zst` in the current directory.

Fixing the readme here to clarify it actually writes to `~/.cache/all-smi` by default.

(To be honest, I find that default a bit confusing - but I'm assuming there's some precedent here?)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1100,17 +1100,17 @@ operator can filter the visible GPUs mid-playback.
 
 **`record` options summary**
 
-| Flag              | Default                           | Description                                      |
-|-------------------|-----------------------------------|--------------------------------------------------|
-| `--output` / `-o` | `all-smi-record.ndjson.zst`       | Output path. Extension picks the codec.          |
-| `--interval` / `-i` | `3`                             | Seconds between frames.                          |
-| `--duration`      | `0` (= record until SIGTERM)      | Accepts `30s`, `5m`, `1h`, `1d`, or bare seconds. |
-| `--source`        | `local`                           | `local` (hardware readers) or `remote` (HTTP scrape). |
-| `--hosts` / `--hostfile` | (none)                     | Required when `--source=remote`.                 |
-| `--include`       | `gpu,cpu,memory,chassis`          | Comma-separated sections (plus `process`).       |
-| `--max-size`      | `100M`                            | Rotation threshold per segment (`1K`, `10M`, `2G`). `0` disables rotation. |
-| `--max-files`     | `10`                              | Max segments on disk (active + rotated).         |
-| `--compress`      | auto (from extension)             | `zstd` / `gzip` / `none`.                        |
+| Flag                     | Default                                              | Description                                                                |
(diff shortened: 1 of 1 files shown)


inureyes:
Thanks @victorhooi ,  and your instinct is right on both counts: the README was wrong, **and** the default is more confusing than it should be. I traced it through the code:

**Your correction is accurate.** The effective default really is `~/.cache/all-smi/records/all-smi-record.ndjson.zst`. It comes from `Settings::default()` always setting `record.output_dir = "~/.cache/all-smi/records"` (`src/common/config_file.rs`), which `RecorderOptions::from_args_with_settings` joins with the default basename even when **no** config file is present (`src/record/mod.rs`).

**Why it feels confusing**- the default is surfaced inconsistently across three places:
- `all-smi record --help` still advertises the `--output` default as `all-smi-record.ndjson.zst` (`src/cli.rs`), i.e. the current directory;
- the source doc comment says the compiled default is `./all-smi-record.ndjson.zst` (`src/record/mod.rs`);
- only the README (after your fix) says `~/.cache/...`.

So this PR fixes the README, but `--help`- where most people look first- would still point at the working directory. On top of that, the effective path is computed by string-matching `--output` against the magic basename to detect "operator didn't pass `-o`", which is brittle (the code comment itself says *"if that ever changes, update this guard"*) and has a latent wart: passing `-o all-smi-record.ndjson.zst` explicitly gets silently redirected into the cache dir.

**Cross-platform caveat:** the `~/.cache/...` default is a hard-coded string expanded via `dirs::home_dir()` (`src/common/paths.rs`), not the platform-aware `dirs::cache_dir()`. So on Windows it lands at `C:\Users\<you>\.cache\all-smi\records\...` (a Unix-style dotdir in the profile) and on macOS at `~/.cache/...` rather than `~/Library/Caches/...`- functional, but not where users on those platforms expect cache data. The *config* path, by contrast, already uses `dirs::config_dir()` (`%APPDATA%\all-smi` on Windows), so there's an internal inconsistency worth reconciling too.

Because of that, we'd like to fix this a little more thoroughly: align `--help`, the doc comment, and the README on one default, make the path resolution explicit rather than sentinel-based, and consider deriving the cache location from `dirs::cache_dir()` so it's platform-correct. **We'll fold your README correction into that change, with credit to you.**

One process note: we can't merge this PR as-is because the CLA check is still unsigned. We'll track the fuller fix as a separate issue/PR. Thank you again for catching this and for the clear write-up- the "I find this default a bit confusing" remark is exactly what prompted the deeper look. 🙏

inureyes:
Tracking the fuller fix in #223 :  it covers the doc alignment (`--help` / doc comment / README), de-brittling the default-path resolution, and the cross-platform cache location. Your README correction will be incorporated there with credit. Thanks again @victorhooi! 🙏

Issue #205, opened 2026-04-28
Fix build failure with libamdgpu_top 0.11.4 — method renamed (get_all_proc_usage → update_proc_usage)


sempervictus:
## Summary

`all-smi` 0.20.1 fails to compile against `libamdgpu_top` 0.11.4 because the upstream crate renamed `FdInfoStat::get_all_proc_usage` to `FdInfoStat::update_proc_usage` in a patch release. Cargo's caret semver resolves `libamdgpu_top = "0.11.3"` to 0.11.4 (released 2026-04-23), breaking all fresh builds on Linux glibc targets that pick up AMD GPU support.

## Background

The `Cargo.toml` declares the AMD-target dependency as:

```toml
[target.'cfg(all(target_os = "linux", not(target_env = "musl")))'.dependencies]
libamdgpu_top = "0.11.3"
```

Cargo treats `"0.11.3"` as `^0.11.3` (i.e., `>=0.11.3, <0.12.0`). Once `libamdgpu_top` 0.11.4 was published on 2026-04-23, fresh resolutions select 0.11.4, and the call at `src/device/readers/amd.rs` fails to compile:

```
error[E0599]: no method named `get_all_proc_usage` found for struct `FdInfoStat` in the current scope
   --> src/device/readers/amd.rs:572:20
help: there is a method `update_proc_usage` with a similar name
```

`libamdgpu_top` 0.11.4 contains additional internal renames (the `pid_map` field became `pre_proc_usage_map`, and `get_cpu_time` changed signature), but we do not consume those names — only `get_all_proc_usage` is reachable from our code.

This is upstream's failure to honor semver: the 0.11.x train shipped a public-API rename in a patch bump rather than a `0.12.0` minor bump. We need to both adapt to the rename and harden our version constraint so a future patch surprise does not recur.

## Proposed Solution

1. **Adopt the new method name.** Update the only call site at `src/device/readers/amd.rs:572` from `fdinfo.get_all_proc_usage(&proc_index);` to `fdinfo.update_proc_usage(&proc_index);`.
2. **Pin the dependency exactly.** Change `libamdgpu_top = "0.11.3"` to `libamdgpu_top = "=0.11.4"` in `Cargo.toml`. The exact-version requirement prevents Cargo from silently picking up a future patch that breaks the API again. Re-evaluate the pin when upstream restores semver discipline or when we are willing to verify each new release.
[...]

PR #207, opened 2026-04-28
fix: pin libamdgpu_top to =0.11.4 and rename get_all_proc_usage call (#205)

inureyes:
## Summary

Fixes build failure caused by `libamdgpu_top` 0.11.4 renaming
`FdInfoStat::get_all_proc_usage` → `FdInfoStat::update_proc_usage` in a
patch release. Cargo's caret resolution for `"0.11.3"` silently picks up
0.11.4, breaking all Linux glibc fresh builds.

Closes #205

## Changes

1. **`Cargo.toml`**: Changed `libamdgpu_top = "0.11.3"` to
   `libamdgpu_top = "=0.11.4"` (exact pin) under the Linux-glibc target
   table, with a comment explaining the reason for pinning.
2. **`src/device/readers/amd.rs` line 572**: Renamed
   `fdinfo.get_all_proc_usage(&proc_index)` →
   `fdinfo.update_proc_usage(&proc_index)` — the only call site.
3. **`Cargo.lock`**: Regenerated via
   `cargo update -p libamdgpu_top --precise 0.11.4`.

## Verification Results

| Check | Result |
|-------|--------|
| `cargo build --release` | PASSED (macOS) |
| `cargo fmt --check` | PASSED |
| `cargo clippy --all-targets -- -D warnings` | Pre-existing failures in `src/doctor/checks/{amd,furiosa,gaudi,privileges}.rs` — 7 unused-import warnings exist identically on `main` before this PR; not introduced by this change. |
| `cargo test` | PASSED (21 passed, 14 ignored) |
| `grep -rn get_all_proc_usage src/` | Returns zero matches — call site fully renamed |
| Linux `cargo check` (AMD target) | Requires CI — AMD reader is gated by `cfg(all(target_os = "linux", not(target_env = "musl")))` and cannot compile on macOS. Final AMD-target verification depends on CI. |

## Notes

- The exact-pin (`=0.11.4`) prevents future upstream patch surprises from
  silently breaking the build again. Re-evaluate the pin when bumping to a
  new `libamdgpu_top` minor and verifying the API surface.
- A patch release (e.g., `0.20.2`) should follow merge so downstream
  library consumers can pick up the fix without waiting for the next minor
  release.

PR #219, opened 2026-05-18
fix: upgrade libamdgpu_top to 0.11.5 to fix AMD GPU file descriptor leak

### LINKED ISSUE #218: AMD GPU readers leak a file descriptor per instantiation on Linux (fixed in libamdgpu_top 0.11.5)
## Summary

On Linux hosts with an AMD GPU present, each call into the public API leaks a single file descriptor that is never released. The following all leak one fd per call on the reporter's system:

1. `AllSmi::new()`
2. `get_gpu_readers()`
3. `AmdGpuReader::default()` / `AmdGpuReader::new()`

With a typical per-process fd limit of 1024, repeatedly instantiating these (~1020 times) exhausts the limit and crashes the process. These descriptors should be released on drop.

Reported by @joshhansen with a minimal reproduction: https://github.com/joshhansen/allsmi-libamdgpu_top-bug (run `cargo run --release`). Reproduced on commit `d5b678d`.

## Background

The leak originates in the upstream `libamdgpu_top` crate, **not** in all-smi's own code. `DevicePath` opened the DRM device via `into_raw_fd()`, which transfers ownership away from the RAII wrapper so the descriptor is never closed. Every new `AmdGpuReader` (constructed through `get_gpu_readers()` → `AllSmi::new()`) opens a fresh device handle, so file descriptors accumulate one per instantiation.

all-smi currently exact-pins `libamdgpu_top = "=0.11.4"` (`Cargo.toml:76`). That pin was added in #207 (closing #205) because 0.11.4 renamed `get_all_proc_usage` → `update_proc_usage` in a patch release, violating semver and breaking caret resolution on fresh installs. The pin comment explicitly notes it must be re-evaluated when bumping.

Upstream issue: https://github.com/Umio-Yasuno/amdgpu_top/issues/163

## Proposed Solution

Bump the exact pin to `libamdgpu_top = "=0.11.5"`.

Version 0.11.5 (published 2026-05-18) fixes the leak in upstream commit [`8ade0d5`](https://github.com/Umio-Yasuno/amdgpu_top/commit/8ade0d5b8f4e8ba6e91b2313ef5bc658d2a57c05): `DevicePath` now caches an `Arc<OwnedFd>` in a `OnceLock<Arc<OwnedFd>>` and `get_fd()` returns the cached `RawFd`, so the descriptor is owned and closed via RAII on drop instead of leaked.

## Implementation Notes

- **`Cargo.toml:76`** — change `libamdgpu_top = "=0.11.4"` → `"=0.11.5"`; update the adjacent comment (currently explaining the 0.11.4 rename rationale) to also note the fd-leak fix and that 0.11.5 retains the `update_proc_usage` API.
- **Regenerate the lockfile**: `cargo update -p libamdgpu_top --precise 0.11.5`.
- **No source changes expected:**
  - all-smi never calls the changed `get_fd()` API (verified — no `get_fd`/`into_raw_fd` usage anywhere in `src/`), so the upstream signature change `io::Result<RawFd>` → `RawFd` does not affect our call sites.
  - The only AMD API call site, `update_proc_usage` at `src/device/readers/amd.rs:572`, was introduced in 0.11.4 and is retained in 0.11.5.
  - Confirm with `cargo build --release` and `cargo clippy` on a Linux glibc target.
- **Severity context:** all-smi's own `view`/`api` binaries construct readers **once** — outside the collection loop (`src/api/collection_loop.rs:60`) and behind the guarded one-time init (`src/view/data_collection/local_collector.rs:139`) — so the CLI leaks at most one fd per AMD reader, once. The unbounded leak primarily affects consumers of the public library API (`AllSmi::new`, `get_gpu_readers`, `AmdGpuReader`) that re-instantiate repeatedly, which is the reporter's scenario.
- **Build matrix scope:** the dependency only applies to `cfg(all(target_os = "linux", not(target_env = "musl")))`; musl/static and non-Linux targets are unaffected.

## Acceptance Criteria

- [ ] `libamdgpu_top` pinned to `=0.11.5` in `Cargo.toml`, with the explanatory comment updated.
- [ ] `Cargo.lock` regenerated to 0.11.5.
- [ ] `cargo build --release` and `cargo clippy` pass on Linux glibc (AMD code path compiles; `update_proc_usage` call site unchanged).
- [ ] Repeatedly instantiating `AllSmi::new()` / `get_gpu_readers()` / `AmdGpuReader::default()` (>1020 iterations) no longer grows the process fd count unbounded and no longer crashes — verified against the reporter's reproduction.
- [ ] No regression in AMD GPU metrics collection (`view`/`api` modes still report AMD GPU info correctly).

---

### Title: `AllSmi::new`, `get_gpu_readers`, and `AmdGpuReader::new` leak file descriptors on Linux with AMD GPU present

A minimal reproduction can be seen here: https://github.com/joshhansen/allsmi-libamdgpu_top-bug
Just run `cargo run --release`.

This appears to stem ultimately from crate `libamdgpu_top` --- I've filed [an issue there](https://github.com/Umio-Yasuno/amdgpu_top/issues/163) but wanted to make this project aware.

These three lines of code invoking the allsmi API each appear to leak a single file descriptor on my system:
1. `AllSmi::new().unwrap();`
2. `get_gpu_readers();`
3. `AmdGpuReader::default()`

Since my per-process file descriptor limit is 1024, all it takes to trigger this is to instantiate `AllSmi` 1020 times. (The process's own fds do the rest.)

It should be expected that these would be cleaned up on drop.

Thanks

joshhansen:
As reported in my issue https://github.com/Umio-Yasuno/amdgpu_top/issues/163 libamdgpu_top has a file descriptor leak that's fixed in their 0.11.5 now released.

This PR unpins libamdgpu_top and brings it up to 0.11.5. I confirmed with [my bug repro repository](https://github.com/joshhansen/allsmi-libamdgpu_top-bug) that this change resolves the FD leak coming from all-smi.

---

_Maintainer note: merged with `libamdgpu_top` kept **exact-pinned** at `=0.11.5` (rather than unpinned to a caret range) to preserve the semver-break protection added in #207 — this crate has renamed public APIs in patch releases before. See the review comment for details._

Closes #218

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2426,9 +2426,9 @@ checksum = "09edd9e8b54e49e587e4f6295a7d29c3ea94d469cb40ab8ca70b288248a81db2"
 
 [[package]]
 name = "libamdgpu_top"
-version = "0.11.4"
+version = "0.11.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "767e7df82c1bba462f6093a4b611ac59b33a768290cd8c7fdb32a5b8cb8ec013"
+checksum = "20e9cb5c980bdf72b4b5e8dc3efb186749ceaa9b527b91ec388c6e1d95b58012"
 dependencies = [
  "libdrm_amdgpu_sys",
  "nix 0.31.2",
@@ -2443,9 +2443,9 @@ checksum = "48f5d2a454e16a5ea0f4ced81bd44e4cfc7bd3a507b61887c99fd3538b28e4af"
 
 [[package]]
 name = "libdrm_amdgpu_sys"
-version = "0.8.13"
+version = "0.8.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "380c3cfe531e0237f58a16c75710b17eaa006614c6ab115efa921ac8b019c25b"
+checksum = "3b63f219bdee8484c65e0f54f81b6c671f1ab433f460e40c3174dea688abf31b"
 dependencies = [
  "libc",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -71,9 +71,7 @@ libloading = "0.9"
 
 # AMD GPU support (glibc only, not compatible with musl static linking)
 [target.'cfg(all(target_os = "linux", not(target_env = "musl")))'.dependencies]
-# Pinned exactly: 0.11.4 renamed get_all_proc_usage → update_proc_usage in a
-# patch bump (issue #205). Re-evaluate when bumping.
-libamdgpu_top = "=0.11.4"
+libamdgpu_top = "0.11.5"
 
 [target.'cfg(target_env = "musl")'.dependencies]
 openssl = { version = "0.10.75", features = ["vendored"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -71,7 +71,11 @@ libloading = "0.9"
 
 # AMD GPU support (glibc only, not compatible with musl static linking)
 [target.'cfg(all(target_os = "linux", not(target_env = "musl")))'.dependencies]
-libamdgpu_top = "0.11.5"
+# Pinned exactly: this crate has shipped semver-violating breaking changes in
+# patch releases (0.11.4 renamed get_all_proc_usage → update_proc_usage, #205).
+# 0.11.5 fixes a file-descriptor leak (#218). Re-evaluate the exact version on
+# each bump.
+libamdgpu_top = "=0.11.5"
 
 [target.'cfg(target_env = "musl")'.dependencies]
 openssl = { version = "0.10.75", features = ["vendored"] }


joshhansen:
I assign the copyright to this minor PR to you but I do not sign that extensive multi-lingual contributor agreement

inureyes:
Thanks for tracking this down to the root cause and verifying the fix against your reproduction repo — much appreciated, and thanks as well for the copyright-assignment note above.

Before merging I pushed one small commit to this branch that keeps the dependency **exact-pinned** at `=0.11.5` (rather than the caret `0.11.5`) and restores the explanatory comment above the dependency. Context for why:

In #207 (which closed #205) we exact-pinned `libamdgpu_top` specifically because this crate has shipped **semver-violating breaking changes in patch releases** — 0.11.4 renamed the public `get_all_proc_usage` → `update_proc_usage` in a patch bump, which broke every Linux glibc build on fresh installs through Cargo's caret resolution. A caret requirement (`0.11.5`) would allow a future `0.11.6` to be selected automatically and re-introduce that risk. Pinning to `=0.11.5` keeps your fd-leak fix while preserving that protection; we re-evaluate the exact version on each intentional bump.

I also added `Closes #218` to the description so the merge resolves the tracking issue, and I'll squash-merge. Thanks again for the contribution!

joshhansen:
Makes sense - thanks for the quick merge!

PR #227, opened 2026-05-24
fix: resolve record default output path consistently (#223)

### LINKED ISSUE #223: record: default output path is inconsistently documented and resolved (and not platform-aware)
## Summary

The `record` subcommand's default output path is documented inconsistently across `--help`, the source doc comments, and the README, and the effective path is resolved through a brittle string-sentinel mechanism. The cache base is also a hard-coded `~/.cache/...` string that ignores platform conventions on Windows and macOS.

This was surfaced by #210 (@victorhooi), which correctly fixed the README table but left the deeper inconsistencies untouched. The CLA on that PR is unsigned, so the fix is re-implemented here, incorporating the README correction with credit to the original reporter.

## Current behavior

- **Effective default** (no `-o`, no config file): `~/.cache/all-smi/records/all-smi-record.ndjson.zst`. `Settings::default()` always sets `record.output_dir = "~/.cache/all-smi/records"` (`src/common/config_file.rs`), and `RecorderOptions::from_args_with_settings` joins the default basename under it even when no config file is present (`src/record/mod.rs`).
- `all-smi record --help` advertises the `--output` default as `all-smi-record.ndjson.zst` — i.e. the current working directory (`src/cli.rs`).
- The source doc comment says the compiled default is `./all-smi-record.ndjson.zst` (`src/record/mod.rs`).
- The README (after #210) is the only surface that states `~/.cache/...`.

## Problems

1. **Docs disagree** — `all-smi record --help`, where most users look first, contradicts the README and the actual behavior.
2. **Brittle resolution + latent wart** — "operator did not pass `-o`" is detected by string-matching `args.output` against the magic basename (`src/record/mod.rs`); the in-code comment itself flags this as fragile (*"if that ever changes, update this guard"*). As a side effect, passing `-o all-smi-record.ndjson.zst` explicitly is silently redirected into the cache dir instead of the current directory.
3. **Not platform-aware** — `~/.cache/...` is a hard-coded string expanded via `dirs::home_dir()`, not the platform-aware `dirs::cache_dir()`. So on Windows it lands at `C:\Users\<you>\.cache\all-smi\records\...` (a Unix-style dotdir in the profile) and on macOS at `~/.cache/...` rather than `~/Library/Caches/...`. By contrast the config path uses `dirs::config_dir()` (`%APPDATA%\all-smi` on Windows), so the codebase is internally inconsistent.

## Proposed solution

**Scope A — core fix (low-risk, no change to where files land):**
- Make `RecordArgs::output` an `Option<PathBuf>` (drop the clap `default_value`); resolve the effective default explicitly in `from_args_with_settings` (`None` → `<output_dir>/<DEFAULT_BASENAME>`, `Some(p)` → honored verbatim). This removes the string-sentinel fragility and the explicit-`-o` redirect wart.
- Introduce a single shared `DEFAULT_BASENAME` constant (dedupe the literal currently in `src/cli.rs` and `src/record/mod.rs`).
- Align all doc surfaces on one default string: the clap `long_help`, the `mod.rs` doc comment, and the README (fold in #210's README correction, credit @victorhooi).

**Scope B — platform-correct cache location (decision needed; may be a separate issue):**
- **B1:** derive the cache base from `dirs::cache_dir()` so Windows → `%LOCALAPPDATA%\all-smi\records`, macOS → `~/Library/Caches/all-smi/records`, Linux → `$XDG_CACHE_HOME/all-smi/records`. Apply consistently to the energy WAL and users CSV as well. **Note:** this moves existing files on macOS/Windows (and Linux when `$XDG_CACHE_HOME` is set) and requires a migration/changelog note.
- **B2:** keep `~/.cache/...` deliberately for cross-platform consistency, but document the rationale and the platform expansion explicitly.

Recommendation: land Scope A first; treat Scope B (especially B1, a behavior change) as a separate decision/PR to keep changes focused. Either way, the "config is platform-aware but cache is hard-coded `~/.cache`" inconsistency is worth reconciling.

## Affected files

- Scope A: `src/cli.rs`, `src/record/mod.rs`, `README.md`
- Scope B: `src/common/config_file.rs`, `src/common/paths.rs`, and for consistency `src/metrics/energy_wal.rs`, `src/view/event_handler.rs`

## Acceptance criteria

- [x] `all-smi record --help`, the source doc comment, and the README all state the same default output path.
- [x] Default-path resolution no longer relies on string-matching the basename; an explicit `-o <path>` is always honored verbatim, including when it equals the default basename.
- [x] `DEFAULT_BASENAME` is defined once and shared between the CLI arg and the resolver.
- [ ] (Scope B, if chosen) the cache location is platform-correct via `dirs::cache_dir()`, applied consistently across record output / energy WAL / users CSV, with a migration/changelog note.
- [x] Tests cover: no `-o` + no config; no `-o` + config `output_dir` set; explicit `-o` equal to the default basename; explicit `-o` absolute path.

## References

- Originated from #210 (README correction by @victorhooi) — to be incorporated here with credit; that PR cannot be merged directly because its CLA is unsigned.

inureyes:
## Summary

Fixes the inconsistency between what `all-smi record --help`, the source doc comment on `RecorderOptions::from_args_with_settings`, and the README each claimed as the default output path, and removes the brittle string-sentinel resolver that silently redirected an explicit `-o all-smi-record.ndjson.zst` under the cache directory.

## Behaviour change

Before this PR, the resolver detected "operator did not pass `-o`" by string-comparing `args.output` against the magic basename `all-smi-record.ndjson.zst`. As a side effect, an operator who passed `-o all-smi-record.ndjson.zst` explicitly was silently redirected under `record.output_dir` (the cache dir) instead of having their flag honoured verbatim. After this PR, `-o <path>` is always honoured exactly as passed — including when `<path>` happens to equal the default basename. There is no change to where files land when `-o` is omitted.

## What changed

- `src/cli.rs` — `RecordArgs::output` is now `Option<PathBuf>` with no clap `default_value`. The doc comment is rewritten as a clap `long_help` so `-h` stays terse and `--help` describes the actual resolution rules (config `record.output_dir` default `~/.cache/all-smi/records/`, basename `all-smi-record.ndjson.zst`, or just the basename in CWD when no config is loaded).
- `src/record/mod.rs` — adds a shared `pub const DEFAULT_BASENAME: &str = "all-smi-record.ndjson.zst"` next to `INDEX_EVERY_N_FRAMES`. Rewrites the output-resolution block in `from_args_with_settings` to `match &args.output`: `Some(p)` is honoured verbatim; `None` joins `DEFAULT_BASENAME` under the (tilde-expanded) config `output_dir`, or falls back to `DEFAULT_BASENAME` in CWD. The old `args.output == default_basename && let Some(dir)...` guard (and its self-deprecating "if that ever changes, update this guard" comment) is gone. The precedence doc block on `from_args_with_settings` is rewritten to match.
- `README.md` — the `record` options-summary table now shows the actual default for `--output` as `~/.cache/all-smi/records/all-smi-record.ndjson.zst` with the override hints inline. This folds in the README correction originally proposed in #210 by @victorhooi; that PR could not be merged directly due to an unsigned CLA, so the fix is reapplied here with credit.

## Tests

Four unit tests added to `record::tests` cover all four acceptance-criteria branches of the resolver:

- `resolve_output_no_cli_no_config` — no `-o`, no config → just `DEFAULT_BASENAME` in CWD.
- `resolve_output_no_cli_with_config_dir` — no `-o`, config `output_dir = /tmp/all-smi-records` → `/tmp/all-smi-records/all-smi-record.ndjson.zst`.
- `resolve_output_explicit_matching_basename_is_honored` — regression guard for the silent-redirect bug: `-o all-smi-record.ndjson.zst` with config `output_dir` set is honoured verbatim, not redirected under the cache dir.
- `resolve_output_explicit_absolute_path_is_honored` — `-o /var/log/cluster.ndjson.zst` is honoured verbatim, with or without config.

## Scope B is intentionally deferred

The issue defines a Scope B (derive the cache base from `dirs::cache_dir()` so Windows / macOS land in their platform-correct cache directories instead of `~/.cache/...`) and recommends "land Scope A first; treat Scope B as a separate decision/PR to keep changes focused." This PR follows that recommendation and does not touch `src/common/config_file.rs`, `src/common/paths.rs`, `src/metrics/energy_wal.rs`, or `src/view/event_handler.rs`. A follow-up issue could track the platform-aware cache work, which is a behaviour change that moves existing files on macOS, Windows, and Linux-with-`$XDG_CACHE_HOME`-set and warrants its own migration / changelog note.

## Test plan

- [x] `cargo check --lib --tests` passes.
- [x] `cargo test --bin all-smi record::` — 33 passed, including the 4 new resolver tests (the `record` module lives under the `all-smi` binary target, not the library).
- [x] `cargo clippy --bin all-smi --tests -- -D warnings` passes.
- [x] `cargo fmt --check` passes.

Credit to @victorhooi for the original README correction in #210.

Closes #223

### DIFF
--- a/README.md
+++ b/README.md
@@ -1102,7 +1102,7 @@ operator can filter the visible GPUs mid-playback.
 
 | Flag              | Default                           | Description                                      |
 |-------------------|-----------------------------------|--------------------------------------------------|
-| `--output` / `-o` | `all-smi-record.ndjson.zst`       | Output path. Extension picks the codec.          |
+| `--output` / `-o` | `~/.cache/all-smi/records/all-smi-record.ndjson.zst` (override via `record.output_dir` config, or `-o` flag) | Output path. Extension picks the codec. |
 | `--interval` / `-i` | `3`                             | Seconds between frames.                          |
 | `--duration`      | `0` (= record until SIGTERM)      | Accepts `30s`, `5m`, `1h`, `1d`, or bare seconds. |
 | `--source`        | `local`                           | `local` (hardware readers) or `remote` (HTTP scrape). |
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -438,12 +438,20 @@ impl std::fmt::Display for RecordCompression {
 #[derive(Parser, Clone, Debug)]
 pub struct RecordArgs {
     /// Output path for the NDJSON stream.
-    ///
-    /// Extension drives compression: `.ndjson.zst` → zstd,
-    /// `.ndjson.gz` → gzip, anything else → plain NDJSON. Use
-    /// `--compress` to override detection.
-    #[arg(long, short = 'o', default_value = "all-smi-record.ndjson.zst")]
-    pub output: PathBuf,
+    #[arg(
+        long,
+        short = 'o',
+        long_help = "Output path for the NDJSON stream.\n\n\
+            When omitted, the file lands under the `record.output_dir` config \
+            value (default `~/.cache/all-smi/records/`) with basename \
+            `all-smi-record.ndjson.zst`. If no config file is present, the \
+            file is written to the current working directory as \
+            `all-smi-record.ndjson.zst`.\n\n\
+            Extension drives compression: `.ndjson.zst` → zstd, \
+            `.ndjson.gz` → gzip, anything else → plain NDJSON. Use \
+            `--compress` to override detection."
+    )]
+    pub output: Option<PathBuf>,
 
     /// Interval in seconds between frames.
     #[arg(long, short = 'i', default_value_t = 3)]
--- a/src/record/mod.rs
+++ b/src/record/mod.rs
@@ -52,6 +52,11 @@ use writer::{Codec, RotatingWriter};
 /// walk index frames.
 const INDEX_EVERY_N_FRAMES: u64 = 1000;
 
+/// Default basename used for the recorder's NDJSON stream when the
+/// operator does not pass `-o`. Defined once here so the CLI layer,
+/// resolver, doc comments, and tests stay aligned (see issue #223).
+pub const DEFAULT_BASENAME: &str = "all-smi-record.ndjson.zst";
+
 /// Resolved, validated options for a recording run.
 pub struct RecorderOptions {
     pub output: PathBuf,
@@ -80,15 +85,19 @@ impl RecorderOptions {
     /// Merged CLI + `[record]` config constructor. Precedence
     /// (highest → lowest):
     ///
-    /// * CLI flag (`--output`, `--compress`)
+    /// * Explicit CLI flag (`-o <path>` / `--compress`) — always
+    ///   honored verbatim, including when the path happens to equal
+    ///   [`DEFAULT_BASENAME`].
     /// * Config file (`record.output_dir`, `record.compress`)
-    /// * Compiled default (`./all-smi-record.ndjson.zst`, codec zstd)
+    /// * Compiled default ([`DEFAULT_BASENAME`] in the current working
+    ///   directory, codec zstd)
     ///
-    /// `record.output_dir` is treated as the *directory* the default
-    /// basename is placed under — the operator can always pass `-o
-    /// /some/file.ndjson` to bypass it. Tilde expansion is applied so
+    /// When `-o` is omitted, [`DEFAULT_BASENAME`] is placed under
+    /// `record.output_dir` (tilde-expanded so
     /// `output_dir = "~/.cache/all-smi/records"` resolves to the
-    /// user's home at record time.
+    /// user's home at record time). If no config is supplied,
+    /// the default resolves to just [`DEFAULT_BASENAME`] in the
+    /// current working directory.
     pub fn from_args_with_settings(
         args: &RecordArgs,
         settings: Option<&RecordSettings>,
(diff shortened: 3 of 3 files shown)


inureyes:
## PR Finalization Complete

### Test coverage
Added one new resolver test () in . The four existing resolver tests used absolute paths, leaving the  call in the `None`-CLI/Some-config branch untested. The new test passes `output_dir = "~/my-records"` through the resolver and derives the expected path using the same `expand_tilde` helper, making it portable across environments where `$HOME` may or may not be set. Total: 8 record tests, all passing.

### Documentation
- `src/cli.rs` `long_help` on `RecordArgs::output`: clearly states both cases (config default `~/.cache/all-smi/records/` + no-config fallback to CWD). Accurate.
- `src/record/mod.rs` doc comment on `from_args_with_settings`: lists all three precedence levels and names `DEFAULT_BASENAME`. Accurate.
- `README.md` record options table row for `--output` (line 1105): shows `~/.cache/all-smi/records/all-smi-record.ndjson.zst` as the effective default. Accurate (matches the config-layer default).
- No localized READMEs exist in this repository.
- `docs/man/all-smi.1` does not cover the `record` subcommand; no update needed (no `--output` entry to correct).
- No CHANGELOG file exists in this repository.

### Lint / format
- `cargo fmt --check`: clean (no diffs).
- `cargo clippy --bin all-smi --tests -- -D warnings`: clean (no warnings).

### Build sanity
- `cargo check --bin all-smi --tests`: finished successfully.
- `cargo test --bin all-smi record::tests::`: 8 passed, 0 failed.

PR #230, opened 2026-05-24
feat: unify cache paths via platform-aware dirs::cache_dir() (#229)

### LINKED ISSUE #229: cache: unify record/energy-WAL/users-CSV paths via platform-aware dirs::cache_dir() (Scope B from #223)
## Background

Follow-up tracking issue for **Scope B** of #223, intentionally deferred when #227 landed Scope A (the doc/resolution-consistency fix for `all-smi record`'s default output path). #227 deliberately left the cache-path code untouched so this platform-aware work could be decided and shipped on its own — it relocates files on macOS and Windows and needs a migration/changelog note.

**Decision (from review on this issue): adopt B1** — resolve every cache path (record output, energy WAL, users CSV) through a single platform-aware helper built on `dirs::cache_dir()`, and retire the hard-coded `~/.cache/...` strings. Rationale below.

## Why B1 (it's more than "platform correctness")

A code audit shows the cache base is currently resolved **three different ways**, so the locations already diverge per platform *today*:

| Consumer | Source | Linux (`$XDG_CACHE_HOME` set) | macOS | Windows |
|---|---|---|---|---|
| users CSV export | `cache_dir_for_all_smi()` — `src/view/event_handler.rs:671` | `$XDG_CACHE_HOME/all-smi` | `~/.cache/all-smi` | `%LOCALAPPDATA%\all-smi\cache` |
| record output (default) | `record.output_dir = "~/.cache/all-smi/records"` — `src/common/config_file.rs:256` | `~/.cache/all-smi/records` (ignores `$XDG_CACHE_HOME`) | `~/.cache/all-smi/records` | `…\.cache\all-smi\records` (Unix-style dotdir in the profile) |
| energy WAL | `wal_path = "~/.cache/all-smi/energy-wal.bin"` — `src/common/config.rs:197` | `~/.cache/all-smi/energy-wal.bin` (ignores `$XDG_CACHE_HOME`) | `~/.cache/all-smi/energy-wal.bin` | `…\.cache\all-smi\energy-wal.bin` (dotdir) |

- **Windows is already split**: the users CSV lands under `%LOCALAPPDATA%\all-smi\cache`, but record output and the energy WAL land in a Unix-style `.cache` dotdir inside the user profile.
- **Linux is split when `$XDG_CACHE_HOME` is set**: the users CSV honors it; record output and the WAL ignore it.
- **macOS** puts all three under `~/.cache/...` (none use the Apple-correct `~/Library/Caches/...`).
- `cache_dir_for_all_smi()` is a **hand-rolled reimplementation of `dirs::cache_dir()`**, with a comment that it avoids pulling in the `dirs` crate "for a single lookup." That rationale no longer holds: `dirs` is already a direct dependency (used by `src/common/paths.rs` for `config_dir()`/`home_dir()`), and there are now three lookups, not one.
- The config path is already platform-aware via `dirs::config_dir()` (`src/common/paths.rs:95`), so the cache path being hard-coded is an internal inconsistency.

B1 collapses these three paths into one helper, deletes `cache_dir_for_all_smi()`, and aligns the cache layer with the config layer.

## Target layout

Resolve a single cache root via `dirs::cache_dir().join("all-smi")`:

| Platform | `dirs::cache_dir()` | all-smi cache root |
|---|---|---|
| Linux | `$XDG_CACHE_HOME` or `~/.cache` | `…/all-smi` |
| macOS | `~/Library/Caches` | `~/Library/Caches/all-smi` |
| Windows | `%LOCALAPPDATA%` | `%LOCALAPPDATA%\all-smi` |

Then: record output → `<cache-root>/records/`, energy WAL → `<cache-root>/energy-wal.bin`, users CSV → `<cache-root>/users-<ts>.csv`.

## Recordings also live under `cache_dir()` (decided)

`dirs::cache_dir()` maps to OS-purgeable locations (macOS may reclaim `~/Library/Caches` under storage pressure; cleanup tools target cache dirs). The energy WAL (crash-recovery scratch) and users CSV (on-demand export) are unambiguously cache-like.

`record` outputs are user-created artifacts, so `dirs::data_local_dir()` was considered for them specifically. **Decision: keep all three — record output included — under `dirs::cache_dir()`.** The default output path is a convenience fallback; operators who want durability already pass `-o <path>`, and a single cache root keeps the resolver and the docs simple and consistent. (`data_dir()` was explicitly rejected: on Windows it is Roaming `%APPDATA%`, which would sync potentially-large recordings across machines.)

## Migration

- **Linux without `$XDG_CACHE_HOME`** (the common case): `dirs::cache_dir()` == `~/.cache`, so **nothing moves** — no migration needed.
- **macOS, Windows, and Linux with `$XDG_CACHE_HOME` set**: the cache root relocates. The affected data is low-stakes — recordings remain replayable via an explicit `view --replay <path>`, the WAL is recoverable scratch, and the users CSV is an on-demand export whose path is printed to the user.
- Ship a release/changelog note documenting the new per-platform locations, plus an optional one-time best-effort migration that moves an existing `~/.cache/all-smi/{records,energy-wal.bin}` to the new root only when the new root is empty (never overwrite).

## Implementation notes

- Add a `cache_dir()` helper in `src/common/paths.rs` mirroring the existing `config_dir()`: `dirs::cache_dir().map(|d| d.join(APP_DIR_NAME))`.
- Route all three consumers through it; delete the hand-rolled `cache_dir_for_all_smi()` in `src/view/event_handler.rs` (≈ lines 667–695) and its bespoke `XDG_CACHE_HOME`/`LOCALAPPDATA`/`HOME` probing.
- Reconcile the Windows users-CSV subpath: today it is `%LOCALAPPDATA%\all-smi\cache`; under the shared helper it becomes `%LOCALAPPDATA%\all-smi`.
- Keep `expand_tilde` for **user-supplied** values only (explicit `-o <path>`, and config-file `record.output_dir` / energy `wal_path` overrides); the compiled *default* should come from the platform helper, not a `~/.cache/...` string literal.
- Update the defaults in `src/common/config_file.rs` (`output_dir`) and `src/common/config.rs` (`wal_path`) so they derive from the helper rather than hard-coded `~/.cache/...` literals.
- Preserve the existing symlink / `O_NOFOLLOW` / `0o600` defenses around the users-CSV and any new cache writes.

## Affected files

- `src/common/paths.rs` — new `cache_dir()` helper.
- `src/common/config_file.rs` — `record.output_dir` default.
- `src/common/config.rs` — energy `wal_path` default.
- `src/view/event_handler.rs` — delete `cache_dir_for_all_smi()`, use the helper.
- `src/metrics/energy_wal.rs` — consume the resolved WAL path (no hard-coded literal).
- Docs: `src/cli.rs` (`--output` `long_help`, `ALL_SMI_ENERGY_WAL_PATH` text), `README.md`, `docs/man/all-smi.1` if applicable.

## Acceptance criteria

- [x] A single platform-aware `cache_dir()` helper exists in `src/common/paths.rs` (built on `dirs::cache_dir()`), mirroring `config_dir()`.
- [x] record output, energy WAL, and users CSV all default under `<cache_dir()>/all-smi/...` through that one helper; `cache_dir_for_all_smi()` is removed and records are **not** split out to `data_local_dir`.
- [x] Hard-coded `~/.cache/all-smi/...` default literals are gone from `config_file.rs` and `config.rs`; `expand_tilde` is used only for user-supplied overrides.
- [x] A one-time best-effort migration (or an explicit decision not to migrate) is implemented, with a changelog/release note describing the new per-platform locations.
- [x] `--help`, source docs, README, and the manpage agree on the new defaults.
- [x] Existing symlink / `O_NOFOLLOW` / `0o600` write protections are preserved.

## References

- #223 — defines Scope A/B (closed; Scope A criteria met).
- #227 — landed Scope A; deferred Scope B (merged).
- #210 — original README correction by @victorhooi.

inureyes:
## Summary

Route every `all-smi` cache writer — `record` output, energy WAL, and the users-tab CSV export — through a single platform-aware helper built on `dirs::cache_dir()`. Retires the three hard-coded `~/.cache/all-smi/...` literals and the hand-rolled `cache_dir_for_all_smi()` reimplementation, so the cache layer is finally as platform-correct as the config layer (`config_dir()` already was). Builds on #227 (Scope A of #223).

## What changed

- **New helper.** `crate::common::paths::cache_dir()` mirrors `config_dir()`: `dirs::cache_dir().map(|d| d.join(APP_DIR_NAME))`. Returns `Option<PathBuf>` so callers can downgrade gracefully on bare CI shells / containers without `$HOME`.
- **`record.output_dir` and `energy.wal_path` are now `Option<String>`.** `None` (the new compiled default) means "use the platform helper"; `Some(s)` is an explicit operator override resolved through `expand_tilde`. The CLI flag surface and the TOML key names are unchanged; `ALL_SMI_ENERGY_WAL_PATH` / `ALL_SMI_RECORD_OUTPUT_DIR` continue to work.
- **Consumers updated.** `record::RecorderOptions::from_args_with_settings` now layers `-o` flag → config `output_dir` (`expand_tilde`) → `paths::cache_dir().join("records")` → CWD fallback. `api::server::run_api_mode` resolves the WAL path through a new `metrics::energy_wal::resolve_wal_path(...)` helper that follows the same precedence. `view::event_handler::export_users_csv` calls `paths::cache_dir()` directly. The hand-rolled `cache_dir_for_all_smi()` in `event_handler.rs` is deleted.
- **Energy-WAL signatures cleaned up.** `spawn_wal_flush_task` / `flush_cycle` / `compact_wal` now take `PathBuf` / `&Path` instead of `String` / `&str`, dropping the per-cycle `to_string()` and `Path::new(...)` re-wrapping.
- **Best-effort one-time migration.** New `src/common/cache_migration.rs` is called once from `main()` before any subcommand touches the cache. Linux without `$XDG_CACHE_HOME` is a no-op (old == new). On macOS, Windows, and Linux with `$XDG_CACHE_HOME` set it moves `~/.cache/all-smi/records` and `~/.cache/all-smi/energy-wal.bin` into the new root **only when the destination is empty** (never overwrites) and **refuses to follow a pre-planted symlink at the source** (same threat model as the writer-side `O_NOFOLLOW` defences). The users-CSV exports are deliberately not migrated — they carry per-second timestamps and bulk-renaming hundreds of files would be more disruptive than helpful.
- **`config print` shows the resolved on-disk path** when no override is set, both in TOML and JSON outputs, so operators can confirm at a glance where data will land.
- **Docs.** README updates the `[energy] wal_path`, `[record] output_dir`, `--output` and users-CSV export rows + the WAL persistence paragraph + the security note about the users-CSV export; manpage updates the `FILES` and `ENERGY ACCOUNTING` sections; `RecordArgs::output` `long_help` and the example `config.toml` text spell out the per-platform cache locations. The v0.21.0 changelog entry gets a closing sentence describing the unification and the migration.

## Migration

- **Linux without `$XDG_CACHE_HOME`** (the common case): `dirs::cache_dir() == ~/.cache`, so the old path and the new path are identical and the migration helper is a no-op.
- **macOS, Windows, and Linux with `$XDG_CACHE_HOME` set**: the cache root relocates. Operators will see one-time `all-smi: migrated legacy cache <old> -> <new>` lines on stderr at startup as `records/` and `energy-wal.bin` are moved into the new root. The migration never overwrites an existing destination — if you have data at both the old and the new path, the old data is left in place for manual reconciliation.
- **Symlink-planted sources are refused** with a warning on stderr (`all-smi: skipping cache migration of <path> (is a symlink)`), matching the writer-side `O_NOFOLLOW` rationale.
- **Cross-filesystem moves are not silently doubled** — if `$XDG_CACHE_HOME` points at a tmpfs while `~` is on persistent disk, `std::fs::rename` will return `EXDEV`, the helper logs the error, and recordings stay where they were. Operators can move them by hand.

## New tests

- `paths::tests::cache_dir_ends_with_app_name` — sanity for the new helper.
- `record::tests::resolve_output_no_cli_no_config_uses_platform_cache_dir` — guards the issue-#229 promise that record output flows through `paths::cache_dir()` (skipped on hosts where `cache_dir()` returns `None`, same pattern as `config_dir_ends_with_app_name`).
- `cache_migration::tests` (5 tests): `try_move_skips_when_source_missing`, `try_move_renames_file_when_dst_missing`, `try_move_renames_directory_when_dst_missing`, `try_move_no_op_when_dst_exists`, `try_move_refuses_symlink_source` (Unix-only — symlink creation requires it).
- Existing `record::tests` resolver tests rewritten for the `Option<String>` shape; `config::tests::energy_config_defaults_match_issue_spec` and the `ALL_SMI_ENERGY_WAL_PATH` env test now assert `None` / `Some(...)`.

## Security hardening preserved

The writer-side `O_NOFOLLOW` + mode `0600` defences in `src/record/writer.rs`, `src/metrics/energy_wal.rs::open_secure_append`, and `src/view/event_handler.rs::open_export_secure` apply unchanged to the new resolved paths — only the resolver tier changes. The migration helper has its own defence: it refuses to `rename` through a symlink at the source.

## Test plan

- [x] `cargo check --bin all-smi --tests`
- [x] `RUSTFLAGS="-L /tmp/drmlinks" cargo test --bin all-smi paths::tests::`
- [x] `RUSTFLAGS="-L /tmp/drmlinks" cargo test --bin all-smi record::tests::`
- [x] `RUSTFLAGS="-L /tmp/drmlinks" cargo test --bin all-smi cache_migration::tests::`
- [x] `RUSTFLAGS="-L /tmp/drmlinks" cargo test --bin all-smi config::tests::`
- [x] `RUSTFLAGS="-L /tmp/drmlinks" cargo test --bin all-smi config_file::tests::`
- [x] `RUSTFLAGS="-L /tmp/drmlinks" cargo test --bin all-smi metrics::energy_wal::`
- [x] `RUSTFLAGS="-L /tmp/drmlinks" cargo test --bin all-smi --test config_file_integration_test` (14/14)
- [x] `RUSTFLAGS="-L /tmp/drmlinks" cargo clippy --bin all-smi --tests -- -D warnings`
- [x] `cargo fmt --check`

References #223 and #227. Closes #229.


inureyes:
## PR Finalization Complete

### Summary

**L1 and L2 LOW security-checker fixes applied** in commit b169549.

**Tests added/updated:**
- `src/record/mod.rs`: new `resolve_output_whitespace_only_config_dir_falls_through` — asserts a whitespace-only `output_dir` does not propagate as a literal path; total record resolver tests: 10.
- `tests/config_file_integration_test.rs`: new `whitespace_only_wal_path_and_output_dir_treated_as_unset` — loads TOML with `wal_path = "   "` and `output_dir = "   "` and asserts both resolve to `None`; total integration tests: 15.

**Documentation:**
- README changelog entry for the Users tab corrected: `~/.cache/all-smi/users-<timestamp>.csv` replaced with `<platform cache dir>/all-smi/users-<timestamp>.csv [^cache]` to match the platform-aware semantics introduced by this PR.
- All other doc surfaces (manpage FILES + ENERGY ACCOUNTING, cli.rs ENERGY_HELP, RecordArgs::output long_help, config example.rs, [^cache] footnote) verified correct — no Linux-only literal paths remaining.
- No localized READMEs exist in the repository (only README.md).

**Code quality:**
- `cargo fmt --check`: clean
- `cargo clippy --bin all-smi --tests -- -D warnings`: clean (collapsed nested-if to let-chain &&-form per clippy collapsible_if)
- 1172 unit tests pass, 15 integration tests pass, 1 ignored (platform-specific skip)

**L3 (dst.exists() vs dst.symlink_metadata() in cache_migration.rs)** left as-is per the brief — the security checker confirmed no vulnerability; the tightening is follow-up if desired.

PR #240, opened 2026-05-26
fix(tui): restore terminal cursor on exit (Linux/tmux)

### LINKED ISSUE #235: fix(tui): restore terminal cursor on exit (Linux/tmux leave cursor hidden)
## Problem / Background

After running `all-smi` on Linux and quitting with `q`, the terminal cursor remains hidden in the host shell. The user must manually run `tput cnorm` or `reset` to restore it. `htop` does not exhibit this behavior because it explicitly re-enables cursor visibility on exit (via ncurses `endwin()` which emits `\e[?25h`), and additionally installs signal handlers so SIGINT/SIGTERM/SIGSEGV still restore the cursor.

The bug is invisible on macOS and on some terminal emulators (notably iTerm2 and certain xterm builds) because those terminals happen to save/restore cursor visibility together with the alternate-screen buffer. On the Linux console (TTY), VTE-based terminals (GNOME Terminal, Konsole, etc.), and inside tmux/screen, the two states are independent — so the cursor stays hidden after exit.

## Root Cause

`src/view/ui_loop.rs` lines 105-114 send `cursor::Hide` once at session start. The inline comment claims the cursor will be restored by `TerminalManager::drop()` via `LeaveAlternateScreen`:

```rust
// Hide cursor once at session start. The cursor is restored in
// TerminalManager::drop() (LeaveAlternateScreen resets it).
// This avoids per-frame Hide/Show churn.
```

But `src/view/terminal_manager.rs` lines 64-73 only send `LeaveAlternateScreen` + `DisableMouseCapture` + `disable_raw_mode()` — there is no `cursor::Show`:

```rust
impl Drop for TerminalManager {
    fn drop(&mut self) {
        if self.initialized {
            let mut stdout = stdout();
            let _ = execute!(stdout, LeaveAlternateScreen, DisableMouseCapture);
            let _ = disable_raw_mode();
        }
    }
}
```

The assumption that `LeaveAlternateScreen` (`CSI ?1049l`) restores cursor visibility (`CSI ?25h`) is wrong in the general case. By the DEC spec these are independent terminal modes, and only some terminals couple them. The current code relies on that coincidence, so the bug only manifests on terminals where the modes are truly independent — Linux TTY, VTE-based terminals, tmux/screen, etc.

## Proposed Solution

1. In `src/view/terminal_manager.rs` `Drop` impl, explicitly send `cursor::Show` along with `LeaveAlternateScreen` and `DisableMouseCapture`. Example:

   ```rust
   let _ = execute!(stdout, cursor::Show, LeaveAlternateScreen, DisableMouseCapture);
   ```

   Import `crossterm::cursor` accordingly.

2. Make sure cursor restoration also runs on abnormal exit paths so the terminal isn't left broken when `Drop` doesn't run:
   - Install a panic hook that emits `cursor::Show`, `LeaveAlternateScreen`, `DisableMouseCapture`, and `disable_raw_mode()` before delegating to the default panic handler.
   - Install a Ctrl-C / signal handler (e.g. `ctrlc` crate or a signal stream) that performs the same cleanup before exiting. Keep this idempotent so the normal `Drop` path is still safe.

3. Verify on Linux TTY, GNOME Terminal, and inside tmux that the cursor is visible after `q`, after Ctrl-C, and after an induced panic.

## Acceptance Criteria

- [ ] After quitting `all-smi` with `q` on Linux (TTY + at least one VTE terminal + tmux), the shell cursor is visible again.
- [ ] After Ctrl-C on the same environments, the cursor is visible and the terminal is out of raw mode / alt-screen.
- [ ] After a forced panic in the view loop, the cursor and screen are restored (no need to run `reset`).
- [ ] No per-frame `Show`/`Hide` churn is introduced — the fix only affects setup/teardown paths, not the render loop.
- [ ] The fix does not regress cursor handling on macOS / iTerm2 (idempotent `Show` is safe to emit even when cursor is already visible).

## Technical Considerations

- `crossterm::cursor::Show` maps to `CSI ?25h` and is safe to emit unconditionally on exit, including in environments where the cursor is already visible.
- Panic hook should preserve and delegate to the existing default panic handler so backtraces and debug output are not lost.
- Signal handler should set a flag and trigger cleanup rather than performing IO directly inside the handler when possible, but `crossterm` IO and `disable_raw_mode` are commonly used in such handlers and are acceptable here.
- Reference: `htop` does the equivalent via ncurses `endwin()` (which calls `curs_set(1)`) and registers handlers for SIGINT/SIGTERM/SIGSEGV.

## Priority

Marked `priority:medium` — the cursor is recoverable with `tput cnorm` or `reset`, but it's a visible regression that affects every Linux/tmux user on every exit.

inureyes:
## Summary

On Linux terminals (VTE family, tmux, kitty) cursor visibility is tracked independently of the alternate-screen mode. Leaving the alternate screen does **not** automatically restore a hidden cursor — so after pressing `q`, Ctrl-C, or a panic the shell cursor remained invisible, requiring the user to run `reset`.

## Root cause

Two independent failure modes:
1. `TerminalManager::Drop` emitted `LeaveAlternateScreen` but not `cursor::Show`, so the normal `q`-exit path left the cursor hidden on non-Apple terminals.
2. The SIGINT/SIGTERM handlers called `std::process::exit(0)` directly, which **bypasses all `Drop` impls** — so even adding `cursor::Show` to `Drop` alone would not fix Ctrl-C. (See issue #235 for full analysis.)

## Changes

- **`src/view/terminal_manager.rs`**: Added `TERMINAL_NEEDS_RESTORE` (`AtomicBool`) static flag and a public `restore_terminal()` function. The flag is set in `initialize()` after the alternate screen is entered and cleared atomically in `restore_terminal()`. The function is idempotent — safe to call concurrently from `Drop`, signal handlers, and a panic hook. Added `PANIC_HOOK_INSTALLED` (`Once`) to install a panic hook exactly once; the hook calls `restore_terminal()` before delegating to the previous hook so panic backtraces render on a usable terminal. `TerminalManager::Drop` now delegates to `restore_terminal()` (including explicit `cursor::Show`). Added 3 unit tests for the swap semantics and idempotency.
- **`src/main.rs`**: Added `view::terminal_manager::restore_terminal()` calls at the start of both the SIGINT and SIGTERM handlers, before `std::process::exit(0)`. Renamed `setup_panic_handler` to `setup_panic_handlers` and removed the `#[cfg(target_os = "macos")]` gate on the function definition; the macOS-specific native-metrics shutdown remains gated inside the closure. The call site in `main()` is now unconditional.
- **`src/view/ui_loop.rs`**: Updated stale comment that claimed `LeaveAlternateScreen` resets cursor visibility — which was precisely the bug.

## Verification

Automated checks run:

| Command | Result |
|---|---|
| `cargo check --features cli --lib` | passed |
| `cargo check --features cli --lib --tests` | passed |
| `cargo clippy --features cli --lib --tests -- -D warnings` | passed |
| `cargo test --features cli --bin all-smi -- view::terminal_manager::tests` | 3/3 passed |
| `cargo fmt --check` | passed |

Acceptance criteria requiring **manual verification** by the reviewer:

- Press `q` on Linux (TTY + VTE terminal) — cursor must be visible in the shell afterward.
- Press `q` inside tmux — cursor must be visible.
- Press Ctrl-C — cursor must be visible and terminal must be out of raw mode / alt-screen.
- Induce a panic in the view loop — terminal must be restored without needing `reset`.

The macOS / iTerm2 regression check (idempotent `cursor::Show` is safe when cursor already visible) does not require separate verification — the escape sequence is a no-op when the cursor is already shown.

Closes #235

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -64,9 +64,8 @@ static NATIVE_METRICS_INITIALIZED: AtomicBool = AtomicBool::new(false);
 static HLSMI_INITIALIZED: AtomicBool = AtomicBool::new(false);
 
 fn main() {
-    // Set up panic handler for cleanup
-    #[cfg(target_os = "macos")]
-    setup_panic_handler();
+    // Set up panic handler for cleanup (cross-platform)
+    setup_panic_handlers();
 
     // Best-effort one-time migration of legacy `~/.cache/all-smi/...`
     // data to the platform-correct cache dir (issue #229). Runs before
@@ -191,6 +190,9 @@ async fn run_command(cli: Cli, settings: Settings) {
         // Set up signal handler for clean shutdown
         tokio::spawn(async {
             signal::ctrl_c().await.expect("Failed to listen for Ctrl+C");
+            // Restore terminal state before exit so the parent shell is usable
+            // (issue #235). No-op when no TUI was running.
+            view::terminal_manager::restore_terminal();
             #[cfg(target_os = "macos")]
             {
                 // Cleanup native metrics manager on signal
@@ -210,6 +212,9 @@ async fn run_command(cli: Cli, settings: Settings) {
             let mut sigterm = signal::unix::signal(signal::unix::SignalKind::terminate())
                 .expect("Failed to listen for SIGTERM");
             sigterm.recv().await;
+            // Restore terminal state before exit so the parent shell is usable
+            // (issue #235). No-op when no TUI was running.
+            view::terminal_manager::restore_terminal();
             #[cfg(target_os = "macos")]
             {
                 // Cleanup native metrics manager on signal
@@ -651,12 +656,17 @@ async fn run_command(cli: Cli, settings: Settings) {
     }
 }
 
-// Set up a panic handler to ensure cleanup
-#[cfg(target_os = "macos")]
-fn setup_panic_handler() {
+// Set up a panic handler to ensure cleanup.
+// Cross-platform: the macOS native-metrics shutdown is gated behind
+// `#[cfg(target_os = "macos")]`. The terminal-restoration hook is installed
+// separately inside `TerminalManager::new()` via `PANIC_HOOK_INSTALLED` —
+// it layers on top of whatever hook this function installs, so the terminal
+// is always restored before this hook's body runs.
+fn setup_panic_handlers() {
     let default_panic = std::panic::take_hook();
     std::panic::set_hook(Box::new(move |panic_info| {
-        // Cleanup native metrics manager before panicking
+        // Cleanup native metrics manager before panicking (macOS only)
+        #[cfg(target_os = "macos")]
         device::macos_native::shutdown_native_metrics_manager();
         default_panic(panic_info);
     }));
--- a/src/view/terminal_manager.rs
+++ b/src/view/terminal_manager.rs
@@ -13,14 +13,32 @@
 // limitations under the License.
 
 use std::io::stdout;
+use std::sync::{
+    Once,
+    atomic::{AtomicBool, Ordering},
+};
 
 use crossterm::{
+    cursor,
     event::{DisableMouseCapture, EnableMouseCapture},
     execute,
     terminal::{
         ClearType, EnterAlternateScreen, LeaveAlternateScreen, disable_raw_mode, enable_raw_mode,
     },
 };
+
+/// Process-global flag: `true` when the terminal has been set to alternate-screen /
+/// raw mode and the cursor has been hidden. Any exit path — normal `q`, SIGINT,
+/// SIGTERM, or panic — must call `restore_terminal()` to clear this flag and
+/// emit the cleanup escape sequences.
+///
+/// Initialised `false` so `restore_terminal()` is a no-op for subcommands that
+/// never touch the TUI (e.g., `snapshot`, `doctor`).
+static TERMINAL_NEEDS_RESTORE: AtomicBool = AtomicBool::new(false);
+
+/// `Once` guard that ensures the panic hook is installed exactly once, even if
+/// `TerminalManager::new()` were called multiple times within the same process.
+static PANIC_HOOK_INSTALLED: Once = Once::new();
 
 pub struct TerminalManager {
     initialized: bool,
@@ -51,6 +69,17 @@ impl TerminalManager {
             return Err("Failed to initialize terminal display".into());
         }
 
+        // Mark the terminal as needing restoration on every exit path.
+        // This must be set before `self.initialized = true` so that if any
+        // subsequent initialisation step panics the flag is already live.
+        TERMINAL_NEEDS_RESTORE.store(true, Ordering::SeqCst);
+
+        // Install the panic hook exactly once. `take_hook` captures whatever is
+        // on top of the panic-hook stack at this point — which may be the macOS
+        // native-metrics hook installed earlier by `setup_panic_handlers` in
+        // main.rs — so the chain is preserved correctly.
+        PANIC_HOOK_INSTALLED.call_once(install_panic_hook);
+
         self.initialized = true;
         Ok(())
     }
@@ -64,11 +93,12 @@ impl TerminalManager {
 impl Drop for TerminalManager {
     fn drop(&mut self) {
         if self.initialized {
-            let mut stdout = stdout();
-            // Leave alternate screen and restore terminal state
-            let _ = execute!(stdout, LeaveAlternateScreen, DisableMouseCapture);
-            let _ = disable_raw_mode();
-            // No "Terminating..." message needed - native APIs don't require cleanup
+            // Restore cursor visibility in addition to leaving the alternate screen.
+            // `LeaveAlternateScreen` alone does NOT guarantee cursor visibility on
+            // Linux terminals (VTE family, tmux, kitty, …) that track cursor state
+            // independently of the alternate-screen mode — hence the explicit
+            // `cursor::Show` (issue #235).
+            restore_terminal();
         }
     }
 }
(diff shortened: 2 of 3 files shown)


inureyes:

### Intent
> Restore terminal cursor (and full TUI state) on every exit path — normal `q`, Ctrl-C / SIGTERM, and panic — so Linux/tmux/VTE users no longer need `reset` after `all-smi` exits.

### Findings

**LOW**

- `src/view/terminal_manager.rs:132-134` — The inline comment "Use `std::io::stdout()` directly — no locks, no shared state" is technically inaccurate. `std::io::stdout()` does acquire an internal Mutex inside `execute!`; the intent appears to be "no application-managed locks held across the boundary," but the wording could mislead a future maintainer who needs to reason about deadlock potential in a panic-hook context. Suggest rewording to: "Use `std::io::stdout()` directly — no application-owned locks are held across this call — so this function is safe to call from a panic hook context, from a tokio task, and from `Drop`."

**NIT**

- `src/view/terminal_manager.rs:75,128,171,etc.` — `Ordering::SeqCst` on `TERMINAL_NEEDS_RESTORE` is conservative but stronger than necessary. Since the flag does not synchronise any data (only gates a self-contained block of terminal IO), `Ordering::Acquire`/`Ordering::Release` (or even `Relaxed`) would be sufficient. SeqCst is defensible as the simpler default; not worth changing unless you care about the (negligible) extra fence cost.
- `src/view/terminal_manager.rs:163-198` — The three unit tests share the global `TERMINAL_NEEDS_RESTORE` flag and can race under cargo's default parallel test execution within a binary. Each test does set the flag to its needed starting state, so the asserts cannot fail — but a parallel race could cause `restore_terminal_is_noop_when_flag_unset` to spuriously emit escape codes (the test does not assert against IO). Consider `--test-threads=1` for this module or a `Mutex<()>` test lock. Optional.
- Commit body is hard-wrapped at ~72 cols, which conflicts with the project's documented commit conventions (paragraphs as one long line, per `commit-conventions` skill). Body content is otherwise excellent — every paragraph clearly explains WHY. Cosmetic only; will not block.

### Verification

- [x] All stated requirements implemented (cursor restore on q, signals, panic)
- [x] No placeholder / mock code
- [x] Integrated into project code flow (signal handlers in `main.rs`, `Drop` impl, panic hook installed inside `TerminalManager::new()`)
- [x] Project conventions followed (license header, import grouping, no `unwrap()` on hot paths, `let _ =` documented)
- [x] Existing modules reused (`crossterm`, `tokio::signal`, `std::panic::take_hook`/`set_hook`)
- [x] Tests pass (`cargo test --features cli --bin all-smi -- view::terminal_manager::tests` → 3/3 ok)
- [x] `cargo check --features cli --lib` — clean
- [x] `cargo check --features cli --lib --tests` — clean
- [x] `cargo clippy --features cli --lib --tests -- -D warnings` — clean
- [x] `cargo fmt --check` — clean

### Cross-cutting checks confirmed

- **Panic hook chain ordering**: `setup_panic_handlers` (macOS-native shutdown, no-op on Linux) installs first in `main()`; `install_panic_hook` (terminal restore) installs second inside `TerminalManager::new()`. Execution order on panic: terminal restore → macOS native shutdown → default backtrace hook. Terminal is usable before backtrace prints. Correct.
- **Signal-handler ordering**: `restore_terminal()` is called before `shutdown_native_metrics_manager` / `shutdown_hlsmi_manager` and before `std::process::exit(0)` in both SIGINT and SIGTERM handlers. Matches issue's prescribed order.
- **Flag set-after-IO ordering**: `TERMINAL_NEEDS_RESTORE.store(true)` runs only after both `enable_raw_mode()` AND `execute!(EnterAlternateScreen, ...)` succeed. A failed init leaves the flag `false`, so `restore_terminal()` is a clean no-op — no spurious escape codes for non-TUI subcommands or for partially-initialised terminals. Confirmed.
- **Idempotency under concurrency**: `AtomicBool::swap(false, ...)` is the correct primitive — exactly one caller wins, all others early-return. No TOCTOU window since the flag-clear and cleanup IO are sequenced atomically per caller (the swap is the gate).
- **Feature gating**: `view::terminal_manager` is reachable only via `src/main.rs` (declared as `mod view;`), and the `all-smi` bin requires the `cli` feature. No extra `#[cfg(feature = "cli")]` gates needed inside the module.
- **No regression on non-TUI subcommands**: `snapshot` / `doctor` / `config` / `record` / `api` paths never call `TerminalManager::new()`, so `restore_terminal()` in signal handlers becomes a single atomic load + early return. `record` skips the new signal handlers entirely via the existing `if !is_record` guard.

### Verdict
**APPROVE** — implementation is correct, well-commented, well-tested, and matches the issue's acceptance criteria. The three LOW/NIT items are non-blocking polish suggestions.

inureyes:
## PR Finalization Complete

### Summary

**Comment fix** — The misleading "no locks, no shared state" comment at `restore_terminal()` lines 132-134 has been reworded. The accurate claim is that no *application-owned* locks are held across the `execute!` call; `stdout()` does acquire the stdlib mutex internally and releases it before returning, which is safe from every call site (panic hook, tokio task, `Drop`).

**Test serialization** — Added a module-level `static TEST_MUTEX: Mutex<()>` inside the `tests` block. Each of the three tests acquires a `_guard` before touching `TERMINAL_NEEDS_RESTORE`, ensuring they run sequentially even when cargo parallelises unit tests within a binary. No new dev-dependency introduced.

**Documentation** — No `CHANGELOG.md` at repo root; skipped. No user-facing docs describe TUI exit behavior explicitly (the fix is invisible when working correctly); no doc updates needed.

**Lint/Format** — `cargo fmt --check` clean. `cargo clippy --features cli --lib --tests -- -D warnings` passes with zero warnings. All 3 targeted tests pass.

### Verification

```
cargo fmt --check        → clean
cargo clippy ...         → Finished, 0 warnings
cargo test ... view::terminal_manager::tests
  restore_terminal_clears_flag          ... ok
  restore_terminal_is_idempotent        ... ok
  restore_terminal_is_noop_when_flag_unset ... ok
  3 passed; 0 failed
```

**Commit:** 65a3013 — ready for merge.

PR #261, opened 2026-05-28
fix: align Docker toolchain with current MSRV

inureyes:
## Summary
- Update Docker builder and dev-container scripts from Rust 1.88 to Rust 1.95.
- Record the current crate MSRV in Cargo.toml.
- Align Debian/PPA packaging metadata and vendoring workflow with Rust 1.95.
- Run Docker Build Check on pull requests so Dockerfile/toolchain drift is caught before merge.

## Reason
The post-merge main CI for #232 failed only in Docker Build Check because the Dockerfile still used rust:1.88-slim, while the current lockfile resolves sysinfo@0.39.2, which requires rustc 1.95.

## Validation
- cargo metadata --format-version 1 --locked --no-deps
- cargo fmt --check
- env -u NO_COLOR RUSTFLAGS="-L /tmp/drmlinks" cargo build --release --bin all-smi
- env -u NO_COLOR RUSTFLAGS="-L /tmp/drmlinks" cargo build --no-default-features --lib

Docker build could not be run locally because this environment cannot access the Docker daemon socket; GitHub CI now runs that path on this pull request.

Refs #232

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -93,7 +93,9 @@ jobs:
     name: Docker Build Check
     runs-on: ubuntu-latest
     needs: test
-    if: github.event_name == 'push' && github.ref == 'refs/heads/main'
+    if: >
+      github.event_name == 'pull_request' ||
+      (github.event_name == 'push' && github.ref == 'refs/heads/main')
     steps:
       - name: Checkout
         uses: actions/checkout@v6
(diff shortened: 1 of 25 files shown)


inureyes:
Closing this duplicate follow-up: main already contains the Docker/MSRV alignment via #233 and the Rust 1.95/sysinfo 0.39 update via #258.

PR #266, opened 2026-06-26
refactor: migrate Tenstorrent reader to upstream luwen 0.8.5 crates

### LINKED ISSUE #265: Migrate Tenstorrent support to upstream `luwen` crates (0.8.x) and retire the `all-smi-luwen-*` republished forks
## Summary

The Tenstorrent NPU reader currently depends on four `all-smi-*` crates that we republished to crates.io from Tenstorrent's `luwen` repository:

```toml
[target.'cfg(target_os = "linux")'.dependencies]
all-smi-luwen-core = "0.2.0"
all-smi-luwen-if   = "0.7.9"
all-smi-luwen-ref  = "0.7.9"
all-smi-ttkmd-if   = "0.2.2"
```

These are produced by `scripts/vendor_all_luwen.sh`, which clones `luwen`, renames its crates under an `all-smi-` prefix, strips `publish = false`, and (manually) publishes them. This issue proposes migrating to the **upstream-published `luwen` crates (0.8.x)** and retiring the fork plus the vendoring script.

## Why the forks exist (context)

all-smi is published to crates.io, and crates.io forbids `git`/`path` dependencies in published crates. The Tenstorrent reader needs APIs that, at the time, lived only on luwen's `main` branch. The dependencies were originally declared as `git = "https://github.com/tenstorrent/luwen.git", branch = "main"`. The crates.io releases of `luwen-if` / `luwen-ref` were (and still are) frozen at 0.4.4 / 0.3.1 from March 2024 and lack those APIs. To keep all-smi publishable with the newer code, we vendored `main`, renamed it `all-smi-luwen-*`, and published it under our own ownership.

## Why migrate now

Upstream restructured the workspace and now **publishes a current crate set to crates.io that tracks `main`**:

| Upstream crate | Latest on crates.io | Role |
|---|---|---|
| `luwen` (umbrella) | 0.8.5 (2026-03-30) | re-exports `api`, `def`, `pci`, `kmd` |
| `luwen-api` | 0.8.5 | successor to `luwen-if` + `luwen-ref` (detection lives here too) |
| `luwen-def` | 0.8.5 | successor to `luwen-core` (common definitions, `Arch`) |
| `luwen-pci` | 0.8.5 | PCIe layer; provides the low-level `detect_chips_silent` |
| `luwen-kmd` | 0.8.5 | successor to `ttkmd-if` (pulled transitively) |

The old crate directories (`luwen-core`, `luwen-if`, `luwen-ref`, `ttkmd-if`) **no longer exist on luwen `main`**, so `scripts/vendor_all_luwen.sh` can no longer re-vendor against current upstream regardless. The original reason for the fork is gone: we can depend on upstream's published crates directly.

## Target dependencies

Replace the four fork crates inside the existing Linux-only target block.

Recommended (granular, mirrors the old split):

```toml
[target.'cfg(target_os = "linux")'.dependencies]
luwen-api = "0.8.5"   # Chip, ChipImpl, Telemetry, ChipDetectOptions, DeviceInfo, UninitChip
luwen-pci = "0.8.5"   # detect_chips_silent(options) -> Vec<UninitChip>
luwen-def = "0.8.5"   # Arch
# luwen-kmd is pulled in transitively; no direct dependency needed.
```

Alternative (single umbrella dep): `luwen = "0.8.5"`, then reference `luwen::api::*`, `luwen::def::Arch`, `luwen::pci::detect_chips_silent`. Either is acceptable; the granular form keeps the dependency set explicit and minimal.

## API mapping (verified)

All call sites live in `src/device/readers/tenstorrent.rs`.

| Current (fork) | New (upstream) |
|---|---|
| `all_smi_luwen_core::Arch` | `luwen_def::Arch` |
| `all_smi_luwen_if::ChipDetectOptions` | `luwen_api::ChipDetectOptions` |
| `all_smi_luwen_if::chip::{Chip, ChipImpl, Telemetry}` | `luwen_api::chip::{Chip, ChipImpl, Telemetry}` |
| `all_smi_luwen_ref::detect_chips_silent(opts)` | `luwen_pci::detect_chips_silent(opts)` |

The entire `Telemetry` surface the reader uses is unchanged in name and signature: `arch`, `try_board_type()`, `board_serial_number_hex()`, `arc_fw_version()`, `eth_fw_version()`, `firmware_date()`, `ddr_fw_version`, `spibootrom_fw_version`, `voltage()`, `current()`, `power()`, `asic_temperature()`, `vreg_temperature()`, `inlet_temperature()`, `board_temperature`, `ai_clk()`, `arc_clk()`, `axi_clk()`, `telemetry_heartbeat()`. `DeviceInfo` (`domain` / `bus` / `slot` / `function` / `vendor` / `device_id` + `pcie_current_link_width()` / `pcie_current_link_gen()`) is likewise unchanged. `ChipDetectOptions` still has a `local_only` field and a `Default` impl, so `ChipDetectOptions { local_only: true, ..Default::default() }` continues to work as-is.

## Breaking changes / required code changes

1. **`detect_chips_silent` moved and split.** The old `luwen-ref::detect_chips_silent(options) -> Vec<UninitChip>` is now `luwen_pci::detect_chips_silent(options) -> Result<Vec<UninitChip>, LuwenError>`: same shape, drop-in. Note that `luwen_api::detect_chips_silent` is a *different* function with signature `(root_chips: Vec<Chip>, options) -> Result<Vec<Chip>, _>`; do **not** use that one. Use the `luwen-pci` entry so the existing `UninitChip::init()` loop is preserved.

2. **`Arch::Grayskull` is now `#[deprecated]`** ("legacy architecture that is no longer supported"). The exhaustive match in `extract_static_info` emits `warning: use of deprecated unit variant ... Arch::Grayskull`, and CI (`cargo clippy -- -D warnings`) turns that into a hard error. Fix by adding `#[allow(deprecated)]` on the match, or by replacing the Grayskull arm with a catch-all (keep `Wormhole` / `Blackhole` named, map the rest to `"Unknown"`/`"Grayskull"`) so the deprecated variant is never named. `Arch` is a plain (non-`#[non_exhaustive]`) enum with exactly `Grayskull`, `Wormhole`, `Blackhole`, so the match must remain exhaustive.

3. **`UninitChip::init` error semantics.** It returns `Result<Chip, InitError<E>>`; even with an `Infallible` callback, `InitError::PlatformError` can occur on a real init failure. The current `Err(_) => None` arm with the comment "This should never happen with Infallible" is now misleading. Keep dropping the chip on error, but update the comment and consider logging the platform error.

4. Drop the now-unnecessary bare `use all_smi_luwen_core;` / `use all_smi_luwen_ref;` lines; reference the crates through their items or paths directly.

## Build / CI considerations

- `luwen-api` has a build script that compiles three protobuf files via `prost-build` 0.13, so it needs `protoc` at build time. all-smi's own `build.rs` already requires `protoc` (via `tonic-prost-build` 0.14 for the TPU gRPC protos on Linux), so this adds **no new system requirement** on the Linux build paths. Confirm the musl-static, aarch64-gnu cross, Debian, and PPA build paths still succeed.
- MSRV: luwen targets Rust 1.75 (edition 2021); all-smi is 1.95 / edition 2024 (no conflict).
- `luwen-kmd` depends on `nix` + `memmap2` (pure Rust, no C deps), so musl static linking remains viable.

## Implementation steps

1. `Cargo.toml`: replace the four `all-smi-*` deps with `luwen-api` / `luwen-pci` / `luwen-def` `= "0.8.5"` (or the umbrella `luwen`). Update the stale `# Tenstorrent dependencies from GitHub` comment (they are no longer git deps).
2. `src/device/readers/tenstorrent.rs`: update imports per the mapping table; switch detection to `luwen_pci::detect_chips_silent`; handle the deprecated `Arch::Grayskull` (allow or catch-all); fix the `init` error-handling comment.
3. Remove `scripts/vendor_all_luwen.sh`. The `references/luwen` and `/vendor/` entries in `.gitignore` can stay or be cleaned up.
4. (Owner action, separate from this PR) Optionally deprecate/yank the `all-smi-luwen-*` crates on crates.io once no released all-smi version depends on them. Do not delete older published all-smi releases: they must remain buildable.

## Acceptance criteria

- [ ] all-smi builds on Linux (glibc) against upstream luwen 0.8.x, with no remaining `all-smi-*` fork references anywhere in the tree.
- [ ] `cargo clippy --all-targets -- -D warnings` is clean (in particular the `Arch::Grayskull` deprecation is handled).
- [ ] musl static build and aarch64-gnu cross build succeed.
- [ ] Tenstorrent reader output is unchanged on real hardware: device name, board type, telemetry detail fields, PCIe details, and the power / temperature / clock / utilization metrics. If hardware is unavailable, at minimum confirm the detection path compiles and the mock server path is unaffected.
- [ ] `scripts/vendor_all_luwen.sh` is removed.

## Verification evidence

A throwaway probe crate depending on `luwen-api` / `luwen-pci` / `luwen-def` `= "0.8.5"` and replicating the reader's exact call sites compiles cleanly (`cargo build` finished; `protoc` present in the build environment). Removing `#[allow(deprecated)]` reproduces `error: use of deprecated unit variant luwen_def::Arch::Grayskull` under `-D warnings`, confirming change #2.

## Out of scope (possible follow-ups)

- Surfacing the richer 0.8.x `Telemetry` fields (per-GDDR temperatures, fan RPM, throttler state, board power limit, etc.) as new metrics. This issue is a like-for-like migration.
- Memory utilization (`used_memory`) remains a `TODO` in the reader and is not addressed here.

inureyes:
## Summary

Replace the four republished fork crates (`all-smi-luwen-core`/`-if`/`-ref` and `all-smi-ttkmd-if`) with the upstream `luwen` crates 0.8.5 (published on crates.io, tracking luwen main) in the Tenstorrent NPU reader, and remove the now-obsolete vendoring script. The fork existed only because all-smi is published to crates.io (which forbids git/path deps) and the reader needed APIs that lived solely on luwen main; upstream now ships a current crate set that tracks main, so the fork is no longer necessary.

## What changed

- `Cargo.toml`: swap the `all-smi-luwen-*` deps in the `cfg(target_os = "linux")` block for `luwen-api`, `luwen-pci`, and `luwen-def` `= "0.8.5"`. `luwen-kmd` (the `ttkmd-if` successor) is pulled in transitively, so no direct dep is declared, and the stale `# Tenstorrent dependencies from GitHub` comment is updated.
- `src/device/readers/tenstorrent.rs`: remap imports to `luwen_api::{ChipDetectOptions, chip::{Chip, ChipImpl, Telemetry}}` and `luwen_def::Arch`; switch detection to `luwen_pci::detect_chips_silent` (the PCIe-layer entry returning `Result<Vec<UninitChip>, _>`, not the differently-signed `luwen_api::detect_chips_silent`), which preserves the existing `UninitChip::init` loop; annotate the `Arch` match with `#[allow(deprecated)]` because `Arch::Grayskull` is now deprecated upstream and CI runs clippy with `-D warnings`; correct the stale init-error comment since `InitError::PlatformError` can occur even with an `Infallible` callback.
- `scripts/vendor_all_luwen.sh`: removed. The `luwen-core`/`-if`/`-ref`/`ttkmd-if` directories no longer exist on luwen main, so the script can no longer re-vendor against current upstream.
- `Cargo.lock`: regenerated to lock `luwen-api`/`luwen-def`/`luwen-pci`/`luwen-kmd` 0.8.5 and drop the fork crates.

The `Telemetry`, `DeviceInfo`, and `ChipDetectOptions` surfaces the reader uses are unchanged in name and signature, so device name, board type, telemetry detail fields, PCIe details, and the power/temperature/clock/utilization metrics are preserved. This is a like-for-like migration; surfacing the richer 0.8.x telemetry fields stays out of scope per the issue.

## Test plan

- [x] `cargo check --lib` resolves and fetches luwen 0.8.5, updates `Cargo.lock`, and finishes clean.
- [x] `cargo clippy --lib --tests -- -D warnings` is clean; the deprecated `Arch::Grayskull` is handled by the `#[allow(deprecated)]` annotation.
- [x] `cargo fmt --check` is clean.

Closes #265


inureyes:

### Intent
> Like-for-like migration of the Tenstorrent NPU reader from the republished `all-smi-luwen-*` / `all-smi-ttkmd-if` forks to upstream `luwen` 0.8.5 (`luwen-api` / `luwen-pci` / `luwen-def`, with `luwen-kmd` transitive), plus removal of the obsolete vendoring script. Runtime behavior is meant to be identical.

### Migration correctness (original commit 57879bc)
- [x] Imports remapped exactly per the issue's verified table: `luwen_def::Arch`; `luwen_api::{ChipDetectOptions, chip::{Chip, ChipImpl, Telemetry}}`; detection via `luwen_pci::detect_chips_silent` (the one-arg PCIe entry returning `Result<Vec<UninitChip>, _>`, not the differently-signed `luwen_api::detect_chips_silent`).
- [x] `ChipImpl` is in scope for the `chip.get_telemetry()` / `chip.get_device_info()` trait methods. `DeviceInfo` is accessed via inference on the `get_device_info()` return, exactly as the fork code did, so no import is needed.
- [x] `Arch::Grayskull` compile-time deprecation handled with `#[allow(deprecated)]`; `cargo clippy --lib --tests -- -D warnings` and `cargo fmt --check` are clean on aarch64-linux where the reader is actually compiled.
- [x] `Cargo.lock` locks `luwen-api` / `luwen-def` / `luwen-kmd` / `luwen-pci` at 0.8.5 and drops the fork crates; `luwen-kmd` is present transitively as intended.
- [x] No `all-smi-luwen-*` / `all-smi-ttkmd` references remain anywhere except the intentional historical comment in `Cargo.toml`.
- [x] `scripts/vendor_all_luwen.sh` removed; no dangling references to it.
- [x] Reader still wired through `reader_factory.rs` (gated by `has_tenstorrent()`); the `TenstorrentReader` and `get_tenstorrent_status_message` public surface is unchanged.
- [x] PR body contains `Closes #265`.

### Findings Addressed
- [x] **Latent runtime panic on Grayskull hardware (HIGH)** fixed in d1c0e78.

  `luwen_pci::detect_chips_silent` calls `start_detect`, which runs `PciDevice::scan()` over every node under `/dev/tenstorrent` and opens each one. `PciDevice::open` resolves the arch via `Arch::try_from(&GetDeviceInfoOut)`, and for a Grayskull device id (`0xfaca`) that conversion is `unimplemented!("grayskull support has been sunset")` in `luwen-kmd 0.8.5` (an unknown id returns `Err` gracefully; only Grayskull panics). The reader's `match detect_chips_silent(options) { Ok / Err }` only handles `Result`, so the panic unwound through the reader and poisoned the `INITIALIZED_CHIPS` mutex. A physically present Grayskull card (e75 / e150 / e300) therefore crashed Tenstorrent collection, whereas the previous `all-smi-luwen` fork rendered it as `Tenstorrent Grayskull ...`. This is a regression against the issue's "output unchanged on real hardware" acceptance criterion that the compile-time `#[allow(deprecated)]` did not cover.

  The fix wraps the detection call in `std::panic::catch_unwind(AssertUnwindSafe(...))`, matching the existing reader convention in `amd.rs` and `intel_gpu_engine.rs`. An unsupported device now degrades to a status message instead of unwinding and poisoning the cache lock. Verified clippy `-D warnings` clean, `fmt --check` clean, change confined to `src/device/readers/tenstorrent.rs`, and behavior unchanged for Wormhole / Blackhole / no-card hosts.

### Remaining Items
- [ ] **Release-build abort on Grayskull hosts (HIGH, partially mitigated)**: `[profile.release]` sets `panic = "abort"`, which bypasses `catch_unwind`, so a present Grayskull card still aborts in the shipping release binary (the guard fully protects dev / test / unwind builds and prevents the mutex poisoning). luwen 0.8.x removed Grayskull entirely (`unimplemented!()` throughout `luwen-kmd` / `luwen-pci` / `luwen-api`), so a complete in-process fix is not possible without either pre-filtering `/dev/tenstorrent` device ids via sysfs before invoking luwen, or relaxing `panic = "abort"` for this path. Both are outside the scope of this like-for-like migration. Recommend tracking as a follow-up (note on #265 or a new issue). Mitigating context: Grayskull is EOL / sunset legacy hardware and increasingly rare; Wormhole / Blackhole and no-hardware hosts are unaffected.

### Verification
- [x] Tests compile clean (`cargo clippy --lib --tests -- -D warnings` and `cargo fmt --check` green; full `cargo test` not run, per the narrow-scope build instruction)

inureyes:
## Follow-up: remaining HIGH resolved in-PR

The review's remaining HIGH (a Grayskull host aborts under the release `panic = "abort"` profile, where `catch_unwind` cannot recover) is now fixed in this PR rather than deferred.

**`33749e9` sysfs Grayskull pre-check.** Before invoking luwen detection (which opens every `/dev/tenstorrent` node and panics via `unimplemented!()` on the Grayskull PCI device id `0xfaca`), the reader scans sysfs for a Tenstorrent device (PCI vendor `0x1e52`) with the Grayskull device id (`0xfaca`). If one is present it sets a status message and skips luwen entirely, so no panic is reached in any build profile, including release `panic = "abort"`. The check is conservative: it returns true only on a positive vendor+device match, so Wormhole (`0x401e`) and Blackhole (`0xb140`) hosts are never affected, and any sysfs read failure falls through to normal detection. `catch_unwind` is kept as defense in depth for other unexpected panics. The device-id to arch mapping was verified against `luwen-kmd 0.8.5` (`src/lib.rs:29-32`).

**`8060647` docs.** README supported-hardware list updated to Wormhole/Blackhole only (Grayskull dropped, since luwen 0.8.x sunset it). The historical Devlog is left as-is.

**Verification** (default features, aarch64-linux): `cargo check --lib`, `cargo clippy --all-targets -- -D warnings`, and `cargo fmt --check` all clean. The optional `furiosa` feature fails to build in my local environment due to a missing libclang `stdarg.h`; that is environmental and unrelated to this PR.

**Security/perf:** the new code is a read-only scan of fixed sysfs paths with no external input (no traversal or injection surface), runs once and is cached, and degrades gracefully on any read error.

inureyes:
## PR Finalization

### Summary

**Lint/Format:** `cargo fmt --check`, `cargo clippy --lib --tests -- -D warnings`, and `cargo check --lib` all pass clean (default features only).

**Tests:** No new tests added. The `grayskull_present()` guard reads fixed sysfs paths and has no existing unit test infrastructure; the reader requires real hardware. A fabricated test would add no meaningful coverage.

**Documentation:** One concrete gap found and fixed. The man page (`docs/man/all-smi.1`) still described Tenstorrent support as "Grayskull and Wormhole architectures" after this migration dropped Grayskull. Updated to "Wormhole and Blackhole architectures" to match the README, which already reflects the correct hardware list. No other user-facing docs reference Grayskull. `docs/Devlog-Tenstorrent.md` is a historical log and was left unchanged.

**Commit pushed:** `7a585f4` docs(man): update Tenstorrent hardware list to Wormhole/Blackhole

Cyberhan123:
Please publish this version to crates.io. Thank you.

PR #268, opened 2026-07-06
fix: support Xe kernel driver hwmon temperature sensor numbering

Titaniumtown:
The Xe kernel driver starts hwmon sensor numbering at temp2 (temp1 is absent). Iterate temp1 through temp21 and return the first parseable value so both i915 (temp1 = package) and xe (temp2 = package) work without driver-specific handling.

### DIFF
--- a/src/device/readers/intel_gpu_linux/tests.rs
+++ b/src/device/readers/intel_gpu_linux/tests.rs
@@ -189,6 +189,29 @@ fn get_gpu_info_populates_basic_fields() {
 }
 
 #[test]
+fn get_gpu_info_reads_xe_temp2() {
+    let dir = tempdir().unwrap();
+    let root = dir.path();
+    let card = make_card(root, 0, "0x8086", "xe", "0xE20B");
+    let hwmon = card.join("device").join("hwmon").join("hwmon0");
+    fs::create_dir_all(&hwmon).unwrap();
+    fs::write(hwmon.join("temp2_input"), "69000\n").unwrap();
+
+    let reader = IntelGpuReader::new_from_root(root);
+    let info = reader.get_gpu_info();
+
+    assert_eq!(info.len(), 1);
+    assert_eq!(info[0].temperature, 69);
+    assert_eq!(
+        info[0]
+            .detail
+            .get("Source: Temperature")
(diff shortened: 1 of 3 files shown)


inureyes:
## Implementation hardening

### Summary

Audited and hardened PR #268 against its stated Xe/i915 hwmon compatibility goal. The kernel Xe hwmon layout reserves `temp1` as label-only and exposes the package sensor at `temp2`, so the original fallback direction is correct. Follow-up commit `9f91f19` was pushed to this PR.

- **Target PR reviewed:** https://github.com/lablup/all-smi/pull/268
- **Linked original issues reviewed:** none (the PR does not close or reference an issue)

### Review findings and fixes

- **Correctness:** the original nested loop preferred every channel in the first `hwmon*` directory before `temp1` in later directories. The reader now prioritizes lower channel numbers across all hwmon directories, preserving the documented i915-before-Xe fallback semantics.
- **Reliability:** milli-Celsius values larger than `u32` can represent no longer wrap to a small/zero value before the caller's validation cap. They now saturate to `u32::MAX` so the existing cap remains effective.
- **Coverage:** added regression tests for Xe `temp2`, cross-directory channel priority, the full `temp1..=temp21` range, overflow saturation, and end-to-end `GpuInfo`/metric-source decoration.
- **Portability:** portable sysfs helper tests now run on non-Linux development hosts under `cfg(test)`; the Linux reader integration test remains covered by Linux CI.
- **Security review:** no authentication, authorization, network, command-execution, or user-controlled path surface is involved. Production input is kernel-owned sysfs, and the channel scan remains bounded to 21 entries.
- **Performance/reliability review:** the implementation enumerates hwmon directories once, reuses the collected paths, and performs a bounded scan; no unbounded polling, allocation, retry, or background work was introduced.

### Validation

Passed locally:

- `cargo fmt --check`
- `cargo test intel_gpu_sysfs::tests -- --nocapture`
- `cargo test --all`
- `cargo clippy -- -D warnings`
- `cargo build --release --bin all-smi`
- `cargo build --no-default-features --lib`

Linux-specific integration coverage and merge-result validation are delegated to this PR's CI checks.

PR #269, opened 2026-07-06
feat: compute Xe GPU utilization from gtidle/idle_residency_ms

Titaniumtown:
The Xe kernel driver does not expose engine busy counters via sysfs (unlike i915), so utilization was permanently 0 on Xe GPUs. The Xe driver does expose gtidle/idle_residency_ms. We compute utilization from the idle-time delta across scrape intervals, matching the semantics of the existing engine-busy delta tracker.

- Add read_gtidle_ms() to intel_gpu_sysfs.rs
- Add GtidleState delta tracker
- Fall back to gtidle in get_gpu_info() when engine busy is unavailable
- Override detail[Utilization] and Source: Utilization label when gtidle provides data after apply_engine_readout sets the unavailable note

### DIFF
--- a/src/device/readers/intel_gpu_gtidle.rs
+++ b/src/device/readers/intel_gpu_gtidle.rs
@@ -1,0 +1,331 @@
+// Copyright 2025 Lablup Inc. and Jeongkyu Shin
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Xe GT idle-residency utilization fallback.
+//!
+//! Xe does not expose the i915-style engine-busy sysfs counters on every
+//! kernel. In that case, the time a GT spends outside its deepest idle state
+//! is used as an activity estimate. It is deliberately treated as a fallback:
+//! active residency is not the same as engine busy time and must not replace a
+//! valid zero-percent engine-busy sample.
+
+use std::collections::HashMap;
+use std::path::Path;
+use std::sync::Mutex;
+use std::time::Instant;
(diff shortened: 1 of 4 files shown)


inureyes:
## Implementation hardening

**Target reviewed:** PR #269 (`feat: compute Xe GPU utilization from gtidle/idle_residency_ms`)

**Linked original issues:** None were referenced by the PR body, comments, or `closingIssuesReferences`.

### Findings and fixes

- Restricted the gtidle path to Xe cards whose DRM engine counters are actually unavailable. The original `utilization == 0.0` condition also replaced valid zero-percent engine-counter samples.
- Represented unavailable, seeded, and available gtidle states separately. A live zero-percent gtidle sample now receives the correct source label instead of remaining marked unavailable.
- Preserved `gt0`/`gt1` identity across partial sysfs read failures and retained a per-GT timestamp, preventing one GT from being compared with another GT's prior counter or a missed interval from using the wrong wall time.
- Reseeded a counter after driver/device resets instead of converting a backwards counter into a false 100-percent activity spike.
- Moved the tracker into a focused bounded module and added regression tests for fallback selection, live zero-percent samples, exact delta computation, resets, and missing GT files.

### Security and reliability review

No authentication, authorization, input-injection, secret-handling, or unsafe-code surface is introduced by this hardware sysfs reader. The fallback performs at most two bounded sysfs reads per Xe card and only when engine counters are unavailable; it does not add unbounded allocation, retries, polling, or background tasks. Mutex poisoning now resets the local tracker and continues serving telemetry.

### Validation

- `cargo fmt --check` — passed.
- `cargo test` — passed on the PR head: 1,169 library tests plus all integration and doc tests.
- Merge simulation against current `origin/main`, followed by `cargo test` — passed: 1,021 library tests, 1,187 binary tests, and all integration/doc tests.
- Linux Docker `cargo test --lib intel_gpu_gtidle` — 6 passed.
- Linux Docker `cargo test --lib intel_gpu_sysfs` — 12 passed.
- `cargo clippy -- -D warnings` on macOS — passed.
- Linux Docker `cargo clippy --lib -- -D warnings` on Rust 1.97 reached the project crate and reported only three pre-existing `clippy::useless_borrows_in_formatting` findings in `src/device/readers/google_tpu.rs`; no finding was emitted for this PR's code. Repository CI remains the authoritative Linux lint run.

Hardening commit: `cd929b0e8fe40a25904a83e6b172d8494b19006c`.

PR #283, opened 2026-07-20
fix: report VRAM and power for Battlemage GPUs on the xe driver

Titaniumtown:
Fixes reporting of VRAM usage due to battlemage reporting it under a different path.

Also fixes reporting of power draw because intel only reports total energy (in Joules), so power draw has to be derived.

### DIFF
--- a/src/device/readers/intel_gpu_fdinfo.rs
+++ b/src/device/readers/intel_gpu_fdinfo.rs
@@ -474,6 +474,81 @@ pub fn collect_intel_gpu_processes(
     out
 }
 
+/// Sum `drm-resident-vram0` bytes across all processes for a given
+/// card. Used as a fallback when the xe driver does not expose
+/// card-level VRAM usage counters in sysfs.
+///
+/// Deduplication: when a process holds multiple fds pointing at the
+/// same DRM client (same `drm-client-id`), the VRAM block is counted
+/// only once (max-wins). Distinct client ids within the same process
+/// do sum.
+pub fn sum_card_vram_from_fdinfo(
+    card_index: usize,
+    intel_drm_basenames: &HashMap<String, usize>,
+    proc_root: &Path,
+) -> u64 {
+    if intel_drm_basenames.is_empty() {
+        return 0;
+    }
+    let Ok(entries) = std::fs::read_dir(proc_root) else {
+        return 0;
+    };
+
+    let mut seen: HashMap<(u32, u64), u64> = HashMap::new();
+
+    for entry in entries.flatten() {
+        let path = entry.path();
+        let Some(name) = path.file_name().and_then(|n| n.to_str()) else {
+            continue;
+        };
+        let Ok(pid) = name.parse::<u32>() else {
+            continue;
+        };
+
+        let fds = intel_drm_fds_for_pid(pid, intel_drm_basenames, proc_root);
+        for fd in fds {
+            if fd.card_index != card_index {
+                continue;
+            }
+            let Some(content) = read_fdinfo_to_string(&fd.fdinfo_path) else {
+                continue;
+            };
+            let mut vram_bytes: u64 = 0;
+            let mut drm_client_id: Option<u64> = None;
+            let mut is_drm = false;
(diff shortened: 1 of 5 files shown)


inureyes:

Reviewed both commits on this branch (VRAM fallbacks and energy-derived power) and pushed fixes in 0a0c31e. All findings were verified against the code before changing anything.

**High, security: energy cache in /tmp was symlink-attackable and world-poisonable.** `/tmp/all-smi-energy-<pci>` is a predictable name in a world-writable directory, and `std::fs::write` follows symlinks. A local user could pre-plant a symlink for a root-run all-smi to clobber an arbitrary file, or pre-create the cache to skew reported power. The cache now lives in the per-user cache directory (`~/.cache/all-smi/energy-hwmon-<pci>`, the same location the energy WAL uses), writes go through the shared hardened writer in `common::secure_write` (O_NOFOLLOW, mode 0600, atomic rename), and reads refuse symlinks, mirroring the WAL replay hardening. When no home-like directory exists the reader degrades to in-memory state only.

**Medium: derived power skipped the validation cap.** The energy-derived wattage was assigned after the `MAX_GPU_POWER_WATTS` clamp, so a stale or corrupted cache (the file path trusts wall-clock timestamps) could emit absurd values. The derived value now goes back through the clamp. Cache timestamps also gained millisecond resolution, and writes are throttled to one per 60 s so a 1 s sampling loop does not fsync the file on every tick.

**Medium: BAR2 heuristic misclassified integrated GPUs.** `resource2 >= 256 MiB` flips every Intel iGPU to Discrete, because iGPUs expose BAR2 as the fixed 256 MiB GMADR aperture; they would have gained a fabricated 256 MiB VRAM total and lost the shared-memory note. The probe is now gated to the xe driver and requires the BAR to be strictly larger than 256 MiB. One side effect worth knowing: a Battlemage card without Resizable BAR also shows a 256 MiB BAR2, so it keeps reporting no VRAM total instead of a wrong 256 MiB one; the fallback works when ReBAR maps the whole VRAM, which is the recommended configuration for these cards anyway.

**Low, performance: per-card /proc re-walk.** `sum_card_vram_from_fdinfo` walked all of /proc once per card per refresh. Replaced with `sum_vram_by_card_from_fdinfo`: a single walk per refresh shared by all cards, computed lazily only when some xe card actually needs the fallback.

**Low, correctness: the fallback summer diverged from the existing fdinfo pipeline.** The hand-rolled parse loop accepted any `drm-driver`, collapsed fds without `drm-client-id` into one max-wins bucket (undercounting multi-context processes), double counted fork-inherited clients across pids, had no `MAX_GPU_PROCESSES` cap, and summed without saturation. The new implementation reuses `parse_fdinfo` (which rejects foreign drivers) via a new `resident_vram_bytes` field, dedupes by client id per card, credits no-client-id fds per fdinfo path, honours the cap, and uses saturating arithmetic.

**Low: provenance labels.** `Source: Memory` said "DRM sysfs" when the total actually came from the BAR2 size; it now reports "PCI BAR2 size (xe)" and "fdinfo used, PCI BAR2 total (xe)" as appropriate.

Validation: cargo fmt, cargo clippy --all-targets -- -D warnings, and the full cargo test suite pass. 21 new unit tests cover the BAR2 threshold, classify_variant gating, energy cache symlink behaviour, power derivation (seeding, file-cache bridge, stale cache), and the per-card VRAM sum.

Behavioural note for snapshot users: the one-shot power bridging now requires a resolvable per-user cache directory; environments without HOME report 0 W on the first sample, matching the energy WAL downgrade behaviour.


**Addendum:** commit 0a0c31e broke the library-only CI build (`cargo build --no-default-features --lib`) by importing `common::paths` / `common::secure_write` unconditionally; both modules are gated behind the `cli` feature for the optional `dirs` dependency. Fixed in 451d7ce: the energy file cache (a CLI snapshot-only concern) is now `cfg(feature = "cli")` with no-op stubs for library builds, so library consumers keep the fully functional in-memory power delta. Both build configurations, clippy, and the test suite pass in both feature sets.

Issue #288, opened 2026-07-27
test: validate the local-mode interval change on non-Apple-Silicon platforms


inureyes:
## Problem / Background

PR #286 (merged as commit 18fb595) changed local mode's default collection interval. `run_local_mode` in `src/view/data_collector.rs` used to call `EnvConfig::adaptive_interval(1)`, where the argument `1` reads as "one host" but actually selects the `1..=10 => 3` remote-nodes arm, so local mode polled every 3 seconds. It now calls `EnvConfig::local_interval()`, which is `adaptive_interval(0)`, the arm documented as "Local monitoring only (no remote nodes)".

`adaptive_interval(0)` in `src/common/config.rs` returns 1 on macOS Apple Silicon and 2 everywhere else. The effective default therefore changed from 3s to 1s on Apple Silicon, and from 3s to 2s on Linux, Windows, and Intel Mac. On the non-Apple-Silicon platforms that is a 1.5x increase in polling frequency.

The reason the higher polling rate was considered affordable is that PR #286 also replaced the IOReport sampling strategy with a continuous delta against the previous poll's retained sample, which cut a single collection from about 495ms to about 24ms. That part of the change is gated to macOS: the `macos_native` module is declared `#[cfg(target_os = "macos")]` in `src/device/mod.rs`, and IOReport is an Apple Silicon API. Non-Apple-Silicon platforms consequently received the 1.5x polling increase without any offsetting collection cost reduction, because they never had the 495ms blocking call in the first place. This issue tracks measuring that gap instead of assuming it away.

## What was measured

All benchmarking for PR #286 was done on a single machine: an Apple M1 Ultra, release build, in a 200x50 tmux session, sampling `ps -o %cpu=` every 2 seconds for 60 seconds per configuration. Full-TUI CPU before the change was 8.16% at 1s, 5.32% at 2s, and 4.06% at 3s; after the change it was 4.71% at 1s, 3.47% at 2s, and 2.13% at 3s. Default local mode with no `-i` flag measured 4.06% before (3s interval) and 3.52% after (1s interval), so on that one machine the default became both faster and cheaper at the same time.

[...]

PR #291, opened 2026-07-27
test: record core topology and support CPU pinning in the benchmark

### LINKED ISSUE #290: test: bench-local-interval.sh CPU-time metric is ambiguous on heterogeneous CPUs
## Problem / Background

`scripts/bench-local-interval.sh` (added by PR #289, tracking issue #288) reports CPU as "percent of one core", derived from the process CPU-time delta over the measurement window. The script's stated purpose is that results from different machines are comparable: it fixes the terminal at 200x50 because render cost scales with terminal size, and it prints an environment block because collection cost depends on which device readers are active.

On a heterogeneous CPU that comparability does not hold, because "one core" is not one quantity. The same work costs a different amount of CPU time depending on which core cluster the scheduler places the threads on, and nothing in the output records which cluster that was.

## Measurement

Measured on an NVIDIA GB10 (DGX Spark), Ubuntu 24.04.4, kernel 6.17.0-1026-nvidia, aarch64, all-smi 0.24.2 release build. The part is heterogeneous: CPUs 5-9 and 15-19 are Cortex-X925 at 3900 MHz, CPUs 0-4 and 10-14 are Cortex-A725 at 2808 MHz.

Same binary, same interval, same 60s window after 8s warmup, pinned with `taskset` to one cluster at a time by wrapping the binary and passing it via the script's own `-b` flag, so the measurement code path is identical:

| config | perf (X925) | eff (A725) | ratio |
|---|---|---|---|
| default | 1.13% | 1.67% | 1.48x |
| `-i 2s` | 1.12% | 1.70% | 1.52x |
| `-i 3s` | 0.95% | 1.40% | 1.47x |

Placement alone moves the headline number by about 1.5x. Unpinned runs land between the two pinned values, as expected for a thread mix.

For scale: the effect the benchmark exists to measure, the 3s to 2s default step that non-Apple-Silicon platforms absorbed in PR #286, is +19.4% on this host. The placement artifact is larger than the signal.

## Why this is not GB10-specific

Any hybrid part has this property: Intel P/E-core desktops and servers, ARM big.LITTLE, and Apple Silicon P/E cores. The Apple M5 Max reference run in PR #289 carries the same ambiguity, so the macOS and Linux numbers currently in issues #286, #287, #288, and PR #289 are all affected to an unknown degree.

Run-to-run variance is the visible symptom. Three unpinned repeats on this host gave a standard deviation of roughly 0.08 to 0.13 percentage points per configuration, against a 3s to 2s effect of 0.233 percentage points, so the effect is only about 2x the noise. Pinning collapses that variance: pinned `default` and pinned `-i 2s`, which should be identical because local mode resolves to 2s on Linux, agreed to within 0.9% on the performance cluster and 1.8% on the efficiency cluster.

## What holds up

The relative step is robust even though the absolute value is not. The 3s to 2s step measured +17.9% pinned to performance cores, +21.4% pinned to efficiency cores, and +19.4% unpinned. Conclusions drawn from ratios within a single host are therefore still sound. What is not sound is comparing one host's absolute percentage against another's.

## Proposed Solution

1. Record core topology in the environment block, not just the core count: which clusters exist and how many of each. The `lscpu` fallback added for aarch64 already surfaces cluster names (it prints `Cortex-X925 + Cortex-A725` on this host), so the information is partly there already.
2. Offer an optional affinity flag so a reporter can pin the run to one cluster and say which, for example `-c 5-9,15-19`. This both removes the artifact and cuts run-to-run variance substantially.
3. State in the script header and in the DEVELOPERS.md section that absolute percentages are comparable across machines only when core placement is stated, and that the ratio between intervals on one host is the durable quantity.
4. Consider defaulting to more than one repeat per configuration and reporting a mean, since on low-cost hosts a single 60s window puts the measured effect close to the noise floor.

## Acceptance Criteria

- [ ] The environment block records core topology sufficient to tell a heterogeneous host from a uniform one.
- [ ] A reporter can pin the measurement to a specific set of cores and that choice appears in the output.
- [ ] Documentation states the comparability limit explicitly.
- [ ] No change to what `all-smi` itself collects or reports; this is benchmark tooling only.

## References

- PR #289: adds `scripts/bench-local-interval.sh`.
- Issue #288: asks for non-Apple-Silicon measurements, which is where this was found.
- PR #286: the interval change being measured.

inureyes:
Closes #290.

## Why

`scripts/bench-local-interval.sh` reports CPU as percent of one core, and the whole point of its design (fixed 200x50 terminal, environment block, CPU-time delta instead of `ps -o %cpu`) is that results from different machines are comparable. On a heterogeneous CPU that does not hold, because "one core" is not one quantity. Identical work costs a different amount of CPU time on a performance core than on an efficiency core, and nothing in the output recorded which one ran.

Measured on an NVIDIA GB10 (Cortex-X925 at 3.9GHz plus Cortex-A725 at 2.8GHz), pinning the same run to one cluster or the other moves the result by about 1.5x. That is larger than the interval effect the script exists to measure, which was +19.4% on the same host. Intel P/E hybrids and Apple Silicon have the same property, so the existing M5 Max reference numbers carry it too.

## Changes

**`topology` line in the environment block.** Groups logical CPUs by maximum frequency, which separates ARM big.LITTLE clusters and Intel P/E cores alike, falling back to the device-tree `cpu_capacity` on ARM systems without cpufreq, and to `hw.nperflevels` on macOS. When none is readable it reports `unknown` rather than assuming uniform, since assuming uniform is the specific error this is meant to prevent. Uniform hosts print just count and speed; heterogeneous hosts additionally print each cluster's CPU list, range-collapsed so it can be pasted straight into the new flag.

**`-c CPUS` pins the run** via `taskset`, and the choice appears on a new `affinity` line. An unpinned run on a heterogeneous host now says so explicitly instead of quietly reporting a placement-dependent number. Rejected on macOS with an explanation rather than silently ignored, because the kernel exposes no userspace CPU affinity control and a run genuinely cannot be pinned to P or E cores there. Invalid lists are validated by handing them to `taskset` itself rather than reimplementing range syntax.

**`-r COUNT` repeats each configuration** and reports mean and standard deviation, because a single window on a low-cost host puts the effect close to the run-to-run spread. Default stays 1, so existing invocations keep both their runtime and their exact output format; the mean-and-deviation format appears only when repeats are requested. `measure` was split into `measure_once` plus an aggregating wrapper, with failures carried through exit codes so a partial run reports how many windows succeeded rather than silently averaging fewer.

## Output

```
  cpu           Cortex-X925 + Cortex-A725 (20 cores)
  topology      heterogeneous: 10x 3.90GHz (cpus 5-9,15-19), 10x 2.81GHz (cpus 0-4,10-14)
  affinity      cpus 0-4,10-14 (taskset)
  window        20s measured after 8s warmup, 2 repeats

=== results (percent of one core) ===
  default        cpu=  1.58% +/- 0.04 (n=2)   cpu_time=0.32s / 20s   rss=31MB
  -i 2s          cpu=  1.58% +/- 0.04 (n=2)   cpu_time=0.32s / 20s   rss=31MB
  -i 3s          cpu=  1.40% +/- 0.07 (n=2)   cpu_time=0.28s / 20s   rss=30MB
```

## Review findings, all fixed in later commits

Implementation and security review each found real defects, every one reproduced on hardware before being fixed.

**The affinity line could lie.** `taskset` does not validate a CPU list the way the first commit's comment claimed: it fails only when *no* CPU in the list exists, so it rejects `99-200` but silently accepts `0,99` and narrows it to CPU 0. The line then advertised `cpus 0,99` for a run pinned to one core, which is worse than printing nothing, since this line exists precisely so the context travels with the number. The effective mask is now read back after launch and reported, with the requested list shown alongside when they differ.

**Exact-key grouping over-split real hardware.** This machine's `cpu_capacity` reads 718, 731, 997, 1017, and 1024 across a two-cluster part, so the fallback reported five tiers, one of them a single CPU. Intel parts with per-core turbo binning have the same shape on the cpufreq path. Keys are now bucketed within 5% of the group's fastest member, compared against that fixed representative so small steps cannot drift a bucket across a real boundary, and the range collapser sorts numerically since a bucket can merge keys whose CPU runs interleave.

**The measured PID was a race.** It came from diffing `pgrep -x all-smi` snapshots and taking the first new PID. `-c` made that likely rather than theoretical, since the natural way to use it is one pinned run per cluster: running both at once made them report byte-identical numbers, cpu_time included, while each `affinity` line attested to a different placement. The PID now comes from `tmux list-panes -F '#{pane_pid}'`, the process tmux actually started. Two concurrent pinned runs now report distinct numbers in the expected direction.

**Nothing cleaned up on interrupt**, so Ctrl-C left an orphan rendering a 200x50 TUI forever, competing for CPU with the next run and, under `-c`, for exactly the pinned cluster. A trap now kills this invocation's sessions. The INT and TERM handlers exit rather than only cleaning up, because a bash trap returns to the point of interruption, so cleanup alone killed the window in flight and then opened the next one.

**Unvalidated input reached the shell command string.** `-i '$(touch /tmp/x)1'` created the file and a bare `-i '*'` glob-expanded so filenames became intervals; `-d 0.5` was accepted silently while being shorter than the clock resolution, understating CPU roughly twofold. Both are whole seconds by definition and are now checked, with globbing off for the interval walk.

Also: tmux stderr is surfaced on launch failure instead of a bare "could not start", and the window is timed with bash 5's `EPOCHREALTIME` where available, since whole-second truncation at both ends puts up to 1.7% of error on a 60s window, landing directly on the spread `-r` exists to expose.

## Tests

Validated on the GB10 host across unpinned, performance-pinned, and efficiency-pinned runs at 8s, 10s, 20s, and 30s windows with 1 and 2 repeats.

Pinning demonstrably does what it is for. Pinned to the efficiency cluster, `default` and `-i 2s` agreed **exactly** at 1.58%, where they should be identical because local mode resolves to 2s on Linux, and the deviation tightened to about 0.04 from the 0.08 to 0.13 measured unpinned.

Topology detection was exercised against the shipped awk for uniform x86, Intel P/E (8P+16E), Intel favoured-core binning, the real ARM capacity values, single core, non-contiguous clusters, three genuine tiers, and CPU indices past 9. The 5% bucket merges binned siblings without merging real tiers.

Interrupt handling was tested through a real terminal Ctrl-C rather than a background kill: bash sets SIGINT to ignored for asynchronous jobs and a signal ignored at entry cannot be trapped, which made the first test runs look like a script bug when the harness was at fault. Under a genuine Ctrl-C the session and child both go away.

Flag validation is covered: `-r 0`, `-r abc`, `-d abc`, `-d 0`, `-i '$(touch ...)'`, `-i '1;id'`, `-i '*'`, and an out-of-range `-c 99-200` are each rejected with a specific message.

`shellcheck` 0.11.0 is clean at default severity. The `-r 1` success-line format string is byte-identical to the pre-PR version, so existing reports stay comparable.

Shell only, no Rust touched, so the existing suite is unaffected and nothing about what `all-smi` collects or reports changes, satisfying #290's fourth acceptance criterion.

Also documents the comparability limit, all flags, and the container caveat for `-c` under Testing in `DEVELOPERS.md`, along with the `libdrm-dev` build prerequisite on Linux.

### DIFF
--- a/DEVELOPERS.md
+++ b/DEVELOPERS.md
@@ -247,7 +247,7 @@ The script prints an environment block (OS, CPU, core topology, affinity, GPU, c
 
 So the durable number is the **ratio between two intervals measured on one host**, not the absolute percentage. Compare absolute percentages across machines only when both reports state their core placement. The `topology` and `affinity` lines in the environment block carry that context automatically.
 
-On a heterogeneous host, prefer `-c` to pin one cluster and say which one you pinned; it also cuts run-to-run variance substantially. Use `-r` to average several windows when the effect you are chasing is close to the spread between repeats. `-c` is Linux only, since macOS exposes no userspace CPU affinity control and a run cannot be pinned to P or E cores there. Background and measurements are in issue #290.
+On a heterogeneous host, prefer `-c` to pin one cluster and say which one you pinned; it also cuts run-to-run variance substantially. Use `-r` to average several windows when the effect you are chasing is close to the spread between repeats. Two limits on `-c`: it is Linux only, since macOS offers no way to select a specific CPU set (only E-core confinement is reachable at all, via `taskpolicy -b`, which also changes scheduling priority); and it should be run on bare metal, because inside a container without a cpuset `all-smi` sizes its own CPU view from `sched_getaffinity`, so pinning would shrink the set of cores it parses and renders rather than only relocating the same work. Background and measurements are in issue #290.
 
 Coverage is currently thin outside Apple Silicon. See issue #288 if you have Linux or Windows hardware and can contribute measurements.
 
--- a/DEVELOPERS.md
+++ b/DEVELOPERS.md
@@ -232,16 +232,17 @@ cargo build --release --bin all-smi
 scripts/bench-local-interval.sh                    # default: 60s window, intervals 1/2/3
 scripts/bench-local-interval.sh -d 120 -i "2 5"    # longer window, custom intervals
 scripts/bench-local-interval.sh -c 5-9,15-19 -r 3  # pin to one core cluster, average 3 windows
-scripts/bench-local-interval.sh -h                 # usage
+scripts/bench-local-interval.sh -b ~/bin/all-smi   # measure a binary built elsewhere
+scripts/bench-local-interval.sh -h                 # full usage and flag list
 ```
 
 It requires `tmux` (macOS and Linux only) so the TUI runs detached at a fixed 200x50 size; terminal size affects render cost, so fixing it is what makes results from different machines comparable. CPU is computed from the process CPU-time delta rather than `ps -o %cpu`, because that column is a decaying recent average on macOS and a lifetime average on Linux. Results are percent of one core.
 
 On Linux, building needs `libdrm-dev`. Without it the link fails on `-ldrm` and `-ldrm_amdgpu` even on a host with no AMD GPU, because `libamdgpu_top` is a hard dependency of the glibc Linux target.
 
-The script prints an environment block (OS, CPU, core topology, affinity, GPU, core count, process count) above the numbers. Include it whenever you report results: collection cost depends heavily on which device readers are active, so numbers without that context cannot be compared.
+The script prints an environment block (all-smi version, OS, CPU model and core topology, affinity, GPU, process count, terminal size, window length) above the numbers. Include it whenever you report results: collection cost depends heavily on which device readers are active, so numbers without that context cannot be compared.
 
-#### Comparing results across machines
+#### Comparing Results Across Machines
 
 "Percent of one core" is not a single quantity on a heterogeneous CPU. ARM big.LITTLE parts, Intel P/E hybrids, and Apple Silicon all mix core types, and identical work costs a different amount of CPU time on each type. On an NVIDIA GB10 (Cortex-X925 at 3.9 GHz plus Cortex-A725 at 2.8 GHz), pinning the same run to one cluster or the other moves the result by about 1.5x, which is larger than the interval effect the script is usually being used to measure. Unpinned runs land somewhere between the two depending on where the scheduler put the threads, which is also why they vary more between repeats.
 
--- a/DEVELOPERS.md
+++ b/DEVELOPERS.md
@@ -229,14 +229,25 @@ For comprehensive testing documentation, see [TESTING.md](TESTING.md).
 
 ```bash
 cargo build --release --bin all-smi
-scripts/bench-local-interval.sh                 # default: 60s window, intervals 1/2/3
-scripts/bench-local-interval.sh -d 120 -i "2 5" # longer window, custom intervals
-scripts/bench-local-interval.sh -h              # usage
+scripts/bench-local-interval.sh                    # default: 60s window, intervals 1/2/3
+scripts/bench-local-interval.sh -d 120 -i "2 5"    # longer window, custom intervals
+scripts/bench-local-interval.sh -c 5-9,15-19 -r 3  # pin to one core cluster, average 3 windows
+scripts/bench-local-interval.sh -h                 # usage
 ```
 
 It requires `tmux` (macOS and Linux only) so the TUI runs detached at a fixed 200x50 size; terminal size affects render cost, so fixing it is what makes results from different machines comparable. CPU is computed from the process CPU-time delta rather than `ps -o %cpu`, because that column is a decaying recent average on macOS and a lifetime average on Linux. Results are percent of one core.
 
-The script prints an environment block (OS, CPU, GPU, core count, process count) above the numbers. Include it whenever you report results: collection cost depends heavily on which device readers are active, so numbers without that context cannot be compared.
+On Linux, building needs `libdrm-dev`. Without it the link fails on `-ldrm` and `-ldrm_amdgpu` even on a host with no AMD GPU, because `libamdgpu_top` is a hard dependency of the glibc Linux target.
+
+The script prints an environment block (OS, CPU, core topology, affinity, GPU, core count, process count) above the numbers. Include it whenever you report results: collection cost depends heavily on which device readers are active, so numbers without that context cannot be compared.
+
+#### Comparing results across machines
+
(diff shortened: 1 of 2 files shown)


inureyes:
## Finalization pass

### Documentation

`-h` was missing from the script's own help text. `usage()` renders the header comment block, so the header is the help text, and it documented `-d`, `-b`, `-i`, `-c`, and `-r` but not the flag that prints it. Added to both the synopsis and the flag list. The `-d` and `-i` entries now state whole seconds, matching the validation that rejects fractional values.

In DEVELOPERS.md, `-b` had no example, which also left "measure a binary that is not at the default relative path" undocumented. Added one. The environment block description now matches what the block prints: the previous list omitted the all-smi version, the terminal size, and the window length, and listed a core count line that does not exist as such (the count is part of the `cpu` line). Retitled `#### Comparing results across machines` to Title Case, which is what the rest of the file uses.

Checked the remaining claims in the section against the script and they hold: 60s default window, default intervals 1/2/3, fixed 200x50 tmux size, CPU-time delta rather than `ps -o %cpu`, the `libdrm-dev` link requirement, the macOS `-c` limitation, and the container caveat. Issue references are correct and both targets are open: results go to #288, background to #290.

### Translations

No translated copy of DEVELOPERS.md exists. The only Korean documents in the repository are per-report `TECHNICAL_REPORTS/*.ko.md` pairs, none of which cover this work, so there was nothing to mirror.

### Tests

No Rust changed, and #290 requires that nothing about what all-smi collects or reports changes, so no Rust tests were added. Validation for the shell harness:

- `shellcheck` 0.11.0 clean at default severity, `bash -n` clean.
- `-h` renders correctly, and exits before the tmux and binary checks, so it works on a host with neither.
- Short run on the GB10 host (`-d 1 -i ""`) confirming the environment block prints what the documentation now says it prints, with no orphan tmux session left behind.
- Style-only `-o all` suggestions (SC2250, SC2292, SC2249, SC2310, SC2312) deliberately not applied: the file does not follow those conventions elsewhere.

### Housekeeping

Working tree clean apart from the two intended files. No stray artifacts, no temp files, no wrong issue numbers. No AI attribution and no em dashes in the diff, the commit messages, or the PR body.

PR #299, opened 2026-07-31
perf: parallelize the local collection pipeline (#287)

### LINKED ISSUE #287: perf: parallelize the local collection pipeline instead of serializing sync readers inside tokio::join!
## Problem / Background

The local collection pipeline in `src/view/data_collection/local_collector.rs` is written as if it collects every metric source concurrently, but it does not. The `tokio::join!` block that starts around line 278 in `collect_parallel_first_iteration` looks like a fan-out across GPU, CPU, memory, GPU processes, full processes, storage, chassis, vGPU, and MIG, and reads like nine independent collectors running side by side. In practice almost all of that work is serialized onto a single task, so the join provides ordering and result aggregation but no actual overlap.

## Current behavior

`tokio::join!` polls all of its futures on the SAME task. It does not spawn anything. Concurrency only materializes when a future yields at an await point, and the reader arms never yield inside the work itself: `reader.get_gpu_info()`, `reader.get_cpu_info()`, `reader.get_memory_info()`, and `reader.get_gpu_processes()` are synchronous calls, and the only awaits in those arms are lock acquisition (`readers.read().await`) and status channel sends. The same is true of the storage, chassis, vGPU, and MIG arms. The result is that these arms execute one after another, in source order, on one worker thread.

The single exception is the full process collection arm, which genuinely runs off-task: it wraps its work in `tokio::task::spawn_blocking` (around line 327-329, with the comment "use spawn_blocking to avoid blocking tokio runtime"). That arm is the only one that can overlap with anything else.

The runtime this happens on is multi-threaded with `worker_threads(4)`, built by the default runtime builder around line 171 in `src/main.rs`. So there are idle workers available while the collection arms take turns on one of them.

## Proposed change

Parallelize the pipeline for real. Either move all synchronous reader work off the async task so the join actually overlaps its arms, or restructure the collection so it stops presenting itself as async when it is not. The point is the pipeline as a whole, not any single arm.

A partial fix is explicitly not the goal here. Moving only the GPU arm to `spawn_blocking` buys almost nothing, because every remaining synchronous arm still serializes behind the others: the critical path is the sum of all the synchronous arms, and removing one term from that sum leaves the rest untouched. Any design proposed for this issue should account for all of the synchronous arms together.

It is also worth being precise about what `spawn_blocking` does: it relocates work, it does not reduce it. The same CPU cycles run on the blocking pool instead of a worker thread. The win is worker availability and wall clock overlap, not total CPU.

The implementation cost is what makes this non trivial. The reader collections are held in `Arc<RwLock<...>>` using tokio's async `RwLock`, so `.read().await` cannot be used inside a `spawn_blocking` closure. Moving an arm off-task requires switching to `blocking_read()` or restructuring how the readers are borrowed, and it requires satisfying `'static + Send` bounds for the closure. That is a real refactor of ownership and locking across every reader, not a one line change.

## Why this is deferred

PR #286 already captured the large win. Before it, the GPU arm blocked its worker for about 495ms per collection because IOReport opened a synthetic sampling window (sample, sleep 100ms, sample, repeated four times and averaged). PR #286 replaced that with a continuous delta against the previous poll's retained sample, which cut collection to about 24ms measured on an M1 Ultra release build.

With that in place, the remaining measurable benefit is below noise at current intervals. Moving the GPU arm off-task now would take worker occupancy at a 1 second interval from 33.0% to 2.4% of one worker, which is roughly 0.6% of a 4 worker runtime, and the potential wall clock saving is about 24ms out of a 1000ms cycle. That does not justify a locking refactor on its own.

One blocking call does remain after PR #286: the first collection of a session still pays a single blocking `sample_interval_ms` (100ms) window, because there is no previous sample to delta against. This happens once at startup while the loading state is displayed, so it is not on any steady state path.

Revisit this issue if a reader is added whose synchronous cost is significant, or if the collection interval is lowered further. Either change makes the serialized sum matter again.

## Acceptance criteria

- [ ] Measure the current total synchronous portion of the join (per arm and summed) before changing any code, so the baseline is recorded rather than assumed.
- [ ] Produce a design that moves all synchronous reader work off the async task, not just one arm, including how the `Arc<RwLock<...>>` reader borrows and the `'static + Send` closure bounds are satisfied.
- [ ] Report before/after wall clock numbers for one full collection cycle on the same machine and build profile.
- [ ] Show no CPU regression: total CPU per collection cycle must not increase, given that `spawn_blocking` relocates work rather than removing it.
- [ ] Show no change to collected metric values: GPU, CPU, memory, process, storage, chassis, vGPU, and MIG output must match the pre-change values.

## References

- PR #286: IOReport continuous delta sampling and the local mode interval fix.
- PR #285: history graph scrolling window.

inureyes:
## Summary

The `tokio::join!` in `collect_parallel_first_iteration` fanned out over nine arms but polled all of them on one task. Every arm except full-process collection called synchronous reader methods with no yield point, so the join only provided ordering and result aggregation: the arms took turns on a single worker and the cycle cost the sum of all of them. `collect_sequential`, the steady-state path used for every cycle after the first, was serial by construction and had exactly the same problem, so both paths are fixed here.

Both now dispatch every synchronous reader pass to the blocking pool up front and join only at the end. Nothing synchronous is left on the async task in either path.

## Design

Work is grouped per reader collection, not per query. GPU device info, GPU processes, vGPU and MIG stay back to back inside one task because they all run against the same `Box<dyn GpuReader>` instances, and those carry internal sampling state (IOReport deltas on Apple Silicon, cached NVML handles on NVIDIA); issuing them concurrently against one reader would change the order in which that state is touched, which is exactly what acceptance criterion 5 forbids. CPU, memory, chassis, storage and full-process collection each get their own task, giving six groups that genuinely overlap.

**How the `Arc<RwLock<...>>` borrows are handled.** `spawn_reader_pass` takes the guard with `Arc<RwLock<T>>::read_owned()` in async context and moves it into the closure. This is what makes the refactor possible at all: `read()` is a future and unusable inside a blocking closure, while `blocking_read()` from the blocking pool would risk deadlocking against a queued async writer (tokio's `RwLock` is fair, so a queued writer blocks subsequent readers). `OwnedRwLockReadGuard<T>` is `'static + Send` whenever `T: Send + Sync`, and `GpuReader`/`CpuReader`/`MemoryReader`/`ChassisReader` are all declared `Send + Sync`, so the `spawn_blocking` bounds are satisfied without cloning or re-owning a single reader. No guard is ever held across an `.await`.

**Why holding the guard for the task's lifetime is safe.** I checked every writer of these locks: the only one is `initialize_readers`, which runs once and completes before the first collection. There is no hotplug or periodic-refresh path that takes a write lock while a collection is in flight, so a long-lived read guard cannot starve anything.

**The one dependency that is kept.** In the steady-state path `update_process_cache` needs this cycle's GPU pid set to decide which cached entries are GPU-attributed, so the process task is spawned after the GPU group joins rather than alongside it. Breaking that dependency would change collected values. Every other group is already in flight by then, so it only extends the critical path when GPU plus process collection outlasts all of them (it does not on the measured machine, and the code comment says so). The first-iteration path has no such dependency: it deliberately seeds the cache with an empty pid set, so all six groups start together there.

Blocking-pool sizing was checked: six concurrent tasks per cycle sits far below the default `max_blocking_threads`, and below the `max_blocking_threads(32)` used by the snapshot runtime.

## Measurements

Recorded on this machine (Apple M5 Max, 18 cores), release profile, via the `measure_collection_arms` harness added in `src/view/data_collection/local_collector/tests.rs`. The baseline was measured first, before any production code changed, by checking out `origin/main` into a separate worktree, appending the same harness to it, and building it with its own target directory. Two runs of each, alternating, on an otherwise idle machine. Wall clock and CPU are 100 back-to-back steady-state cycles divided by 100; CPU time is `getrusage(RUSAGE_SELF)` user + system, which counts every thread, so blocking-pool work is charged exactly like inline work.

| | baseline run 1 | baseline run 2 | after run 1 | after run 2 |
|---|---|---|---|---|
| wall clock per cycle | 20.118 ms | 21.717 ms | 14.883 ms | 14.783 ms |
| process CPU per cycle | 19.320 ms | 21.129 ms | 19.372 ms | 19.112 ms |

Wall clock drops roughly 28%. CPU time is flat within run-to-run noise and never above baseline, which is the "no regression" check: `spawn_blocking` relocated the cycles rather than adding any. There is no busy-wait and no duplicated work; the same reader calls run the same number of times.

Per-arm synchronous cost on this machine (mean of 12 samples, measured before the change): storage 14.371 ms, full process refresh 8.147 ms, GPU info 0.010 ms, CPU info 0.006 ms, memory 0.003 ms, chassis 0.002 ms, vGPU/MIG/GPU-processes below 0.001 ms. Serialized sum 22.541 ms. After the change the critical path is bounded by the storage pass, which matches the observed 14.8 ms.

**Limitations, stated honestly.** This is a single-machine, single-platform measurement. On Apple Silicon after PR #286 the GPU reader is effectively free (10 microseconds), so the win here comes from overlapping storage with process collection rather than from getting the GPU arm off the worker. On an NVIDIA host where `get_gpu_info` costs tens of milliseconds the GPU group would dominate and the ratio would differ; I have no NVIDIA hardware here to measure that, and the numbers above should not be read as a cross-platform claim. The `measure_collection_arms` harness is committed (ignored by default) precisely so the numbers can be reproduced on other hardware. The first-iteration path is not in the table because its one-shot nature makes a 100-cycle average meaningless; it is covered by a functional test instead.

## What changed

- `src/view/data_collection/local_collector.rs`: added `GpuCollection` plus `collect_from_gpu_readers` for the grouped GPU pass, added the `spawn_reader_pass` helper, rewrote `collect_parallel_first_iteration` to spawn six blocking groups instead of a nine-arm `tokio::join!`, and rewrote `collect_sequential` the same way.
- Renamed `collect_sequential` to `collect_steady_state`, since the old name no longer describes what it does.
- The first-iteration status updates now use `blocking_send` from inside the blocking closures, so each "collected" line still appears as its group finishes rather than all at once at the end.
- Hardened the startup status handler to bound the shifted index (`3 + index`) instead of the raw one, so an out-of-range write cannot panic that task.
- `src/view/data_collection/local_collector/tests.rs` (new): test module split out following the existing `#[path = ".../tests.rs"]` convention used by `cpu_linux`, `memory_linux`, `container_info` and others.

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo clippy --lib --tests -- -D warnings` (clean)
- [x] `cargo test --bin all-smi view::data_collection::local_collector::tests -- --test-threads=1` (3 passed, 1 ignored harness)
- [x] `parallel_collection_matches_serial_reference`: runs the new pipeline and a straight-line reference collection back to back and asserts the GPU uuid set, CPU/memory/vGPU/MIG/chassis row counts, and storage mount-point set match, plus that the process list stays sorted by CPU descending and truncated to `MAX_DISPLAY_PROCESSES`. Absolute metric values are sampled at different instants and cannot be compared for equality, so everything that is not time varying is asserted instead.
- [x] `repeated_collections_complete_without_deadlock`: runs `FULL_REFRESH_INTERVAL + 2` cycles under a timeout, covering both the selective and full process-refresh branches, to catch a lock or blocking-pool deadlock.
- [x] `first_iteration_collection_reports_startup_status`: drives the real first-iteration path through `initialize_readers` and asserts all five arms report completion, which is also what proves `blocking_send` inside `spawn_blocking` does not panic.

Closes #287

### DIFF
--- a/src/utils/system.rs
+++ b/src/utils/system.rs
@@ -40,7 +40,14 @@ pub fn with_global_system<F, R>(f: F) -> R
 where
     F: FnOnce(&mut System) -> R,
 {
-    let mut system = GLOBAL_SYSTEM.lock().unwrap();
+    // Recover from a poisoned lock instead of propagating it: `System` is a
+    // plain metrics snapshot, so continuing to use it after a panic mid-update
+    // elsewhere is safe, and it keeps one panicking caller (there are several,
+    // across GPU/process readers on multiple platforms) from permanently
+    // taking down every later call to this function.
+    let mut system = GLOBAL_SYSTEM
+        .lock()
+        .unwrap_or_else(std::sync::PoisonError::into_inner);
     f(&mut system)
 }
 
@@ -48,7 +55,9 @@ where
 /// This is the primary use case - collecting process information
 #[allow(dead_code)]
 pub fn refresh_global_processes() {
-    let mut system = GLOBAL_SYSTEM.lock().unwrap();
+    let mut system = GLOBAL_SYSTEM
+        .lock()
+        .unwrap_or_else(std::sync::PoisonError::into_inner);
     system.refresh_processes_specifics(
         ProcessesToUpdate::All,
         true,
--- a/src/view/data_collection/local_collector.rs
+++ b/src/view/data_collection/local_collector.rs
@@ -157,6 +157,21 @@ where
     tokio::task::spawn_blocking(move || f(&guard))
 }
 
+/// Await a reader-pass `JoinHandle` and log, rather than silently swallow, a
+/// panic in the underlying blocking task before falling back to that group's
+/// default (empty) result. Matches the `eprintln!("Warning: ...")` style
+/// already used in `initialize_readers` for other degraded-but-recoverable
+/// conditions.
+async fn join_or_log_default<R: Default>(handle: JoinHandle<R>, group_name: &str) -> R {
+    match handle.await {
+        Ok(value) => value,
+        Err(err) => {
+            eprintln!("Warning: {group_name} collection task failed: {err}");
+            R::default()
+        }
+    }
+}
+
 pub struct LocalCollector {
     gpu_readers: Arc<RwLock<Vec<Box<dyn GpuReader>>>>,
     cpu_readers: Arc<RwLock<Vec<Box<dyn CpuReader>>>>,
@@ -208,15 +223,16 @@ impl LocalCollector {
             return;
         }
 
-        // Add startup status with timeout
+        // Add startup status. Pushed unconditionally, like the CPU and memory
+        // lines below: `collect_parallel_first_iteration` counts on exactly
+        // three preamble lines being present (the `3 + index` offset at the
+        // status handler), so silently skipping this push on lock contention
+        // would shift every later status update to the wrong line.
         {
-            let state_result = timeout(Duration::from_secs(2), app_state.lock()).await;
-
-            if let Ok(mut state) = state_result {
-                state
-                    .startup_status_lines
-                    .push("✓ Initializing GPU readers...".to_string());
-            }
+            let mut state = app_state.lock().await;
+            state
+                .startup_status_lines
+                .push("✓ Initializing GPU readers...".to_string());
         }
 
         let gpu_readers = get_gpu_readers();
@@ -421,7 +437,14 @@ impl LocalCollector {
                 // pid set is deliberately empty here: `merge_gpu_processes`
                 // below applies the GPU attribution for the first cycle.
                 let gpu_pids: HashSet<u32> = HashSet::new();
-                let mut cache = process_cache.write().unwrap();
+                // A panic elsewhere while this lock (or `GLOBAL_SYSTEM`, see
+                // `with_global_system`) was held would otherwise poison it and
+                // make every later cycle panic here too. The cached data is a
+                // plain metric map that is safe to keep using after a panic
+                // mid-update, so recover instead of propagating the poison.
+                let mut cache = process_cache
+                    .write()
+                    .unwrap_or_else(std::sync::PoisonError::into_inner);
                 update_process_cache(system, &gpu_pids, &mut cache)
             });
             let _ =
@@ -430,8 +453,8 @@ impl LocalCollector {
         });
 
         // A panicking reader degrades that group to empty rather than taking the
-        // whole collection loop down, matching how the process arm already
-        // handled a failed `spawn_blocking`.
+        // whole collection loop down. The panic is still logged, so a run of
+        // empty data is diagnosable instead of silent.
         let GpuCollection {
             gpu_info: all_gpu_info,
             gpu_processes,
@@ -440,12 +463,12 @@ impl LocalCollector {
             gpu_pids: _,
             vgpu_info: all_vgpu_info,
             mig_info: all_mig_info,
-        } = gpu_task.await.unwrap_or_default();
-        let all_cpu_info = cpu_task.await.unwrap_or_default();
-        let all_memory_info = memory_task.await.unwrap_or_default();
-        let all_chassis_info = chassis_task.await.unwrap_or_default();
-        let all_storage_info = storage_task.await.unwrap_or_default();
-        let all_processes = processes_task.await.unwrap_or_default();
+        } = join_or_log_default(gpu_task, "GPU").await;
+        let all_cpu_info = join_or_log_default(cpu_task, "CPU").await;
+        let all_memory_info = join_or_log_default(memory_task, "memory").await;
+        let all_chassis_info = join_or_log_default(chassis_task, "chassis").await;
+        let all_storage_info = join_or_log_default(storage_task, "storage").await;
+        let all_processes = join_or_log_default(processes_task, "process").await;
 
         // Close the channel and wait for status handler to finish
         drop(status_tx);
@@ -562,7 +585,7 @@ impl LocalCollector {
             gpu_pids,
             vgpu_info: all_vgpu_info,
             mig_info: all_mig_info,
-        } = gpu_task.await.unwrap_or_default();
+        } = join_or_log_default(gpu_task, "GPU").await;
 
         let process_cache = Arc::clone(&self.process_cache);
         let processes_task = tokio::task::spawn_blocking(move || {
@@ -594,16 +617,23 @@ impl LocalCollector {
                 // OPTIMIZATION: Use process cache to reduce memory allocation overhead
                 // Instead of creating new ProcessInfo objects every cycle, we update
                 // existing cached objects and only allocate for new processes.
-                let mut cache = process_cache.write().unwrap();
+                // A panic elsewhere while this lock (or `GLOBAL_SYSTEM`, see
+                // `with_global_system`) was held would otherwise poison it and
+                // make every later cycle panic here too. The cached data is a
+                // plain metric map that is safe to keep using after a panic
+                // mid-update, so recover instead of propagating the poison.
+                let mut cache = process_cache
+                    .write()
+                    .unwrap_or_else(std::sync::PoisonError::into_inner);
                 update_process_cache(system, &gpu_pids, &mut cache)
             })
         });
 
-        let all_cpu_info = cpu_task.await.unwrap_or_default();
-        let all_memory_info = memory_task.await.unwrap_or_default();
-        let all_chassis_info = chassis_task.await.unwrap_or_default();
-        let all_storage_info = storage_task.await.unwrap_or_default();
-        let all_processes = processes_task.await.unwrap_or_default();
+        let all_cpu_info = join_or_log_default(cpu_task, "CPU").await;
+        let all_memory_info = join_or_log_default(memory_task, "memory").await;
+        let all_chassis_info = join_or_log_default(chassis_task, "chassis").await;
+        let all_storage_info = join_or_log_default(storage_task, "storage").await;
+        let all_processes = join_or_log_default(processes_task, "process").await;
 
         let mut all_processes = merge_gpu_processes(all_processes, gpu_processes);
 
--- a/src/view/data_collection/local_collector.rs
+++ b/src/view/data_collection/local_collector.rs
@@ -19,6 +19,7 @@ use std::sync::atomic::{AtomicU32, Ordering};
 use std::time::Duration;
 use sysinfo::Disks;
 use tokio::sync::{Mutex, RwLock};
+use tokio::task::JoinHandle;
 use tokio::time::timeout;
 
 /// Type alias for the process cache using std::sync::RwLock for synchronous access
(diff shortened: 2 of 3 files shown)


inureyes:

### Intent
> Move every synchronous reader pass in the local collection pipeline off the async task so the collection cycle actually overlaps instead of serializing inside `tokio::join!` (issue #287).

### Verified
- **Completeness.** All nine former `join!` arms are accounted for in both paths. GPU device info, GPU processes, vGPU and MIG are one group (`collect_from_gpu_readers`) because they share reader sampling state; CPU, memory, chassis, storage and full-process are five more. `collect_parallel_first_iteration` spawns all six up front; `collect_steady_state` spawns five up front and the process group after the GPU group joins. Nothing synchronous from a reader is left on the async worker in either path.
- **Behavior preservation.** Intra-GPU call order (`get_gpu_info` -> `get_gpu_processes` -> `get_vgpu_info` -> `get_mig_info`) matches the pre-change sequence in both paths. Sort, `MAX_DISPLAY_PROCESSES` truncation, `tracked_pids` update, `refresh_cycle` handling and `inject_gpu_power` are byte-for-byte equivalent. The first-iteration path still seeds `update_process_cache` with an empty pid set (`gpu_pids: _` plus comment), matching the old `let (gpu_processes, _gpu_pids) = ...`. Steady state still feeds this cycle's real pid set, and the spawn-after-join ordering is what preserves that.
- **Locking.** `spawn_reader_pass` takes `Arc::read_owned()` in async context and moves the `OwnedRwLockReadGuard` into the closure. No `blocking_read` / `blocking_lock` / `block_on` anywhere in `src/view/`. No guard is held across an `.await` on either side. The only writers of the four reader locks in production code are the four `timeout(..., self.*.write())` blocks inside `initialize_readers`, which runs once under the `initialized` mutex before any collection, so long-lived read guards cannot starve a writer. Lock order is consistent (`GLOBAL_SYSTEM` then `process_cache`; no reader touches `process_cache`), so no cycle exists.
- **No dead code.** `collect_sequential` and the nine-arm `join!` are gone, not duplicated. `collect_steady_state` has two production call sites (`DataCollectionStrategy::collect`, `collect_with_app_state`), both updated; `LocalCollector` is only referenced from `src/view/data_collector.rs`, so the rename has no other callers.
- **Status-handler hardening is a genuine fix.** The old guard was `if index < len { lines[3 + index] = msg }`, which panics whenever fewer than three preamble lines were pushed. That is reachable: `initialize_readers` skips its first push if the 2s `app_state` lock timeout fires. Bounding `3 + index` closes it, and it only narrows which writes happen, so no caller changes behavior.
- **Conventions.** Test module split follows the existing `#[cfg(test)] #[path = ".../tests.rs"]` pattern (`cpu_linux`, `memory_linux`, `intel_gpu_linux`). Existing shared helpers (`merge_gpu_processes`, `update_process_cache`, `with_global_system`, `filter_docker_aware_disks`, `inject_gpu_power`) are reused, nothing reimplemented.
- **Pool sizing.** The view runtime (`src/main.rs:171`) sets `worker_threads(4)` and leaves `max_blocking_threads` at the default, so six concurrent blocking tasks are well within budget. The snapshot runtime's `max_blocking_threads(32)` is not on this path (`LocalCollector` is unused there).
- **Tests.** The three new tests are real, not smoke tests: `parallel_collection_matches_serial_reference` runs the actual parallel pipeline against an in-test straight-line reference and compares everything non-time-varying; `repeated_collections_complete_without_deadlock` covers both the selective and full refresh branches under a timeout; `first_iteration_collection_reports_startup_status` drives the real first-iteration path and is what proves `blocking_send` inside `spawn_blocking` does not panic (without it, such a panic would be swallowed by `unwrap_or_default`). Re-ran locally: `cargo fmt --check` clean, `cargo check --lib --tests` clean, `cargo clippy --lib --tests -- -D warnings` clean, and the three tests pass both with `--test-threads=1` and with default parallelism (CI runs bare `cargo test`).

### Findings (none blocking, no auto-fix applied)
- **MEDIUM: a panicking reader is now silent.** All six joins use `.unwrap_or_default()`. Previously only the process arm did that; a panic in the GPU, CPU, memory, chassis or storage arm propagated out of the collection. Now it degrades to empty with no log. For GPU info this is the worst case: `update_state` merges by uuid, so an empty `gpu_info` leaves the last-known values frozen on screen with nothing printed. It is also self-perpetuating, since a panic inside `with_global_system` poisons `GLOBAL_SYSTEM` and every later `.lock().unwrap()` panics too. Suggest logging the `JoinError` before defaulting, matching the `eprintln!("Warning: ...")` style already used in `initialize_readers`.
- **LOW: first-iteration ordering against the global `System` is no longer deterministic.** The GPU group and the full-process group now start together, and both can reach `with_global_system`. In the collector's call path only Jetson's `get_gpu_processes` name-matching fallback and Apple Silicon's `get_used_memory` do so, and both only call `refresh_memory()` or read `system.processes()`, so there is no data race (std `Mutex`) and no deadlock. The observable effect is confined to the first cycle on Jetson, where the fallback may now see a freshly refreshed process table instead of an empty one. Steady state is unaffected because the process task is spawned after the GPU group joins. Worth a sentence in the PR body, since acceptance criterion 5 asks for value-for-value equivalence.
- **LOW: the post-join merge and sort still run on the async worker.** `merge_gpu_processes` plus the full-list sort (before truncation to 500) stays on the worker in both paths. Pre-existing, not a regression, but it is the remaining synchronous chunk if "nothing synchronous on the async task" is read strictly.
- **LOW (informational): hung-reader surface moved, not removed.** `src/main.rs` already notes that `spawn_blocking` cannot be cancelled, so a hung hardware reader parks a blocking-pool thread for the life of the process. Five more reader passes now sit on that pool. Net effect is arguably better (a hung reader no longer occupies one of the four workers), and the collection hangs either way, so no change is proposed.

### Verification
- [x] Tests pass

inureyes:

Scope: changed files (`src/view/data_collection/local_collector.rs`, `src/view/data_collection/local_collector/tests.rs`) plus the reader and runtime code they now reach from the blocking pool.

**Result: no CRITICAL and no HIGH findings. No auto-fix commits were made.**

### Verified clean (the stated focus areas)

- **No lock-order inversion.** `GLOBAL_SYSTEM` then `process_cache` is the only nesting in the file, it exists at exactly one site, and it is always taken in that order. Nothing else in the tree takes `process_cache`.
- **No reentrancy on `GLOBAL_SYSTEM`.** Nothing reachable from `collect_from_gpu_readers` calls `with_global_system` from inside another `with_global_system`, so the non-reentrant `std::sync::Mutex` cannot self-deadlock.
- **Steady-state ordering is preserved.** `gpu_task` is joined before `processes_task` is spawned, so this cycle's `gpu_pids` and the GPU readers' `refresh_memory` calls stay ordered ahead of the process refresh exactly as before the change. No torn or mixed-cycle aggregation: `merge_gpu_processes` and `inject_gpu_power` both consume same-cycle inputs.
- **First-iteration `GLOBAL_SYSTEM` concurrency is narrower than it looks.** AMD's (`amd.rs:601`) and Intel's (`enrichment.rs:62`) `with_global_system` calls both live in `get_process_info`, which the collector never calls. Only Jetson `get_gpu_processes` and Apple Silicon `get_used_memory` touch the global `System` from the GPU group, and both only `refresh_memory` or read `processes()`. That is why previously-reported finding (b) correctly stays LOW.
- **macOS triple `collect_once()` is safe.** `gpu_task`, `cpu_task` and `chassis_task` now call the process-global `NativeMetricsManager::collect_once` concurrently on three blocking threads. That method already has double-checked locking plus a 500 ms cache window (`macos_native/manager.rs:104,315`) written specifically for "the several readers that run per collection cycle", so the first caller samples IOReport and the other two hit the cache. No delta splitting, no IOReport handle race. The one-shot first-session branch, which does a blocking `sample_interval_ms` window, is now on the blocking pool instead of an async worker, which is an improvement.
- **No writer starvation.** The only writers of the four reader locks are `initialize_readers` (lines 249 to 286) and the test helper. There is no hotplug or refresh path, so the long-lived owned read guards cannot block anyone. The doc comment on `spawn_reader_pass` is accurate.
- **No task leak, no unbounded growth, no busy-wait.** `run_local_mode` awaits each cycle to completion, is spawned fire-and-forget with no `select!` and no `timeout` (`view/runner.rs:94`, `view/data_collector.rs:144`), so a cycle is never cancelled mid-flight. At most six blocking tasks are in flight and all are joined in the same cycle. A hung reader stalls the loop rather than accumulating tasks, so the blocking pool cannot be exhausted by repeated cycles. Reader call counts are identical to pre-change.
- **`blocking_send` inside `spawn_blocking` is legal** (blocking-pool threads run inside a blocking region, so `block_on` does not panic) and the new first-iteration test covers it.
- **No security surface change.** No new input parsing, no subprocess, no privilege use, no serialization boundary. The single new `unsafe` block (`libc::getrusage`, `tests.rs:52`) is test-only, carries a `SAFETY:` comment, and is correct: `rusage` is POD so `mem::zeroed` is sound, and `tv_usec` maxes at 999999 so the nanos argument to `Duration::new` cannot overflow.

### MEDIUM

**M1. Poison cascade turns one panic into permanently empty data.** `processes_task` takes `GLOBAL_SYSTEM` via `with_global_system` (`utils/system.rs:43`, `.lock().unwrap()`) and then `process_cache.write().unwrap()` inside it (`local_collector.rs:424,597`). A panic in `update_process_cache` poisons both. From that point every later cycle fails: `processes_task` panics and `.unwrap_or_default()` yields an empty process list, and on Apple Silicon `gpu_task` panics too, because `get_gpu_info` reaches `get_used_memory` which reaches the poisoned `GLOBAL_SYSTEM`, so `.unwrap_or_default()` yields empty GPU info forever with no log line. This extends the already-reported finding (a) rather than replacing it: the new part is that the failure is permanent and crosses task boundaries. Cheapest mitigations are recovering from poison (`.unwrap_or_else(PoisonError::into_inner)`) and logging `JoinError` once instead of discarding it.

**M2. New tests drive real hardware and shared global state, but CI runs them in parallel.** The test plan documents `--test-threads=1`; `.github/workflows/ci.yml:46` runs bare `cargo test --verbose`, so the three new tests run concurrently with each other and with the rest of the binary's tests. All three mutate the shared `GLOBAL_SYSTEM`, and `first_iteration_collection_reports_startup_status` additionally calls `initialize_readers` and builds a second full set of platform readers while the other two are collecting. On Apple Silicon that constructs a second handle onto the process-global native metrics manager whose `Drop` calls `shutdown()`. Recommend serializing them behind a shared static `tokio::sync::Mutex` or `serial_test`.

**M3. The hardcoded `+3` status offset can still mis-target, just not panic.** `initialize_readers` pushes its first preamble line under `timeout(Duration::from_secs(2), app_state.lock())` and silently skips it on timeout, while the other two push unconditionally. If that first push is skipped, every completion marker lands one slot early. The bound added in this PR (`local_collector.rs:331`) correctly prevents the out-of-range panic but not the wrong-slot write. Capturing `state.startup_status_lines.len()` when the five placeholders are pushed and sending absolute slot indices would remove the coupling entirely.

### LOW

**L1. Status channel capacity is an undocumented invariant.** Capacity 10 with exactly five single-message senders means `blocking_send` can never block today. If a group is ever added, or a group sends more than one message, `blocking_send` would park a blocking-pool thread until the receiver drains, and the receiver needs `app_state`. Worth one line of comment next to `mpsc::channel::<(usize, String)>(10)` stating that capacity must stay at or above the number of sends.

**L2. Previously reported (c) confirmed.** `merge_gpu_processes`, the sort and the truncation still run on the async worker. Pre-existing and small relative to the arms that were moved.

### Checks run on this branch

- `cargo fmt --check` clean
- `cargo clippy --lib --tests -- -D warnings` clean
- `cargo test --bin all-smi view::data_collection::local_collector::tests` 3 passed, 1 ignored

Issue #307, opened 2026-08-05
feat: Intel Mac GPU monitoring via IOAccelerator: utilization and VRAM for iGPU and AMD dGPU


inureyes:
## Background

Deferred follow-up to #306 (Intel Mac base support). That issue delivers CPU, SMC temperature, and chassis power support for Intel Macs and explicitly accepts an empty GPU list. This issue adds actual GPU monitoring (utilization, VRAM, temperature) for Intel integrated GPUs and AMD discrete GPUs via the IOAccelerator registry, the sudo-free path used by common macOS monitoring tools. To be scheduled when demand materializes; Apple has announced macOS 26 (Tahoe) as the last Intel-supported release, so the useful lifetime is bounded.

Depends on #306, in particular the SMC sp78 discovery fix and the Intel-specific reader wiring.

## Current State

- The only macOS GpuReader is Apple-Silicon-only and is not even instantiated on Intel (src/device/reader_factory.rs:117-125). Its metrics all come from IOReport/SMC via the native manager (src/device/readers/apple_silicon_native.rs:139-256); no Metal API is actually used ("Metal" is only a version label, :200-209).
- The GPU frequency table lookup matches `AppleARMIODevice` (src/device/macos_native/ioreport.rs:206) and the core-count fallback parses `AGXAccelerator` (apple_silicon_native.rs:373-391); both are absent on Intel Macs, which expose `IntelAccelerator` / `AMDRadeonX*` classes instead.

## Scope

- [ ] New IntelMacGpuReader enumerating IOAccelerator registry entries (IntelAccelerator, AMDRadeonX*) and reading PerformanceStatistics: "Device Utilization %", VRAM used/total, "In use system memory".
- [ ] GPU temperature via SMC TG0P/TG0D (keys already listed at src/device/macos_native/smc.rs:501; requires the sp78 discovery fix from #306).
- [ ] Handle multi-GPU topologies: iGPU + dGPU automatic switching (MacBook Pro), multiple dGPUs (Mac Pro), eGPU hot plug and unplug.
[...]

Issue #308, opened 2026-08-05
ci: publish the Intel Mac (x86_64) artifact to the Homebrew formula with on_arm/on_intel split


inureyes:
## Background

Follow-up to #306, which adds an `x86_64-apple-darwin` entry to the release matrix and ships an `all-smi-macos-x86_64.zip` release asset. Once that asset exists, the Homebrew formula should serve Intel Mac users the correct binary instead of only the Apple Silicon one. This issue depends on #306 shipping the `all-smi-macos-x86_64.zip` asset; it cannot be completed before then.

## Current State

Evidence from .github/workflows/update_homebrew_formula.yml:

- The workflow downloads only `all-smi-macos-aarch64.zip` (:91) and rewrites the formula with it (:161).
- The `set_artifact` helper (:130-158) asserts exactly ONE matching `url` stanza in the formula and fails otherwise, so it cannot express two macOS artifacts today.
- The formula therefore has a single macOS block with no `on_arm`/`on_intel` split.

## Scope

- [ ] Restructure Formula/all-smi.rb (in the tap repo) to use `on_macos do on_arm ... on_intel ... end` blocks with per-arch url + sha256.
- [x] Extend update_homebrew_formula.yml to download both macOS assets, compute both checksums, and update both stanzas; rework `set_artifact` (or replace it) to address stanzas by arch instead of asserting uniqueness.
- [x] Handle the transition: releases published before #306 lack the x86_64 asset. The workflow should fail clearly (or skip the intel stanza with a warning) when the asset is missing, rather than corrupting the formula.
- [ ] Verify end to end on the next release: `brew install lablup/tap/all-smi` (adjust to the actual tap name found in the workflow) resolves the correct artifact on both an Apple Silicon Mac and an Intel Mac.

## Decision: per-arch artifacts vs universal binary

Per-arch artifacts are preferred over a universal (lipo) binary:

- The release matrix already produces per-arch builds on different runners, so no extra combining step or artifact hand-off is needed.
- Downloads stay smaller for every user (a universal binary roughly doubles the download size).
- Notarization stays per-artifact and unchanged; a lipo step would require re-signing and re-notarizing the combined binary in an additional job.

lipo (a single universal binary served from one stanza) is the rejected alternative for the reasons above.

## Acceptance Criteria

- [ ] After a release containing both macOS assets, the formula update workflow runs green.
- [ ] The formula contains both `on_arm` and `on_intel` stanzas with correct sha256 values.
[...]

PR #312, opened 2026-08-05
feat: add Intel Mac support for CPU, SMC, and chassis metrics (#306)

### LINKED ISSUE #306: feat: Intel Mac (x86_64-apple-darwin) support: CPU/SMC metrics, sudo-free chassis power, and self-hosted release packaging
## Background

all-smi supports Apple Silicon Macs natively (IOReport/SMC, no sudo), but Intel Macs (x86_64-apple-darwin) are only half-supported today. The codebase compiles cleanly for x86_64-apple-darwin: there are no `target_arch` gates anywhere in src/, all macOS code is gated on `target_os = "macos"` only, and the doctor module already reports the `x86_64-apple-darwin` triple (src/doctor/checks/platform.rs:259-261). The gap is runtime behavior and release packaging.

Motivation: Intel Macs remain useful as remote-viewer clients and are increasingly repurposed as VM host machines, so first-class monitoring support has practical value. Apple has announced macOS 26 (Tahoe) as the last Intel-supported release, so the scope of this work is deliberately bounded.

## Current State

On a physical Intel Mac today:

- The GPU reader list is empty: src/device/reader_factory.rs:117-125 only pushes a macOS reader when `is_apple_silicon()`.
- The chassis reader always returns None: src/device/readers/chassis/mod.rs:38-50 unconditionally picks AppleSiliconNativeChassisReader on macOS, and its `get_chassis_info` (src/device/readers/chassis/apple_silicon_native.rs:50-53) requires the native metrics manager that is never initialized on Intel.
- The CPU path is functional but thin: slow `system_profiler` text parsing, no per-core utilization, no temperature, hardcoded frequency constants.

### A. Runtime correctness fixes (valuable independently of Intel support)

- [x] Remove the hard panic in arch detection: src/device/platform_detection.rs:210-211 calls `execute_command_default("uname", &["-m"]).expect(...)`. Make it degrade gracefully by treating failure as non-Apple-Silicon. Note there is a second, non-panicking copy of the same probe at src/device/cpu_macos.rs:79-85; consider unifying the two.
- [x] Fix the init-gate inconsistency: all three binary entry points gate `initialize_native_metrics_manager` on `is_apple_silicon()` (src/main.rs:294-301, 331-341, 621-629), but the library path src/client.rs:207-219 initializes it unconditionally on macOS. On Intel this fails inside `IOReport::new` because the "Energy Model" channel group does not exist (src/device/macos_native/ioreport.rs:578-580, propagated via src/device/macos_native/manager.rs:110-127). Gate the library path identically.
- [x] Fix the chassis factory: src/device/readers/chassis/mod.rs:38-50 must select an Intel Mac chassis reader (new, see group D) when not on Apple Silicon, instead of unconditionally constructing the Apple Silicon one.

### B. Intel CPU path hardening (src/device/cpu_macos.rs, Intel branch at :333-402)

- [x] Replace the `system_profiler SPHardwareDataType` text parsing (:350-355, parser at :746-796, cached at :341/:793 because it costs ~500 ms) with sysctl queries: `machdep.cpu.brand_string`, `hw.physicalcpu`, `hw.logicalcpu`, `hw.cpufrequency` (present on Intel). This also removes the hardcoded hyperthreading assumption `total_threads = cores * 2` (:774), which is wrong on non-HT models.
- [x] Fill per-core utilization: the Intel branch currently returns `per_core_utilization: Vec::new()` (:399), which leaves the Activity panel with zero core bars (src/ui/activity_panel.rs:88-94 falls to the Individual strategy with no data). Reuse the existing sysinfo-based `get_per_core_utilization_no_refresh` (:945-996); `CoreType::Standard` already exists for Intel cores (:984-986).
- [x] Wire CPU temperature: `get_cpu_temperature()` returns None unconditionally on Intel (:908-912) even though the SMC layer already lists the classic Intel keys TC0P/TC0D (src/device/macos_native/smc.rs:458-460). Read them via a lightweight SMC connection (see group C). The CPU renderer already shows the temperature row whenever the value is Some (src/ui/renderers/cpu_renderer.rs:383-386), and the Intel display branches (single frequency, plain core count, "L3 Cache:" label) already exist (cpu_renderer.rs:322-350, 419-429).
- [x] Replace the hardcoded base/max frequency constants for Intel (:888-906, currently 2400/3000) with `hw.cpufrequency` / `hw.cpufrequency_max` sysctls where available.

### C. SMC layer fixes (src/device/macos_native/smc.rs)

- [x] Extend key discovery to Intel sensor types: discovery currently filters to type `flt ` only (:313 inside :286-337), which matches Apple Silicon sensors but misses Intel Macs, whose temperature keys are `sp78` fixed-point. The decoder already supports sp78 (:369-448); accept sp78 in discovery so TC0P/TC0D/TG0P/TG0D are found dynamically.
- [x] Surface fan speeds and system power: fan speeds (`FNum`, `F0Ac`..., collected at :540-557) and total system power (`PSTR`, :535-537) are already read but marked dead code and discarded by the chassis reader (src/device/readers/chassis/apple_silicon_native.rs:114 hardcodes empty fans). Expose them for the Intel chassis reader; Intel Macs are precisely the machines with real fans.
- [x] Make the SMC + NSProcessInfo thermal path usable without IOReport: NativeMetricsManager hard-fails when `IOReport::new` fails (manager.rs:110-127), so no partial mode exists. Preferred approach: do not refactor the manager; instead create Intel-specific readers that use the SMC and thermal (src/device/macos_native/thermal.rs, fully arch-agnostic) modules directly. This keeps the Apple Silicon path untouched.

### D. Sudo-free approximate chassis power (new IntelMacChassisReader)

- [x] Total power: SMC `PSTR` (system total power) as the primary source. Document that this is an approximation and model-dependent.
- [x] CPU package power: best effort from SMC power keys where present (the PC0C/PCPC family varies by model); omit cleanly when absent.
- [x] Fan RPMs from group C; thermal pressure from NSProcessInfo (works on Intel, macOS 10.10.3+).
- [x] Export through the existing chassis metric family (src/api/metrics/chassis.rs): `all_smi_chassis_power_watts`, `all_smi_chassis_fan_speed_rpm`, `all_smi_chassis_thermal_pressure_info`. Adjust help strings that currently hardcode Apple Silicon wording: "(CPU+GPU+ANE)" at chassis.rs:145-155 and "Thermal pressure level (Apple Silicon)" at chassis.rs:167-180.
- [x] UI: verify the local view renders chassis data for Intel. The UI branches key off data strings like `architecture == "Apple Silicon"` (e.g. src/ui/dashboard.rs:54-58), so Intel data flows down the generic path; the ANE row is correctly hidden via src/ui/gpu_sparkline_panel.rs:618-626.

### E. Release packaging on the self-hosted Intel runner

- [x] Add an `x86_64-apple-darwin` entry to the build matrix in .github/workflows/release.yml (currently the only macOS entry at :48-55 is aarch64-apple-darwin on macos-14). The new entry runs on the org's self-hosted Intel Mac: runner label `self-hosted-macos-15-x64`, runner group `macOS x64`. Example runs-on syntax:
  ```yaml
  runs-on:
    group: macOS x64
    labels: self-hosted-macos-15-x64
  ```
  Matrix fields consistent with the existing entry: artifact_name `all-smi`, asset_name `all-smi-macos-x86_64`, archive_ext `.zip`, protoc_platform `osx-x86_64`.
- [x] Verify codesign and notarization on the new entry: those steps are gated on `runner.os == 'macOS'` (release.yml:184-197 import-codesign-certs + codesign, :204-280 notarytool), so they apply automatically. Confirm keychain handling works on a self-hosted (persistent) machine: import into a temporary keychain and clean it up after the job so certificates do not accumulate, and confirm the base64-encoded packaging secrets decode correctly there.
- [x] The `targets` workflow_dispatch input accepts only the families windows|linux|macos|all (release.yml:56-80); `macos` now means both arches. Optionally extend granularity, or document the behavior.
- [x] Security: the repo is public. Keep the self-hosted runner restricted to the release workflow (tag push and workflow_dispatch by maintainers). Do NOT attach it to PR-triggered CI, and keep the runner group scoped so fork PRs can never schedule jobs on it. A hosted-runner cross-compile from macos-14 (arm64) with `rustup target add x86_64-apple-darwin` is a documented fallback if the self-hosted machine is offline.
- [x] Docs: update the platform support table (src/lib.rs:72 currently lists only Apple Silicon for macOS) and the README.

## Out of Scope

- GPU utilization/VRAM monitoring for Intel iGPU / AMD dGPU via IOAccelerator (follow-up: #307). In this issue an empty GPU list on Intel Macs is acceptable.
- Homebrew formula publication of the x86_64 artifact (follow-up: #308).
- powermetrics-based power data (requires sudo, conflicts with the project's no-sudo principle).

## Acceptance Criteria

- [ ] `cargo build --release` for x86_64-apple-darwin succeeds, and `all-smi view` / `all-smi api` run on a physical Intel Mac without panics.
- [ ] On an Intel Mac, view mode shows: a CPU block with the real model name, correct core/thread counts (no hyperthreading assumption), per-core utilization bars in the Activity panel, and CPU temperature from SMC; a chassis block with total power (PSTR), fan RPMs, and thermal pressure.
- [ ] `all-smi api` exports the chassis metric family and CPU metrics for Intel Macs; no Apple-Silicon-only metric names are emitted for Intel hosts.
- [ ] The release workflow builds, signs, notarizes, and uploads `all-smi-macos-x86_64.zip` using the self-hosted runner.
- [x] The library path (src/client.rs) no longer attempts IOReport init on Intel; no error spam in verbose mode.
- [ ] Integration is complete end to end: the new readers are wired into the reader factory, the data flows through the actual working logic into both the TUI and the API, and the behavior is verified on a physical Intel Mac. Module-only delivery does not satisfy this issue.

## Implementation Notes

Implemented in PR for this issue. Verification limits, recorded honestly:

- No physical Intel Mac was available, so every acceptance criterion phrased as "on a physical Intel Mac" is left unchecked. The Intel build is verified by `cargo check --target x86_64-apple-darwin --lib --bins`, and the Apple Silicon path is regression-tested on real hardware.
- The self-hosted runner path in the release workflow cannot be exercised outside a real release run, so that criterion is also left unchecked.

Correction to a premise in group C: accepting `sp78` in SMC key discovery is not sufficient on its own. Discovery also filtered key names to `Tp*`/`Te*`/`Tg*`, and the Intel keys are `TC*`/`TG*`, so TC0P/TC0D/TG0P/TG0D would still never have been discovered. The name filter was extended alongside the type filter, with each family pinned to its own data type so the two cannot cross.

## References

- src/device/platform_detection.rs:210-211 (panicking arch probe)
- src/device/reader_factory.rs:117-125 (GPU reader gate)
- src/device/readers/chassis/mod.rs:38-50 (chassis factory)
- src/device/readers/chassis/apple_silicon_native.rs:50-53, :114 (manager requirement, hardcoded empty fans)
- src/device/cpu_macos.rs:333-402, :746-796, :888-912, :945-996 (Intel CPU branch)
- src/device/macos_native/smc.rs:286-337, :369-448, :458-460, :535-557 (SMC discovery, decoders, Intel keys, fans and PSTR)
- src/device/macos_native/manager.rs:110-127 and src/device/macos_native/ioreport.rs:578-580 (IOReport hard dependency)
- src/api/metrics/chassis.rs:145-180 (help strings)
- .github/workflows/release.yml:48-80, :184-280 (build matrix, signing and notarization)

inureyes:
## Summary

Intel Macs already compiled for `x86_64-apple-darwin`, but at runtime they got an empty GPU list, a chassis reader that always returned `None`, and a thin CPU path built on slow `system_profiler` text parsing with hardcoded frequency constants, no per-core data, and no temperature. This wires up everything that does not need Apple Silicon hardware (CPU, SMC, chassis) and adds the `x86_64-apple-darwin` release target on the self-hosted Intel builder.

The Apple Silicon path is deliberately untouched in structure: `NativeMetricsManager` still hard-fails without IOReport, and the new Intel readers talk to the SMC and NSProcessInfo modules directly rather than partially initializing it.

### A. Runtime correctness

- `src/device/platform_detection.rs`: `detect_apple_silicon` no longer `.expect()`s on the `uname` probe; a failed probe now degrades to "not Apple Silicon", which every caller already handles. It also recognizes Rosetta 2 via `sysctl.proc_translated`, so an `x86_64` build running on Apple Silicon reports the hardware it is actually on rather than claiming to be an Intel Mac. That probe is compiled out on `aarch64` builds, which can only run natively, so the Apple Silicon path still costs exactly one subprocess. Adds `is_intel_mac()`.
- `src/device/cpu_macos.rs`: the second private copy of the `uname -m` probe now delegates to the shared cached detector, so there is one answer per process.
- `src/client.rs`: the library path gated `initialize_native_metrics_manager` on `is_apple_silicon()`, matching what the three binary entry points in `main.rs` already did. On Intel this init always failed inside `IOReport::new` because the "Energy Model" channel group does not exist there.
- `src/device/readers/chassis/mod.rs`: the factory routes to the new Intel reader instead of unconditionally constructing `AppleSiliconNativeChassisReader`.
- `src/device/reader_factory.rs`: comment recording that the empty Intel GPU list is intentional and tracked in #307.

### B. Intel CPU path

- New `src/device/cpu_macos_intel.rs` owns Intel hardware discovery. sysctl is the primary source (`machdep.cpu.brand_string`, `hw.packages`, `hw.physicalcpu`, `hw.logicalcpu`, `hw.cpufrequency`, `hw.cpufrequency_max`, `hw.l3cachesize`), read in a single batched call.
- `system_profiler SPHardwareDataType` is kept, not deleted, as a gap filler for the L3 cache size and the marketing processor name, which sysctl does not supply on every model. It only runs when sysctl actually left a gap.
- `hw.cpufrequency` is treated as optional because it was removed on later Intel Macs. The fallback chain is `hw.cpufrequency`, then the nominal clock parsed out of the brand string, then `system_profiler`'s "Processor Speed". Max frequency comes from `hw.cpufrequency_max` and falls back to the base clock.
- The `total_threads = cores * 2` hyperthreading assumption is gone. Threads come from `hw.logicalcpu`; if that is unreadable the fallback is the kernel's logical CPU count from sysinfo, and if that also fails it is one thread per physical core. It never doubles.
- Per-core utilization is populated via the existing `get_per_core_utilization_no_refresh` with `CoreType::Standard`, so the Activity panel renders core bars instead of falling to the Individual strategy with no data.
- CPU temperature and package power now come from the SMC.

### C. SMC layer

- Key discovery filtered to type `flt ` **and** to names `Tp*`/`Te*`/`Tg*`. See "Findings" below: extending the type filter alone would not have been enough. Both filters were extended, as two disjoint families each pinned to its own data type.
- Fan speeds (`FNum` / `F<i>Ac`, plus `F<i>Mx` for the rated maximum) and total system power (`PSTR`) were already collected and then discarded. They are now exposed through `get_fan_readings()` and a validated `get_system_power()`, along with a new best-effort `get_cpu_package_power()` trying `PCPT`, `PCPC`, `PC0C`.
- All three reads are plausibility-bounded so a missing or differently-typed key cannot surface as a metric.
- New `SmcConnection` type: a lazily opened, reusable connection that remembers an unreachable SMC instead of retrying an IOKit handshake on every poll. Both the CPU reader and the chassis reader hold one. `SMCMetrics::collect()` still reconnects per call, which is fine because the native metrics manager caches its results.
- The `smc` module and `get_thermal_state` are now public so the Intel readers can use them without going through the manager.

### D. Chassis (new `IntelMacChassisReader`)

- Total power from SMC `PSTR`, documented in the payload as an approximation via a `power_source` detail rather than left to look metered. `powermetrics` would be metered but requires sudo, which this project does not ask for on macOS.
- CPU package power where the model exposes it, omitted cleanly when absent.
- Fan RPMs with their rated maximums, and thermal pressure from NSProcessInfo. Thermal pressure needs no SMC connection, so a chassis block is always produced even if the SMC is unreachable.
- Everything flows through the existing chassis metric family, so `all_smi_chassis_power_watts`, `all_smi_chassis_fan_speed_rpm`, `all_smi_chassis_thermal_pressure_info`, and `all_smi_chassis_cpu_power_watts` are exported with no exporter changes. Help strings no longer hardcode "(CPU+GPU+ANE)" and "(Apple Silicon)".

### E. Release packaging

- `x86_64-apple-darwin` matrix entry on the self-hosted Intel builder, with `artifact_name` `all-smi`, `asset_name` `all-smi-macos-x86_64`, `archive_ext` `.zip`, `protoc_platform` `osx-x86_64`. Its `runs-on` is carried as a `{group, labels}` object in the matrix JSON while every other entry stays a label string; `runs-on` is evaluated per matrix job instance, so the two shapes do not interact.
- Signing and notarization are gated on `runner.os == 'macOS'` and apply automatically, but the self-hosted runner is persistent, so the Developer ID key now imports into a keychain named after the run and an `always()` step deletes it. Without that, the private key would survive the job on disk and every run would append another certificate to the same fixed-name keychain. Deleting a keychain also drops it from every search list, so no separate `security list-keychains` reset is needed, and the step tolerates the keychain not existing when import never ran.
- The Actions cargo cache is skipped for this target as it is for the Windows self-hosted box: the machine keeps its workspace and `~/.cargo` between jobs, so the cache would only mirror state it already has.
- Fork-PR safety: the workflow triggers only on `release: published` and `workflow_dispatch`, both of which need write access. `ci.yml` is the PR-triggered workflow and runs on `ubuntu-latest` only. This is now documented in a comment at the trigger block so nobody adds a PR trigger later.
- The `targets` input description records that `macos` now covers both arches.
- Docs: platform support table in `src/lib.rs` split into Apple Silicon and Intel rows, plus a new "macOS (Intel)" section and a supported-hardware bullet in the README.

### Out of the issue's checklist, but directly adjacent

The `apple.smc` doctor check reported a euid test and advised "re-run with sudo". Nothing on this path needs sudo, and the advice was actively misleading for Intel Macs whose temperature, fans, and chassis power all depend on that connection. It now probes the SMC and reports what it found.

## Findings that contradict the issue's premises

**Group C's "accept sp78 in discovery so TC0P/TC0D/TG0P/TG0D are found" is not sufficient on its own.** Discovery filtered on two axes, not one: data type `flt `, *and* key name prefix `Tp*`/`Te*`/`Tg*`. The Intel keys are `TC*` and `TG*`, so accepting `sp78` while leaving the name filter alone would still have discovered nothing. Both filters were extended together.

**On the regression risk that flagged this as the highest-risk change:** the two families cannot collide, because the Apple Silicon names use a lowercase second character (`p`/`e`/`g`) and the Intel names use uppercase (`C`/`G`), and each family is now pinned to its own data type in the classifier (`flt ` accepts only lowercase, `sp78` only uppercase). Beyond that, discovery is only reached as a fallback when the static key list produced no readings, and the static CPU list already contains `TC0P`/`TC0D` and the GPU list `TG0P`/`TG0D`. So on an Apple Silicon machine the extension cannot introduce a sensor family that the static path was not already probing, and where it could fire at all it would be on a machine currently reporting no temperature. The existing `10.0..=120.0` range filter remains as the last line of defense. A unit test asserts the families do not cross data types.

**Fan speeds on Apple Silicon are deliberately left unexposed.** `SMCMetrics` collects them there too and `AppleSiliconNativeChassisReader` still hardcodes an empty fan list. Surfacing them would start emitting `all_smi_chassis_fan_speed_rpm` for every existing Apple Silicon host, which is an output change for current users that this issue does not ask for. It is a clean follow-up if wanted.

**`all_smi_cpu_frequency_mhz` and the P/E cluster frequency metrics report 0 on this M1 Ultra.** That is pre-existing and unrelated: the released v0.25.0 Homebrew binary was run side by side and reports the same zeros. Noting it so it is not mistaken for a regression from this PR.

## Test plan

- [x] `cargo check --target x86_64-apple-darwin --lib --bins` passes. This is the primary Intel-side verification.
- [x] `cargo clippy --lib --tests -- -D warnings` and `cargo clippy --bins -- -D warnings` clean.
- [x] `cargo fmt` clean.
- [x] New unit tests pass for the parts testable without Intel hardware: SMC key type and name classification including the cross-family negative cases, sp78 fixed-point decoding, sysctl key/value parsing, brand-string frequency extraction, Hz to MHz rounding, the system_profiler fallback leaving `logical_cores` unset, sysctl-wins-over-fallback merging, the no-hyperthreading and non-zero-socket defaults, arch detection being total and stable, and chassis reader selection following the architecture.
- [x] Apple Silicon regression, `api` mode: built and ran on this M1 Ultra, then diffed `/metrics`. GPU utilization/temperature/power, CPU utilization/temperature/power, and the full chassis family (`power_watts`, `thermal_pressure_info`, `cpu_power_watts`, `gpu_power_watts`, `ane_power_watts`) are all still present with `platform="Apple Silicon"`. The Intel reader is correctly not selected.
- [x] Apple Silicon regression, `local` TUI mode: ran in a pty for 8 seconds. CPU model, P-CPU and E-CPU cluster bars, the ANE row, thermal pressure "Nominal", and the GPU Metrics panel all render. No panics.
- [x] `all-smi doctor` reports `PASS apple.smc  SMC reachable (no sudo required)`.
- [x] The release workflow YAML parses, and the matrix JSON round-trips through the same `jq` filter the setup job uses, producing both macOS entries for `targets: macos`.

## Not verified

- **No physical Intel Mac was available.** Nothing here has been run on Intel hardware. The issue's acceptance criteria phrased as "on a physical Intel Mac" are left unchecked on the issue, and the runtime SMC key availability (`PSTR`, the `PCPT`/`PCPC`/`PC0C` package power family, `F<i>Mx`) is best-effort by construction and model-dependent by nature.
- **The self-hosted Intel runner path has not executed.** Whether the base64-encoded packaging secrets decode there, and whether the temporary keychain import and cleanup behave on that specific machine, can only be confirmed by a real release run or a maintainer `workflow_dispatch`. The notarization step's base64 handling and its openssl sanity check are architecture-independent and unchanged.
- `cargo build --release` for `x86_64-apple-darwin` was not run to completion; `cargo check` for that target was used instead to keep the build bounded.

Closes #306

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -146,10 +146,16 @@ jobs:
           echo "RUSTUP_HOME=C:\.rustup" >> $env:GITHUB_ENV
 
       # 2) Cache Cargo build artifacts
-      # Skipped on the self-hosted runners. The Windows box has a persistent
-      # CARGO_HOME (configured above), and the Intel Mac keeps its workspace and
-      # ~/.cargo between jobs, so on both the Actions cache would only mirror
-      # state the machine already has while costing a full upload every run.
+      # Skipped on the self-hosted runners, for different reasons on each.
+      # The Windows box has a persistent CARGO_HOME and RUSTUP_HOME (configured
+      # above), so the Actions cache would only mirror state already on disk.
+      # On the Intel Mac only ~/.cargo survives between jobs: actions/checkout
+      # defaults to clean: true, whose `git clean -ffdx` removes the ignored
+      # target/ directory, so each release compiles the dependency tree cold.
+      # That is accepted here because releases are infrequent and a release
+      # target/ is large enough that caching it would crowd the repository's
+      # 10 GB cache budget shared with the other five targets. Revisit with
+      # real build times if the Intel job becomes the critical path.
       - name: Cache cargo
         if: matrix.target != 'x86_64-pc-windows-msvc' && matrix.target != 'x86_64-apple-darwin'
         uses: actions/cache@v5
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -197,13 +197,35 @@ jobs:
         run: cargo build --release --target ${{ matrix.target }} --locked
 
       # 8) macOS code signing
-      # The Developer ID key goes into a keychain named after this run, and the
-      # cleanup step at the end of the job deletes it unconditionally. That
+      # The Developer ID key goes into a keychain named after this run. That
       # matters on the self-hosted Intel Mac, which is persistent: with the
-      # action's default fixed keychain name the signing key would survive the
-      # job on disk and every run would append another certificate to the same
-      # keychain. GitHub-hosted macOS runners are ephemeral and unaffected, but
-      # they take the same path so there is only one behavior to reason about.
+      # action's default fixed name (`signing_temp`) two runs would collide and
+      # every run would append another certificate to the same keychain.
+      #
+      # Cleanup is the import action's own post step, which runs
+      # `security delete-keychain` at the end of the job on success, failure,
+      # and cancellation. Do NOT add a step that deletes this keychain: the
+      # post step would then fail with "The specified keychain could not be
+      # found" and fail the whole job after the artifacts were uploaded, which
+      # also skips the notify-teams and promote-release jobs.
+      - name: Remove stale signing keychains
+        if: runner.os == 'macOS'
+        run: |
+          # Safety net for the persistent runner, reachable only when an
+          # earlier run was hard-killed before its post step could delete its
+          # keychain. Scoped to this workflow's own name prefix and to files
+          # older than a day, so it can never remove a keychain this workflow
+          # did not create nor one belonging to a run still in flight.
+          STALE=$(find "$HOME/Library/Keychains" -maxdepth 1 -name 'all-smi-signing-*' -mtime +0 2>/dev/null || true)
+          if [ -n "$STALE" ]; then
+            echo "$STALE" | while IFS= read -r kc; do
+              echo "Removing stale signing keychain: $kc"
+              security delete-keychain "$kc" || true
+            done
+          else
+            echo "No stale signing keychains"
+          fi
+
       - name: Import Developer ID certificate
         if: runner.os == 'macOS'
         uses: apple-actions/import-codesign-certs@v6
@@ -381,27 +403,6 @@ jobs:
           files: |
             ${{ matrix.asset_name }}${{ matrix.archive_ext }}
             ${{ matrix.asset_name }}${{ matrix.archive_ext }}.sha256
-
-      # 12) Delete the run-scoped signing keychain.
-      # always() so a failed build, a cancelled run, or a notarization error
-      # still removes the Developer ID private key from the machine. Deleting a
-      # keychain also drops it from every search list, so no separate
-      # `security list-keychains` reset is needed. Failures are tolerated: the
-      # keychain does not exist when certificate import never ran.
-      - name: Remove signing keychain
-        if: ${{ always() && runner.os == 'macOS' }}
-        run: |
-          # The action materializes `<name>` as `<name>.keychain-db`; try both
-          # spellings so a change in that convention cannot silently skip cleanup.
-          KEYCHAIN="all-smi-signing-${{ github.run_id }}-${{ github.run_attempt }}"
-          if security delete-keychain "${KEYCHAIN}.keychain-db" 2>/dev/null \
-            || security delete-keychain "$KEYCHAIN" 2>/dev/null; then
-            echo "Removed signing keychain $KEYCHAIN"
-          else
-            echo "No signing keychain to remove"
-          fi
-          echo "Remaining user keychains:"
-          security list-keychains -d user
 
   # ============================================================================
   # Microsoft Teams release notification (Power Automate Workflows webhook)
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,5 +1,13 @@
 name: Release
 
+# SECURITY: this repository is public and this workflow schedules jobs on
+# self-hosted runners (the Windows signing box and the Intel Mac builder).
+# Both triggers below are maintainer-only: `release: published` requires push
+# access to publish a release, and `workflow_dispatch` requires write access.
+# Neither can be fired by a pull request, so a fork PR can never run code on a
+# self-hosted machine. Do NOT add `pull_request`, `pull_request_target`, or
+# `issue_comment` triggers here, and keep self-hosted runners out of ci.yml
+# (which is PR-triggered and runs on GitHub-hosted runners only).
 on:
   release:
     types: [published]
@@ -15,7 +23,7 @@ on:
         description: 'Tag to build from AND upload artifacts to (e.g. v1.2.3). Empty = build the dispatched ref.'
         required: false
       targets:
-        description: 'Platforms to build (comma-separated: windows, linux, macos, or all). Empty = all.'
+        description: 'Platform families to build (comma-separated: windows, linux, macos, or all). Empty = all. "macos" covers both aarch64 and x86_64.'
         required: false
         default: ''
         type: string
@@ -45,12 +53,20 @@ jobs:
           # Full build matrix; each entry is tagged with os_family for filtering.
           # JSON uses only double quotes, so a single-quoted bash literal is safe
           # and avoids heredoc indentation pitfalls inside this YAML block.
+          #
+          # `os` becomes the job's runs-on value. It is a label string for every
+          # GitHub-hosted runner and for the label-targeted Windows box, and a
+          # {group, labels} object for the self-hosted Intel Mac, which has to be
+          # selected out of a specific runner group. runs-on is evaluated per
+          # matrix job instance, so the two shapes coexist without affecting each
+          # other.
           ALL='[
             {"os_family":"linux",   "target":"x86_64-unknown-linux-gnu",    "os":"ubuntu-22.04",             "artifact_name":"all-smi",     "asset_name":"all-smi-linux-x86_64",       "archive_ext":".tar.gz", "protoc_platform":"linux-x86_64"},
             {"os_family":"linux",   "target":"x86_64-unknown-linux-musl",   "os":"ubuntu-latest",           "artifact_name":"all-smi",     "asset_name":"all-smi-linux-x86_64-musl",  "archive_ext":".tar.gz", "protoc_platform":"linux-x86_64"},
             {"os_family":"linux",   "target":"aarch64-unknown-linux-gnu",   "os":"ubuntu-22.04-arm",        "artifact_name":"all-smi",     "asset_name":"all-smi-linux-aarch64",      "archive_ext":".tar.gz", "protoc_platform":"linux-aarch_64"},
             {"os_family":"linux",   "target":"aarch64-unknown-linux-musl",  "os":"ubuntu-24.04-arm",        "artifact_name":"all-smi",     "asset_name":"all-smi-linux-aarch64-musl", "archive_ext":".tar.gz", "protoc_platform":"linux-aarch_64"},
             {"os_family":"macos",   "target":"aarch64-apple-darwin",        "os":"macos-14",                "artifact_name":"all-smi",     "asset_name":"all-smi-macos-aarch64",      "archive_ext":".zip",    "protoc_platform":"osx-aarch_64"},
+            {"os_family":"macos",   "target":"x86_64-apple-darwin",         "os":{"group":"macOS x64","labels":"self-hosted-macos-15-x64"}, "artifact_name":"all-smi", "asset_name":"all-smi-macos-x86_64", "archive_ext":".zip", "protoc_platform":"osx-x86_64"},
             {"os_family":"windows", "target":"x86_64-pc-windows-msvc",      "os":"windows-on-macmini02-x64","artifact_name":"all-smi.exe", "asset_name":"all-smi-windows-x86_64",     "archive_ext":".zip",    "protoc_platform":""}
           ]'
 
@@ -130,11 +146,12 @@ jobs:
           echo "RUSTUP_HOME=C:\.rustup" >> $env:GITHUB_ENV
 
       # 2) Cache Cargo build artifacts
-      # Skipped on the Windows self-hosted runner: its persistent CARGO_HOME
-      # (configured above) already caches the registry across runs, so the
-      # GitHub Actions cache would only mirror an empty ~/.cargo location.
+      # Skipped on the self-hosted runners. The Windows box has a persistent
+      # CARGO_HOME (configured above), and the Intel Mac keeps its workspace and
+      # ~/.cargo between jobs, so on both the Actions cache would only mirror
+      # state the machine already has while costing a full upload every run.
       - name: Cache cargo
-        if: matrix.target != 'x86_64-pc-windows-msvc'
+        if: matrix.target != 'x86_64-pc-windows-msvc' && matrix.target != 'x86_64-apple-darwin'
         uses: actions/cache@v5
         with:
           path: |
@@ -180,12 +197,20 @@ jobs:
         run: cargo build --release --target ${{ matrix.target }} --locked
 
       # 8) macOS code signing
+      # The Developer ID key goes into a keychain named after this run, and the
+      # cleanup step at the end of the job deletes it unconditionally. That
+      # matters on the self-hosted Intel Mac, which is persistent: with the
+      # action's default fixed keychain name the signing key would survive the
+      # job on disk and every run would append another certificate to the same
+      # keychain. GitHub-hosted macOS runners are ephemeral and unaffected, but
+      # they take the same path so there is only one behavior to reason about.
       - name: Import Developer ID certificate
         if: runner.os == 'macOS'
         uses: apple-actions/import-codesign-certs@v6
         with:
           p12-file-base64: ${{ secrets.DEV_ID_CERT_P12 }}
           p12-password: ${{ secrets.DEV_ID_CERT_PASSWORD }}
+          keychain: all-smi-signing-${{ github.run_id }}-${{ github.run_attempt }}
 
       - name: Code sign macOS binary
         if: runner.os == 'macOS'
@@ -356,6 +381,27 @@ jobs:
           files: |
             ${{ matrix.asset_name }}${{ matrix.archive_ext }}
             ${{ matrix.asset_name }}${{ matrix.archive_ext }}.sha256
+
+      # 12) Delete the run-scoped signing keychain.
+      # always() so a failed build, a cancelled run, or a notarization error
+      # still removes the Developer ID private key from the machine. Deleting a
+      # keychain also drops it from every search list, so no separate
+      # `security list-keychains` reset is needed. Failures are tolerated: the
+      # keychain does not exist when certificate import never ran.
+      - name: Remove signing keychain
+        if: ${{ always() && runner.os == 'macOS' }}
+        run: |
+          # The action materializes `<name>` as `<name>.keychain-db`; try both
+          # spellings so a change in that convention cannot silently skip cleanup.
+          KEYCHAIN="all-smi-signing-${{ github.run_id }}-${{ github.run_attempt }}"
+          if security delete-keychain "${KEYCHAIN}.keychain-db" 2>/dev/null \
+            || security delete-keychain "$KEYCHAIN" 2>/dev/null; then
+            echo "Removed signing keychain $KEYCHAIN"
+          else
+            echo "No signing keychain to remove"
+          fi
+          echo "Remaining user keychains:"
+          security list-keychains -d user
 
   # ============================================================================
   # Microsoft Teams release notification (Power Automate Workflows webhook)
--- a/README.md
+++ b/README.md
@@ -303,6 +303,16 @@ Older releases introduced environment variables with different naming. All alias
   - Provides actual temperature readings from SMC sensors
   - Run with: `all-smi local`
 
+### macOS (Intel)
+- **No sudo required:** Uses the Apple SMC and NSProcessInfo directly
+  - CPU model, socket/core/thread counts, and clocks come from `sysctl`
+  - Per-core utilization bars, plus CPU temperature from the SMC `TC0P`/`TC0D` sensors
+  - Chassis block reports fan RPMs, thermal pressure, and approximate total system power (SMC `PSTR`)
+  - GPU monitoring is not available: the integrated Intel and discrete AMD GPUs in Intel Macs are not read, so the GPU list stays empty
+  - Total power is the SMC's own estimate rather than a metered value, and its accuracy varies by model. `powermetrics` would be metered but needs sudo, which all-smi does not ask for on macOS
+  - Download `all-smi-macos-x86_64.zip` from the releases page. It is signed and notarized like the Apple Silicon archive
+  - Run with: `all-smi local`
+
 ### Linux with AMD GPUs
 - **Sudo Access Required:** AMD GPU monitoring requires `sudo` to access `/dev/dri` devices
 - **ROCm Installation:** AMD GPU support requires ROCm drivers and libraries
@@ -580,6 +590,7 @@ Stable check IDs (greppable across versions):
   - Actual CPU/GPU temperature readings from SMC sensors
   - Thermal pressure monitoring
   - P/E core architecture support
+  - Intel Macs (x86_64): CPU, memory, and chassis monitoring via SMC and sysctl, including per-core bars, CPU temperature, fan RPMs, and approximate system power. No GPU monitoring
 - **NVIDIA Jetson:** 
   - Special support for Tegra-based systems
   - DLA (Deep Learning Accelerator) monitoring
--- a/src/api/metrics/chassis.rs
+++ b/src/api/metrics/chassis.rs
@@ -15,9 +15,10 @@
 //! Chassis metrics exporter for Prometheus
 //!
 //! Exports node-level metrics including:
-//! - Total power consumption (CPU+GPU+ANE)
-//! - Thermal pressure (Apple Silicon)
-//! - Individual power components (CPU, GPU, ANE)
+//! - Total power consumption (CPU+GPU+ANE on Apple Silicon, SMC PSTR on Intel Macs)
+//! - Thermal pressure (macOS)
+//! - Individual power components (CPU, GPU, ANE), whichever the platform reports
+//! - Fan speeds, where the platform reports them
 
 use super::{MetricBuilder, MetricExporter};
 use crate::device::ChassisInfo;
@@ -146,7 +147,7 @@ impl<'a> MetricExporter for ChassisMetricExporter<'a> {
             builder
                 .help(
                     "all_smi_chassis_power_watts",
-                    "Total chassis power consumption in watts (CPU+GPU+ANE)",
+                    "Total chassis power consumption in watts",
                 )
                 .type_("all_smi_chassis_power_watts", "gauge");
 
@@ -164,12 +165,12 @@ impl<'a> MetricExporter for ChassisMetricExporter<'a> {
             }
         }
 
-        // Export thermal pressure metric (Apple Silicon)
+        // Export thermal pressure metric (macOS)
         if flags.has_thermal_pressure {
             builder
                 .help(
                     "all_smi_chassis_thermal_pressure_info",
-                    "Thermal pressure level (Apple Silicon)",
+                    "Thermal pressure level (macOS)",
                 )
                 .type_("all_smi_chassis_thermal_pressure_info", "gauge");
 
--- a/src/client.rs
+++ b/src/client.rs
@@ -204,17 +204,25 @@ impl AllSmi {
         let _ = &config;
 
         // Initialize platform-specific managers
+        // The native metrics manager is built on IOReport's "Energy Model"
+        // channel group, which only exists on Apple Silicon. Initializing it on
+        // an Intel Mac always fails inside `IOReport::new`, so gate it the same
+        // way the binary entry points do instead of failing on every startup.
         #[cfg(target_os = "macos")]
         let macos_initialized = {
-            match initialize_native_metrics_manager(config.sample_interval_ms) {
-                Ok(()) => true,
-                Err(e) => {
-                    // Log but don't fail - some metrics may still work
-                    if config.verbose {
-                        eprintln!("Warning: macOS native metrics init failed: {e}");
+            if crate::device::is_apple_silicon() {
+                match initialize_native_metrics_manager(config.sample_interval_ms) {
+                    Ok(()) => true,
+                    Err(e) => {
+                        // Log but don't fail - some metrics may still work
+                        if config.verbose {
+                            eprintln!("Warning: macOS native metrics init failed: {e}");
+                        }
+                        false
                     }
-                    false
                 }
+            } else {
+                false
             }
         };
 
--- a/src/device/cpu_macos.rs
+++ b/src/device/cpu_macos.rs
@@ -423,7 +423,13 @@ impl MacOsCpuReader {
         }
 
         // Last-resort thread count, used only when `hw.logicalcpu` is missing.
-        let logical_cpu_fallback = self.system.read().unwrap().cpus().len() as u32;
+        // Deliberately not `self.system.cpus().len()`: this runs before the
+        // first `ensure_cpu_refreshed`, and a freshly constructed `System` has
+        // an empty CPU list, so that source reports 0 and the result would be
+        // cached for the process lifetime.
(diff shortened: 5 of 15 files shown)


inureyes:

### Intent
> Make Intel Macs first-class at runtime (sysctl-based CPU discovery, SMC temperature/fans/power, an SMC-backed chassis reader) and add `x86_64-apple-darwin` to the release matrix on the self-hosted Intel builder, without disturbing the Apple Silicon path.

### Findings Addressed
- [x] **Release job deleted its own signing keychain, which would have failed every macOS build** (CRITICAL). `apple-actions/import-codesign-certs@v6` registers a `post:` step that runs `security delete-keychain "<keychain>.keychain"` whenever `create-keychain` is true (the default). Step 12 deleted the same run-scoped keychain first, so the post step would hit "The specified keychain could not be found" (exit 50) and, because it calls `@actions/exec` without `ignoreReturnCode`, would `core.setFailed` and fail the job after the artifacts were already uploaded. `notify-teams` and `promote-release` both declare `needs: build`, so both would have been skipped. This regressed the already-working aarch64 macOS release too. Verified locally that `security delete-keychain "<name>.keychain-db"` succeeds and that a missing keychain exits 50. Fixed in 1ff97fd: cleanup is now the action's own post step (which already runs on success, failure, and cancellation), the run-scoped keychain name is kept, and a pre-import sweep removes leftover `all-smi-signing-*` keychains older than a day as a safety net for hard-killed runs. The prefix and age bound mean it can never remove a keychain this workflow did not create, nor one belonging to a run still in flight.
- [x] **`detect_apple_silicon` could hand Apple Silicon hardware the Intel readers** (HIGH). Removing the old `.expect()` on the `uname -m` probe was right, but it degraded to `false`. On an `aarch64-apple-darwin` build whose PATH cannot resolve `uname`, `is_intel_mac()` becomes true, `create_chassis_reader()` picks `IntelMacChassisReader`, `MacOsCpuReader` takes the Intel CPU path, and IOReport is never initialized. An `aarch64-apple-darwin` binary can only run on Apple Silicon, so detection now answers from the compile-time target first, mirroring the Rosetta short-circuit already present for x86_64. Fixed in 1ff97fd.

### Remaining Items
- [ ] **`runs-on` receiving a `{group, labels}` object cannot be exercised from here** (MEDIUM). The Intel entry carries an object while every other matrix entry carries a label string. This is the documented shape and expressions in `runs-on` are supported, but nothing in this review proves the parser accepts it in this position. Worth a maintainer `workflow_dispatch` with `targets: macos` before the next release, which would also exercise the keychain path on the real machine.
- [ ] **`PSTR` and the `PCPT`/`PCPC`/`PC0C` family are `sp78` on many Intel Macs, saturating at 127.99 W** (MEDIUM). A high-draw machine would wrap negative, and the new `0.0..=2000.0` plausibility filter then drops the reading. That is the safe failure direction (no metric rather than a wrong one), noted so a missing `all_smi_chassis_power_watts` on a large Intel Mac is not mistaken for a bug. Not verifiable without hardware.
- [ ] **The Intel branches of `get_cpu_base_frequency` / `get_cpu_max_frequency` are unreachable** (LOW). Both are only called from `get_apple_silicon_cpu_info` (`cpu_macos.rs:194-195`), inside the `native_data == None` fallback, which is only reached when `is_apple_silicon` is true. Replacing the hardcoded 2400/3000 there is correct but has no runtime effect; the real Intel frequency comes from `get_intel_mac_cpu_info`, which does satisfy the issue's group B intent.
- [ ] **`collection_always_produces_safe_values` spawns a real `system_profiler`** (LOW). On any macOS host, including Apple Silicon CI, `IntelCpuHardware::collect(4)` finds gaps (no `hw.cpufrequency`, no `hw.l3cachesize`) and shells out, adding roughly half a second to the suite. In production this is a one-time cost because `get_intel_hardware_info` caches, so only the test is affected.
- [ ] **`SmcConnection` memoizes `Unavailable` for the process lifetime with no log line** (LOW). Intentional and documented, but a transient handshake failure at startup silently removes CPU temperature and chassis power for the whole run with nothing to diagnose from.

### Notes
- `brew install all-smi` still cannot deliver the Intel build: `update_homebrew_formula.yml` fetches only `all-smi-macos-aarch64.zip` and asserts exactly three sha256 stanzas. Issue #306 puts this in Out of Scope and tracks it as #308, so it is not a defect in this PR, only a reminder that #308 must land before the Intel artifact is reachable through the advertised installer.
- Help-string change: `all_smi_chassis_power_watts` lost the `(CPU+GPU+ANE)` suffix, which brings the real exporter in line with what the mock server has always emitted (`src/mock/templates/common.rs:196`). Metric names, types, and labels are unchanged, so Apple Silicon scrapers see no difference beyond HELP text.

### Verification
- [x] All stated requirements implemented, except the acceptance criteria that require physical Intel hardware, which are correctly left unchecked on #306
- [x] Integrated into project code flow: `create_chassis_reader()` routes to `IntelMacChassisReader`, and it is constructed once per process at `client.rs:250`, `api/collection_loop.rs:63`, `view/data_collection/local_collector.rs:261`, and `snapshot/collector.rs:71`, so the cached SMC connection is opened once rather than per poll. Data reaches the TUI (`ui/renderers/chassis_renderer.rs:195` renders fans) and the Prometheus exporter (`api/metrics/chassis.rs`) through the existing chassis families with no exporter changes. The Intel CPU branch populates per-core utilization, temperature, and package power. No orphaned modules.
- [x] Project conventions followed: `new_command`, `parse_colon_value!`, `get_hostname`, `get_per_core_utilization_no_refresh` with `CoreType::Standard`, `FanInfo`, and the `platform` / `api` / `cpu_power_watts` detail-key convention taken from the Apple Silicon reader
- [x] Existing modules reused where applicable; no reimplementation found, and the duplicate `uname -m` probe in `cpu_macos.rs` was removed in favor of the shared cached detector
- [x] No unintended structural changes: the Apple Silicon path keeps its shape, `NativeMetricsManager` still hard-fails without IOReport, and the SMC discovery extension is reachable only as a fallback the Apple Silicon static key list never needs
- [x] Tests pass: `cargo fmt --check` clean, `cargo clippy --lib --tests -- -D warnings` clean on the host and on `x86_64-apple-darwin`, and the `platform_detection`, `macos_native::smc`, `cpu_macos_intel`, and `readers::chassis` unit tests all green after the fixes

No Intel hardware exists in this environment, so nothing here claims runtime verification on an Intel Mac.

inureyes:

Focused on self-hosted runner exposure, secret handling on a persistent machine, subprocess safety, hot-path cost, and FFI resource safety. Two findings fixed in 8416c4e; the rest are reported below without code changes.

### Self-hosted runner exposure: no exposure found

The threat model holds. Enumerated every trigger and every job's runner:

| Workflow | Triggers | Runners |
|---|---|---|
| `ci.yml` | `push`, **`pull_request`** | `ubuntu-latest` only (3 jobs) |
| `release.yml` | `release: published`, `workflow_dispatch` | self-hosted Intel Mac + Windows |
| `debian_build.yml` | `workflow_run` <- "Release", `workflow_dispatch` | GitHub-hosted Ubuntu only |
| `update_homebrew_formula.yml` | `workflow_run` <- "Release", `workflow_dispatch` | `macos-latest` |
| `launchpad_ppa.yml` | `workflow_dispatch` | `ubuntu-latest` |

- No `pull_request_target` or `issue_comment` trigger exists anywhere in the repo.
- The only PR-triggered workflow is `ci.yml`, and all three of its jobs are `ubuntu-latest`. A fork PR cannot reach a self-hosted machine.
- Both `workflow_run` chains hang off **"Release"**, not CI. Since Release is only reachable via `release: published` or `workflow_dispatch` (both requiring write access), the chain is not reachable from a fork PR.
- No attacker-controlled `github.event.*` value is interpolated into any `run:` block. The `notify-teams` job correctly passes every release field through `env:` and into `jq --arg` rather than interpolating into the shell. The remaining `github.event.inputs.*` interpolations are `workflow_dispatch` inputs, which require write access.
- Repo default workflow permissions are `read`, and `can_approve_pull_request_reviews` is false. No `dependabot.yml` or other automation entry point.

The `SECURITY:` header comment added to `release.yml` accurately states the invariant and tells future editors what not to add. Worth keeping.

### Secret handling on the persistent runner: sound

No secret is written anywhere that outlives the job. The Developer ID p12 goes into a keychain named per `run_id`/`run_attempt` (correct for a persistent box, avoids the cross-run collision the fixed `signing_temp` default would cause), cleaned by the import action's own post step with the stale sweep as a backstop. The notarization `.p8` is written to `RUNNER_TEMP` with `chmod 600` and a `trap cleanup EXIT`. Nothing lands in the workspace.

### Fixed in 8416c4e

**HIGH: `SMC::new` leaked an IOKit service handle on every call** (`src/device/macos_native/smc.rs:217`)

`IOServiceGetMatchingService` returns a +1-retained object the caller owns, and `IOServiceOpen` creates an independent user client rather than adopting that reference. The handle was released on neither the success nor the failure path, and `IOObjectRelease` was not even declared in the file's extern block, unlike `macos_native/ioreport.rs`, which handles this correctly. `SMCMetrics::collect` opens a fresh connection once per collection cycle, so a long-running `view` or `api` process leaked a mach port right steadily for its whole lifetime. Pre-existing on `main`, but this PR builds directly on it and adds two more `SMC::new` call sites.

**HIGH: both new subprocess calls were unbounded** (`src/device/cpu_macos_intel.rs:81`, `:130`)

`Command::output()` waits for the child forever and reads stdout to EOF with no cap, bypassing the device layer's own `execute_command_default` (timeout, kill-child-on-timeout, 16 MiB cap, argument validation) that `platform_detection.rs` in this same PR already uses. `system_profiler` queries IOKit and can block on unresponsive hardware, and both callers run on a `spawn_blocking` worker Tokio cannot cancel, so a wedged child would strand that worker permanently. `system_profiler` got a 5s ceiling since it legitimately takes ~500ms and does not fit the shared fast-fail budget.

Also in that commit: both binaries now invoked by absolute path (macOS `sudo` does not set `secure_path`, so a writable directory ahead of `/usr/sbin` could substitute either), a null-port guard on `Drop for SMC`, and a corrected `unsafe impl Send` justification (the old one claimed IOKit thread safety; the real reason is that `conn` is a mach port name with no thread affinity, and `Sync` is withheld on purpose).

### Reported, not fixed

**MEDIUM: `logical_cpu_fallback` is always 0 at the point it is read** (`src/device/cpu_macos.rs:426`). `self.system` is `System::new()`, which loads nothing until a refresh, and `get_intel_hardware_info()` runs before `ensure_cpu_refreshed()`. So `cpus().len()` is 0 and the fallback never fires. On a machine where `hw.logicalcpu` is unreadable, the result is cached permanently, in the worst case as `1 core / 1 thread / Unknown Intel CPU`. Suggest `std::thread::available_parallelism()`, and consider not caching an all-defaults record.

**MEDIUM: per-poll SMC key re-probing with no negative cache** (`intel_mac.rs:129-141`, `smc.rs:373-393`). `read_value` always does `read_key_info` then `read`, two IOKit round trips per key, and absent keys are re-probed forever. On a model lacking all of `PCPT`/`PCPC`/`PC0C` that is 6 wasted round trips per poll, plus 12 more from `get_cpu_temperature` probing Apple Silicon `Tp*` keys that cannot exist on Intel. Caching `KeyInfo` per key would halve every read and give the negative cache for free.

**MEDIUM: `socket_count` is unclamped and sizes a per-poll allocation** (`cpu_macos.rs:360-372`). Sourced from `hw.packages` with no upper bound and rebuilt every cycle. Suggest clamping in `apply_defaults`.

**MEDIUM: `MAX_FANS` and the key formatter can drift** (`smc.rs:42`, `:616`). `format!("F{index}Ac")` yields 5 chars for `index >= 10`, at which point `str_to_fourcc` returns 0 and an SMC request is issued with key code 0 instead of being skipped. Safe only because `MAX_FANS == 8`, with the two 570 lines apart. A `const _: () = assert!(MAX_FANS <= 10);` would pin it.

**MEDIUM: third-party actions are not SHA-pinned on a signing runner.** `dtolnay/rust-toolchain@stable` (a mutable branch ref), `apple-actions/import-codesign-certs@v6`, `softprops/action-gh-release@v2`, `actions/checkout@v6`, `actions/cache@v5` all execute on the self-hosted machines in a job that has the Developer ID cert, its password, and the notarization key in scope. `sha_pinning_required` is false on the repo. Pre-existing for the Windows box, but this PR extends the same exposure to the macOS signing identity.

**LOW: the cargo-cache skip rationale is half wrong.** The comment says the Intel Mac "keeps its workspace and `~/.cargo` between jobs". `~/.cargo` does persist, but `actions/checkout` runs with default `clean: true`, which does `git clean -ffdx`, and `-x` removes ignored files including `/target/`. So `target/` is wiped every run and, with the Actions cache disabled for this target, there is no restore path: every Intel Mac release build is a full cold compile. Registry downloads are avoided, compilation is not.

**LOW, also reported:** `hz_to_mhz` can overflow in debug builds and truncates silently on `as u32` (`cpu_macos_intel.rs:213`); Rosetta detection still fails open if `sysctl` is unresolvable on `PATH` and would hand Apple Silicon the Intel readers, which `libc::sysctlbyname` would close (`platform_detection.rs:243`); `system_profiler` label matching is locale-sensitive and assumes MB for `L3 Cache`, both fixable with `-json`; `read()` ignores the returned `output_size` so a short response is indistinguishable from zeros; `data_size` is not validated against decoder width (no OOB risk, the buffer is a fixed `[u8; 32]`, but a mistyped key yields a bogus value); `DISCOVERED_KEYS` init can issue 1,200-3,200 IOKit round trips in one poll while holding the reader mutex; `SmcConnection::Unavailable` latches permanently and silently; `.unwrap()` on the mutex at `cpu_macos.rs:423` disagrees with the graceful `let Ok(..) else` used ten lines above; `objc_msgSend` is declared C-variadic and is now reached on x86_64 for the first time (works on both ABIs, but Apple documents casting to the exact prototype).

### Checked and clean

No command injection: every argument in the new code is a `&'static str` literal, passed via `args()`, no `sh -c`, no `format!`-built commands, no env/hostname/config/CLI value reaching any argument. No `unwrap()`/`expect()` outside `#[cfg(test)]` in the new files. No unbounded caches. `KeyData` is `#[repr(C)]` at 80 bytes matching Apple's `SMCParamStruct`, size arguments cannot disagree with the Rust type, no `mem::zeroed`/`MaybeUninit`, no OOB indexing, no double-close, no use-after-close. `Send`/`Sync` is sound: `SMC: Send` only, wrapped in `Mutex` at both storage sites, which satisfies `ChassisReader: Send + Sync` with no further `unsafe`.

**No Apple Silicon regression.** Steady-state polls still spawn zero subprocesses; `SmcConnection::default()` is `Unopened` and Apple Silicon never opens it; the hardware cache is intact; and `MacOsCpuReader::new()` actually got faster, since the per-construction `uname -m` spawn is now a compile-time `cfg!` on aarch64. Verified at runtime on an M1 Ultra after the fix: `doctor` reports the SMC reachable, and CPU temperature and chassis power still export.

Verification: `cargo check --lib --tests`, `cargo clippy --lib --tests -- -D warnings`, `cargo fmt --check`, and the `cpu_macos_intel` (9), `smc` (11), and `chassis` (9) test modules all pass.

inureyes:
## PR Finalization

### Tests

Reviewed test coverage for the four files in scope: `src/device/cpu_macos_intel.rs`, `src/device/readers/chassis/intel_mac.rs`, the SMC changes in `src/device/macos_native/smc.rs`, and `src/device/platform_detection.rs`. Coverage was already solid (parsers, classifiers, defaults, and arch-detection invariants), so the only change needed was fixing the LOW finding from the security review rather than adding net-new tests:

- `collection_always_produces_safe_values` in `cpu_macos_intel.rs` called `IntelCpuHardware::collect()`, which finds a gap on every macOS host (Apple Silicon has neither `hw.cpufrequency` nor `hw.l3cachesize`) and shells out to `system_profiler` on every run. Replaced it with `apply_defaults_produces_safe_values_from_a_fully_empty_record`, which exercises the same worst-case invariant directly on `IntelCpuHardware::default()` with no subprocess involved. `collect()`'s own gap-detection and gap-filling logic remains covered by the existing pure-logic tests in the same module.
- Extracted the SMC power/fan-RPM range checks in `get_system_power`, `get_cpu_package_power`, and `get_fan_readings` into two named predicates (`is_plausible_power_watts`, `is_plausible_fan_rpm`) and added boundary tests (0, upper bound, negative, NaN, +/-infinity). These bounds are the safety net the earlier review called out for the sp78 saturation case, and were previously only reachable through a live SMC connection.

`cargo test --lib` on the four scoped modules: 36/36 passing (cpu_macos_intel 9, macos_native::smc 13, readers::chassis 9, platform_detection 5).

### Documentation

Checked the platform support table in `src/lib.rs`, the `README.md` platform sections, and the Prometheus HELP strings in `src/api/metrics/chassis.rs`. All accurately scope Intel Mac support to CPU, memory, and chassis (power/fan/thermal), state plainly that there is no GPU reader, and do not claim verification on physical hardware anywhere. No changes needed.

### Lint & format

`cargo fmt --check` and `cargo clippy --lib --tests -- -D warnings` are both clean. The two new private helper functions live inside `src/device/macos_native/smc.rs`, which is entirely gated behind `#[cfg(target_os = "macos")]` at the module level, so they carry no Ubuntu dead-code risk.

### Changed files

- `src/device/cpu_macos_intel.rs`
- `src/device/macos_native/smc.rs`

Commit: `5f3e71b`

Issue #309, opened 2026-08-05
feat(service): Linux systemd service support for API mode (service subcommand framework, systemd backend, deb/PPA integration)


inureyes:
## Summary

Run `all-smi api` as a supervised systemd service on Linux. This issue delivers three integrated pieces:

1. A canonical systemd unit shipped in the repository and installed by the Debian/PPA package.
2. A new cross-platform `all-smi service` subcommand framework (shared by the macOS and Windows follow-up issues) with the Linux systemd backend as its first implementation.
3. System-level config discovery (`/etc/all-smi/config.toml`) so a daemon running as a dedicated user finds its configuration.

## Background

- API mode (`all-smi api`) is the data source for `all-smi view --hosts/--hostfile`. Cluster operators need it to start at boot, restart on failure, and log to journald. Today there is no unit file anywhere in the repo or in the packages.
- API mode already supports non-interactive configuration: env vars `ALL_SMI_API_PORT`, `ALL_SMI_API_INTERVAL_SECS`, `ALL_SMI_API_PROCESSES`, `ALL_SMI_API_SOCKET`, `ALL_SMI_API_CORS_ALLOWED_ORIGINS`, `ALL_SMI_API_MAX_SSE_SUBSCRIBERS`, a TOML config file (`all-smi config` subcommand family), and a Unix domain socket (`--socket`).
- Logging goes to stdout/stderr via `tracing_subscriber` with `RUST_LOG` filtering (`src/api/server.rs`), which is exactly what journald wants. No daemonize/fork code is needed; systemd supervises the foreground process.
- Debian packaging exists (`debian/` directory, Launchpad PPA via `.github/workflows/launchpad_ppa.yml` and `debian_build.yml`). debhelper-compat is 13, so `dh_installsystemd` is available.
- Graceful shutdown on SIGTERM is already implemented (`shutdown_signal` in `src/api/server.rs`), including the energy WAL flush.

### 1. Canonical unit file: `packaging/systemd/all-smi.service`

Single source of truth, consumed by both the deb package and the `service` subcommand (embedded via `include_str!`).

```ini
[Unit]
Description=all-smi GPU/NPU metrics exporter (API mode)
Documentation=https://github.com/lablup/all-smi
After=network-online.target
Wants=network-online.target

[Service]
Type=exec
ExecStart=/usr/bin/all-smi api
EnvironmentFile=-/etc/default/all-smi
Restart=on-failure
RestartSec=5
User=all-smi
Group=all-smi
[...]

PR #321, opened 2026-08-05
feat(service): run API mode as a launchd service on macOS (#310)

### LINKED ISSUE #310: feat(service): macOS launchd service support (Homebrew services block, launchd backend)
## Summary

Make `all-smi api` a first-class background service on macOS:

1. Add a `service do` block to the Homebrew formula so `brew services start all-smi` (per-user) and `sudo brew services start all-smi` (boot-time system daemon) work.
2. Implement the launchd backend for the `all-smi service` subcommand (framework: #309) for non-Homebrew installs.
3. Add `/Library/Application Support/all-smi/config.toml` as the system-level config candidate.
4. Validate the Apple Silicon native readers (IOReport, SMC, thermalState) in a daemon context with no logged-in GUI session.

## Background

- The formula lives in **lablup/homebrew-tap** (`Formula/all-smi.rb`), currently binary-only with no service block. The update workflow (`.github/workflows/update_homebrew_formula.yml`) rewrites only the version/url/sha256 stanzas, located by stanza name, so adding a service block does not disturb it. Verify `brew style` still passes.
- `brew services` uses the modern service DSL to generate a launchd plist (label `homebrew.mxcl.all-smi`). Without sudo it bootstraps into the `gui/$UID` domain and stops at logout; with sudo it goes into the system domain and starts at boot. Headless monitoring nodes (rack-mounted Mac minis / Studios) need the sudo variant; the docs must spell this out.
- Apple Silicon metrics use direct IOReport/SMC/NSProcessInfo calls with no sudo requirement (`src/device/macos_native/`). These are IOKit/Foundation APIs and are expected to work without a GUI session, but this has never been verified from a boot-time LaunchDaemon; the acceptance criteria make it explicit.
- Logging goes to stdout/stderr; launchd redirects it via `StandardOutPath` / `StandardErrorPath`.

### 1. Homebrew formula service block (lablup/homebrew-tap)

```ruby
service do
  run [opt_bin/"all-smi", "api"]
  keep_alive true
  log_path var/"log/all-smi.log"
  error_log_path var/"log/all-smi.log"
end
```

- No hardcoded port/interval: configuration comes from the TOML config file or env vars, keeping the generated plist stable across settings changes.
- The same block serves Linuxbrew users automatically (brew services maps it to systemd user units there).
- The tap is a separate repository, so this issue includes: the tap PR adding the block, plus README/docs changes in this repo describing `brew services` usage. Track the tap change with a checklist item here.

### 2. launchd backend for `all-smi service` (this repo)

Canonical template `packaging/launchd/com.lablup.all-smi.plist`, embedded via `include_str!`:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
<dict>
  <key>Label</key><string>com.lablup.all-smi</string>
  <key>ProgramArguments</key>
  <array><string>{BIN_PATH}</string><string>api</string></array>
  <key>RunAtLoad</key><true/>
  <key>KeepAlive</key>
  <dict><key>SuccessfulExit</key><false/></dict>
  <key>ProcessType</key><string>Background</string>
  <key>StandardOutPath</key><string>{LOG_PATH}</string>
  <key>StandardErrorPath</key><string>{LOG_PATH}</string>
</dict>
</plist>
```

- System scope (default, needs root): plist at `/Library/LaunchDaemons/com.lablup.all-smi.plist`, logs at `/var/log/all-smi/all-smi.log` (create the directory 0755 root:wheel at install). Commands: `launchctl bootstrap system <plist>`, `launchctl bootout system/com.lablup.all-smi`, restart via `launchctl kickstart -k system/com.lablup.all-smi`, status via `launchctl print system/com.lablup.all-smi` (parse state and pid).
- User scope (`--user`): plist at `~/Library/LaunchAgents/com.lablup.all-smi.plist`, domain `gui/$UID`, logs at `~/Library/Logs/all-smi/all-smi.log`. Print the logout caveat.
- Homebrew detection: `current_exe()` under `/opt/homebrew` or `/usr/local/Cellar` refuses with `use 'brew services start all-smi' (or 'sudo brew services start all-smi' for boot-time)` unless `--force`.
- Marker: an XML comment `<!-- Managed by 'all-smi service' -->` guards uninstall/overwrite, same semantics as the Linux backend.
- Elevation: system scope with euid != 0 returns NeedsElevation suggesting sudo or `--user`.

### 3. System config candidate

Append `/Library/Application Support/all-smi/config.toml` to the macOS `candidate_config_paths()` after the user-level candidates (parity with `/etc/all-smi/config.toml` from #309). Update `config path`, the runtime help block, README, and the man page.

### 4. Daemon-context validation (Apple Silicon)

Explicitly verify that a boot-time LaunchDaemon (no user logged in, SSH only) reports: GPU utilization, GPU/CPU power, SMC temperatures, thermal pressure, and P/E cluster frequencies. If any metric degrades without a GUI session, document it in the README macOS service section and file a follow-up issue.

## Out of scope

- Linux backend and subcommand framework: #309 (prerequisite).
- Windows backend: #311.
- Homebrew cask or .pkg installer.

## Acceptance criteria

- [x] The tap formula gains the service block; `brew style` passes; `brew install lablup/tap/all-smi && brew services start all-smi` works; `curl localhost:9090/metrics` returns data; the log file receives tracing output.
- [ ] `sudo brew services start all-smi` bootstraps into the system domain; the service persists across reboot without login (manual checklist).
- [ ] From a zip/tarball install: `sudo all-smi service install --now` creates the LaunchDaemon; `all-smi service status` reports running with pid; `sudo all-smi service uninstall` boots it out and removes the plist, leaving logs in place.
- [x] `all-smi service install --user --now` works without sudo; the logout behavior is documented.
- [x] A Homebrew-installed binary refuses subcommand install without `--force` and prints the brew services guidance.
- [ ] The boot-time daemon context reports the full Apple Silicon metric set, or degradations are documented with a follow-up issue.
- [ ] `/Library/Application Support/all-smi/config.toml` is honored and listed by `all-smi config path`.
- [x] SIGTERM from `launchctl bootout` takes the graceful shutdown path (energy WAL flush visible in the log).
- [x] `cargo fmt --check`, `cargo clippy`, and `cargo test` pass; unit tests cover plist rendering, `launchctl print` parsing, and Homebrew-path detection.
- [x] The README "Running as a service" section gains the macOS subsection covering both brew services variants and the subcommand; man page updated.
- [x] Fully integrated: the backend is selected automatically on macOS by the service subcommand dispatch from #309; no dead template files.

## Test plan

- Unit tests with fixture `launchctl print` outputs (running, stopped, not found).
- CI: a macOS runner job running the plist render tests and, where the runner permits, the user-scope install/status/uninstall flow behind an env gate (GitHub macOS runners may lack a bootstrappable gui domain; if `launchctl print gui/$UID` fails, the gated test must skip cleanly).
- Manual checklist in the PR: Apple Silicon reboot test over SSH, brew services in both domains.

## Related issues

- Linux (framework prerequisite): #309
- Windows: #311

inureyes:
## Summary

Replaces the macOS arm of `service_cmd::backend()`, which until now returned `NotSupported` pointing at #310, with a real launchd backend, and adds `/Library/Application Support/all-smi/config.toml` as the macOS system-wide config candidate. The #309 contract is untouched: same `ServiceBackend` methods, same scope semantics, same exit codes, same error variants.

Along the way this fixes a pre-existing bug that made API mode's graceful shutdown unreachable on SIGTERM, which is exactly how `launchctl bootout` and `systemctl stop` end a service.

## What changed

- **`packaging/launchd/com.lablup.all-smi.plist`** (new). The canonical job definition, a self-contained system LaunchDaemon an operator can also copy into `/Library/LaunchDaemons` by hand. Same role `packaging/systemd/all-smi.service` plays on Linux.
- **`src/service_cmd/plist.rs`** (new, 370 lines) + `plist_tests.rs`. Embeds that file with `include_str!` and rewrites `ProgramArguments`, the log paths, and the account keys per scope. Walks key/value pairs rather than lines, because a plist value can be a multi-line `<array>`/`<dict>` and dropping a key has to drop its whole value with it. XML-escapes paths and rejects control characters and non-POSIX account names.
- **`src/service_cmd/launchctl.rs`** (new, 349 lines) + `launchctl_tests.rs`. Layout resolution (plist path, log path, launchd domain, service target), `launchctl` invocation, and parsing of `launchctl print` and `launchctl print-disabled`.
- **`src/service_cmd/launchd.rs`** (new, 319 lines) + `launchd_tests.rs`. Verb policy and the on-disk half: the managed-marker guard, the atomic `0644` plist write, log-directory creation.
- **`src/service_cmd/mod.rs`**. The macOS `backend()` arm now returns `LaunchdBackend`. The user-scope persistence note and the "where settings live" line are now per-platform, because `loginctl enable-linger` and "the environment file" do not exist on macOS.
- **`src/service_cmd/detect.rs`**. The Homebrew refusal gains a macOS-only hint naming `sudo brew services start all-smi`.
- **`src/common/paths.rs`**. `/Library/Application Support/all-smi/config.toml` added as a Tier 2 candidate, ordered after every per-user one.
- **`src/main.rs`**. The API-mode shutdown fix, below.
- **`README.md`**, **`docs/man/all-smi.1`**. macOS subsections appended into the slots #309 left, plus the config-candidate tables.
- **`.github/workflows/ci.yml`**. New gated `launchd-service` job on `macos-14`.

## How launchd differs from systemd, and how the verbs map

launchd has no separate "enabled at boot" state. A plist in `LaunchDaemons`/`LaunchAgents` is bootstrapped automatically at boot or login and `RunAtLoad` starts it from there. So:

- `install` without `--now` writes the plist and stops. That *is* "enabled at boot, not running yet".
- `install --now` boots the job out and back in, because launchd caches a loaded job's definition and bootstrapping over it fails rather than replacing it.
- `stop` boots the job out and leaves the plist, so the service returns at the next boot, matching `systemctl stop`.
- `status` reads plist presence from disk for `installed`, since `launchctl print` only knows about loaded jobs, and reads `launchctl print-disabled` for `enabled`.
- `install` runs `launchctl enable` to clear a persistent disable override, which outlives both the plist and a reboot. `uninstall` deliberately does not `disable`, for the same reason.

## The user-scope trap, measured rather than assumed

The user-scope render drops `UserName`, `GroupName`, and `InitGroups`. This is the launchd analogue of #309's `ProtectKernelModules=` problem, but **the failure mode is the opposite one**, and I want to be precise about it rather than repeat the systemd rationale by analogy.

systemd refuses a user unit it cannot apply: `ProtectKernelModules=` dies at `218/CAPABILITIES` before `ExecStart`. launchd does not refuse. Probe agent bootstrapped into `gui/501` on macOS 26.6 (Darwin 25.6), program printed `id`:

| Plist keys | `launchctl bootstrap` | Effective uid/gid |
|---|---|---|
| none | succeeds | `501` / `20` |
| `UserName root`, `GroupName wheel` | succeeds | `501` / `20` |
| `InitGroups` + `UserName root` | succeeds | `501` / `20` |

They are silently ignored. So they are dropped **not** to avoid a crash, but because keeping them would ship a plist that reads, to anyone auditing what runs privileged on the machine, as a root job when it is not one; and because `launchd.plist(5)` documents them as requiring root without documenting the fallback, so it is unspecified behaviour that is free to become fatal later. Since no bootstrap will ever catch a regression here, a render-time test asserts their absence, and a second test asserts every top-level key in the shipped plist is classified by one of the two lists so adding one without deciding fails in CI.

`HardResourceLimits` is deliberately absent from the template for the mirror-image reason (only root can raise a hard limit, lowering one buys nothing). `SoftResourceLimits` is kept in both scopes, since raising a soft rlimit up to the inherited hard limit needs no privilege.

`--service-user` sets `UserName` and drops `GroupName` rather than mirroring the account name the way the systemd renderer does: macOS has no convention that an account owns an eponymous group, so omitting the key makes launchd use the primary group straight from the password database.

## Bug fix: API mode never took its graceful shutdown path

`run_command` installed a SIGTERM handler calling `std::process::exit(0)`. It won the race against `run_api_mode`'s post-serve cleanup, so **every** SIGTERM dropped the last batch of accumulated Joules and left a stale Unix socket behind. Not an edge case for a service: SIGTERM is how both `launchctl bootout` and `systemctl stop` end one, so it happened on every restart.

`Api` now owns its shutdown the way `Record` already did, including through the `[general].default_mode` redispatch (the recursive call cannot uninstall a handler the outer call already spawned), and the collectors are torn down after `run_api_mode` returns like `view` already does.

Before: SIGTERM produced no `energy WAL: shutdown requested` line. After: it appears, and the process exits in 0.45 s.

## Test plan

Scoped commands only (the orchestrator runs the full suite):

- [x] `cargo fmt --check`
- [x] `cargo clippy --bin all-smi -- -D warnings` (run separately on purpose. The library target caught nothing, but the binary target flagged `SERVICE_NAME` as dead: it is reachable only from the systemd backend, which is not compiled in a non-Linux non-test build. Per-target dead-code blindness is real here and the scoped library check does not see it.)
- [x] `cargo test --lib service_cmd::`: 110 passed
- [x] `cargo test --lib common::paths`: 17 passed
- [x] `plutil -lint` on the rendered plist, and `man ./docs/man/all-smi.1` renders clean

Live on an M1 Ultra, macOS 26.6, **user scope only** (see below):

- [x] `status` before install exits 3 and reports "not installed"
- [x] `install --user` writes the plist, creates `~/Library/Logs/all-smi`, and leaves the job unloaded; `status` reports `installed, enabled, stopped` / `not loaded`, exit 3
- [x] `start` bootstraps it; `status` reports running with a pid
- [x] `install --user --now`, `restart`, `stop`, `uninstall`, and a second `install` over our own plist (idempotent)
- [x] `curl localhost:9090/metrics` serves 72 `all_smi_*` lines from the LaunchAgent
- [x] `stop` (i.e. `launchctl bootout`) reaches the final energy-WAL flush, visible in the log
- [x] A hand-written plist is refused by both `install` and `uninstall`, the file is left untouched, and `--force` lifts the refusal
- [x] `launchctl disable` is reported as `"enabled": false` by `status` and cleared by `install`
- [x] System scope without root refuses with "requires root" and writes nothing to `/Library/LaunchDaemons`
- [x] `all-smi config path` lists `/Library/Application Support/all-smi/config.toml` as candidate 3, ordered last

### Apple Silicon metrics under launchd

The LaunchAgent's metric **name set is byte-identical** to a foreground `all-smi api` (`diff` of the sorted metric names: no difference). Values sampled from the agent:

```
all_smi_gpu_utilization             17.19
all_smi_gpu_power_consumption_watts  0.56
all_smi_gpu_temperature_celsius     52
all_smi_cpu_temperature_celsius      65
all_smi_ane_power_watts               0
all_smi_thermal_pressure_info         1
all_smi_cpu_p_cluster_frequency_mhz 3223
all_smi_cpu_e_cluster_frequency_mhz 1978
all_smi_chassis_power_watts         48.12
```

IOReport, the SMC, and `NSProcessInfo.thermalState` all resolve from a launchd context with no TTY and no sudo prompt. Cluster frequencies are non-zero thanks to #317.

Two environment notes worth recording:

1. **`ProcessType Background` costs startup time.** Foreground bind: 0.6 s. Under `taskpolicy -b`: 2.9 s. As a LaunchAgent: 6.3 s. Enumerating the IOReport channel list is the expensive part and background QoS runs it on the E-cores. Kept anyway: a GPU monitor should not steal P-cores from the job it is watching. This is why every readiness wait in CI polls `/metrics` rather than `service status`, which reports running the moment launchd spawns the program.
2. **A binary on an external volume hangs under launchd.** The first attempt pointed the plist at `target/debug/all-smi` on an external volume; the job spawned and blocked indefinitely in `dyld` `open()` on its own binary (TCC gate for launchd-spawned processes on removable volumes). Not a code issue; re-verified from an internal-disk path. Worth knowing if anyone installs from a mounted volume.

## Not verified in this environment

These were explicitly out of scope for this run and their acceptance boxes are left unchecked on the issue.

- **System-domain LaunchDaemon.** No `/Library/LaunchDaemons` write and no `sudo launchctl bootstrap system`. The system-scope code path is therefore exercised only by unit tests plus the live "refuses without root, writes nothing" check.
- **Reboot persistence**, in either the subcommand or the `sudo brew services` form. Not reachable without the system domain.
- **`sudo brew services start all-smi`** and the tap formula end to end. Nothing was pushed to `lablup/homebrew-tap`; see the checklist below.
- **The Homebrew refusal live.** `detect::classify` is unit-tested for all three Homebrew prefixes and a new test asserts the macOS hint text, but no binary was placed under `/opt/homebrew` to trigger it for real.
- **`/Library/Application Support/all-smi/config.toml` being loaded.** Listing is verified live; creating the file needs root. The load path itself is the same `discover_existing_config()` used by the Linux `/etc` candidate, and is covered by unit tests for presence, ordering, and never being the `config init` target.
- **Full `cargo test`.** Only the two scoped filters above were run here.
- **Issue item 4 (daemon context with no GUI session) is only partly reached.** A `gui/$UID` LaunchAgent still belongs to a login session. What is genuinely demonstrated is that the native readers need neither a controlling terminal nor sudo and work at background QoS. Whether they behave identically from a boot-time LaunchDaemon with nobody logged in needs a real headless daemon. **No degradation was found in the part that was reachable**, so no follow-up issue is filed; if the system-domain test later shows one, that is the moment to file it.

## Checklist for a human: the tap `service do` block

**Nothing was pushed to `lablup/homebrew-tap`.** Apply this by hand, the way PR #313 handled the Intel formula stanza. Against `Formula/all-smi.rb` at its current HEAD:

```diff
@@ -27,6 +27,13 @@
     man1.install "all-smi.1"
   end
 
+  service do
+    run [opt_bin/"all-smi", "api"]
+    keep_alive true
+    log_path var/"log/all-smi.log"
+    error_log_path var/"log/all-smi.log"
+    process_type :background
+  end
+
   test do
     output = shell_output("#{bin}/all-smi --version")
     assert_match(/all-smi\s+#{version}/, output)
```

Verified locally against a scratch copy, never the tap: `ruby -c` passes, and `brew style` reports exactly the same 4 offenses on the patched file as on the unpatched upstream file (all four are artifacts of linting a bare `.rb` outside a tap directory: Sorbet sigils, frozen-string-literal, class documentation). The service block itself adds none.

Two notes on the block versus the version proposed in the issue:

- `process_type :background` is added so a brew-managed service and a subcommand-installed one get the same CPU scheduling. Without it, which install method you used silently changes the QoS class, which is an unpleasant thing to debug. `Homebrew::Service#process_type` accepts `:background`.
- No hardcoded port or interval, and no `require_root`, so both `brew services start all-smi` and `sudo brew services start all-smi` keep working. The update workflow rewrites only the version/url/sha256 stanzas by name, so this block does not disturb it.

Closes #310

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -351,6 +351,224 @@ jobs:
           sudo rm -f "$UNIT"
           sudo systemctl daemon-reload
 
+  # ═══════════════════════════════════════════════════════════════
+  # Job: launchd service smoke test (issue #310)
+  # ═══════════════════════════════════════════════════════════════
+  # Runs on Apple Silicon so the native readers (IOReport, SMC,
+  # thermalState) are actually exercised from a launchd context, not
+  # just the plist renderer.
+  #
+  # Split into two tiers because a hosted runner may have no
+  # bootstrappable GUI domain:
+  #
+  #  1. Render tier. `install --user` without `--now` writes the plist
+  #     and touches launchd not at all, so plist validity, the managed
+  #     marker, the user-scope key drops, and both refusal paths are
+  #     checkable anywhere. Always runs.
+  #  2. Lifecycle tier. Gated on `launchctl print gui/$UID` succeeding.
+  #     Skips with a notice rather than failing where the domain is
+  #     absent.
+  #
+  # The system domain is deliberately never touched: no
+  # /Library/LaunchDaemons, no `sudo launchctl bootstrap system`.
+  # ───────────────────────────────────────────────────────────────
+  launchd-service:
+    name: launchd Service Smoke Test
+    runs-on: macos-14
+    needs: test
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+
+      - name: Setup protoc
+        uses: ./.github/actions/setup-protoc
+        with:
+          version: ${{ env.PROTOC_VERSION }}
+          platform: osx-aarch_64
+
+      - name: Cache cargo
+        uses: actions/cache@v5
+        with:
+          path: |
+            ~/.cargo/registry
+            ~/.cargo/git
+            target
+          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-cargo-
+
+      - name: Build all-smi
+        run: cargo build --bin all-smi
+
+      - name: Probe for a bootstrappable GUI domain
+        id: probe
+        run: |
+          if launchctl print "gui/$(id -u)" >/dev/null 2>&1; then
+            echo "gui_domain=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "gui_domain=false" >> "$GITHUB_OUTPUT"
+            echo "::notice::no gui/\$UID domain on this runner; running the render tier only"
+          fi
+
+      # ── Render tier: no launchd domain required ──
+      #
+      # launchd accepts a LaunchAgent carrying UserName/GroupName and
+      # silently ignores them, so a regression that leaves those keys in
+      # a user-scope plist cannot be caught by bootstrapping it. It has
+      # to be caught here, on the rendered file.
+      - name: Rendered plist is valid and correctly scoped
+        run: |
+          set -eEux
+          BIN="$PWD/target/debug/all-smi"
+          PLIST="$HOME/Library/LaunchAgents/com.lablup.all-smi.plist"
+
+          diagnose() {
+            set +x
+            echo "::group::diagnostics: rendered LaunchAgent plist"
+            echo "--- $PLIST ---"
+            cat "$PLIST" 2>&1 || echo "(absent)"
+            echo "--- plutil ---"
+            plutil -lint "$PLIST" 2>&1 || true
+            echo "::endgroup::"
+          }
+          trap diagnose ERR
+
+          "$BIN" service status --user && rc=0 || rc=$?
+          test "$rc" -eq 3
+
+          "$BIN" service install --user
+          test -f "$PLIST"
+
+          # launchd parses this with CFPropertyList; a stray `--` inside
+          # the marker comment would make the whole job unloadable.
+          plutil -lint "$PLIST"
+          grep -qF "<!-- Managed by 'all-smi service' -->" "$PLIST"
+          grep -qF "<string>api</string>" "$PLIST"
+          grep -qF "$HOME/Library/Logs/all-smi/all-smi.log" "$PLIST"
+
+          # Keys a gui/$UID domain cannot honour must be absent.
+          for key in UserName GroupName InitGroups; do
+            if grep -qF "<key>${key}</key>" "$PLIST"; then
+              echo "::error::${key} must not survive into a user-scope plist"
+              exit 1
+            fi
+          done
+
+          # Installed but not bootstrapped is the documented state here.
+          "$BIN" service status --user && rc=0 || rc=$?
+          test "$rc" -eq 3
+          "$BIN" service status --user --json | grep -q '"installed": true'
+
+          "$BIN" service uninstall --user
+          test ! -f "$PLIST"
+
+      - name: Refuse an unelevated system install and a foreign plist
+        run: |
+          set -eEux
+          BIN="$PWD/target/debug/all-smi"
+          PLIST="$HOME/Library/LaunchAgents/com.lablup.all-smi.plist"
+
+          # System scope must refuse before writing anything at all.
+          "$BIN" service install && rc=0 || rc=$?
+          test "$rc" -eq 1
+          "$BIN" service install 2>&1 | grep -q "requires root"
+          test ! -f /Library/LaunchDaemons/com.lablup.all-smi.plist
+
+          printf '%s\n' '<?xml version="1.0" encoding="UTF-8"?>' \
+            '<plist version="1.0"><dict><key>Label</key><string>hand written</string></dict></plist>' \
+            > "$PLIST"
+          "$BIN" service install --user && rc=0 || rc=$?
+          test "$rc" -eq 1
+          "$BIN" service uninstall --user && rc=0 || rc=$?
+          test "$rc" -eq 1
+          grep -q "hand written" "$PLIST"
+
+          "$BIN" service uninstall --user --force
+          test ! -f "$PLIST"
+
+      # ── Lifecycle tier: needs a real gui/$UID domain ──
+      - name: User-scope lifecycle
+        if: steps.probe.outputs.gui_domain == 'true'
+        run: |
+          set -eEux
+          BIN="$PWD/target/debug/all-smi"
+          PLIST="$HOME/Library/LaunchAgents/com.lablup.all-smi.plist"
+          LOG="$HOME/Library/Logs/all-smi/all-smi.log"
+          TARGET="gui/$(id -u)/com.lablup.all-smi"
+
+          # A job that fails to spawn under launchd reports almost
+          # nothing through the exit code alone, so dump the three
+          # places the reason can be before giving up.
+          diagnose() {
+            set +x
+            echo "::group::diagnostics: launchd user-scope service"
+            echo "--- $PLIST ---"
+            cat "$PLIST" 2>&1 || echo "(absent)"
+            echo "--- launchctl print $TARGET ---"
+            launchctl print "$TARGET" 2>&1 || true
+            echo "--- launchctl print-disabled gui/$(id -u) ---"
+            launchctl print-disabled "gui/$(id -u)" 2>&1 | grep -i all-smi || true
+            echo "--- $LOG ---"
+            cat "$LOG" 2>&1 || echo "(absent)"
+            echo "--- service status ---"
+            "$BIN" service status --user 2>&1 || true
+            echo "::endgroup::"
+          }
+          trap diagnose ERR
+
+          # `service status` reports running the moment launchd spawns
+          # the program, which is well before it has finished coming up:
+          # enumerating the IOReport channel list dominates startup and
+          # ProcessType=Background runs it at background QoS. Every
+          # readiness wait below therefore polls the metrics endpoint,
+          # not the process state. Polling status instead races, and the
+          # race is invisible until an assertion downstream fails for an
+          # unrelated-looking reason.
+          ready() {
+            for _ in $(seq 1 60); do
+              curl -sf --max-time 5 localhost:9090/metrics >/dev/null && return 0
+              sleep 2
+            done
+            echo "::error::the exporter never served /metrics"
+            return 1
+          }
+
+          "$BIN" service install --user --now
+          ready
+          "$BIN" service status --user
+          "$BIN" service status --user --json | grep -q '"running": true'
+          "$BIN" service status --user --json | grep -q '"enabled": true'
+          curl -sf --max-time 10 localhost:9090/metrics | grep -q '^all_smi_'
+          grep -q "API server listening" "$LOG"
+
+          "$BIN" service restart --user
+          ready
+
+          # `stop` boots the job out, which is a SIGTERM. The final
+          # energy-WAL flush has to be reached before launchd's
+          # ExitTimeOut turns it into a SIGKILL (issues #191, #310).
+          "$BIN" service stop --user
+          sleep 3
+          grep -q "energy WAL: shutdown requested" "$LOG"
+          "$BIN" service status --user && rc=0 || rc=$?
+          test "$rc" -eq 3
+          "$BIN" service status --user --json | grep -q '"installed": true'
+
+          # A persistent disable override outlives the plist and a
+          # reboot, so `install` has to clear it and `status` has to see
+          # it.
+          launchctl disable "$TARGET"
+          "$BIN" service status --user --json | grep -q '"enabled": false'
+          "$BIN" service install --user
+          "$BIN" service status --user --json | grep -q '"enabled": true'
+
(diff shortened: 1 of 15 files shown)


inureyes:
### CI reproduced the user-scope lifecycle independently

The `macos-14` runner turned out to have a bootstrappable `gui/$UID` domain, so the gated lifecycle tier ran rather than skipping. Full run: https://github.com/lablup/all-smi/actions/runs/31000492760

- `Test Suite` (Linux) pass: full `cargo test --verbose`, `cargo fmt --check`, `cargo clippy -- -D warnings`. That covers the plist renderer, the `launchctl print` / `print-disabled` parsers, and the layout resolution, all of which compile under `cfg(test)` off macOS.
- `launchd Service Smoke Test` (macos-14) pass in 3m09s, both tiers. Every assertion executed on a clean machine, including the two that matter most:
  - `grep -q 'energy WAL: shutdown requested'` after `service stop`, so the SIGTERM fix in `main.rs` holds outside my machine.
  - `launchctl disable` reported as `"enabled": false` and cleared by the next `install`.
  - Teardown checked: plist gone, `launchctl print` exits 113, log file deliberately still present.
- `systemd Service Smoke Test` still passes, so the shared changes in `mod.rs`, `detect.rs`, and `main.rs` did not regress the Linux backend.

The acceptance criterion for `cargo fmt`/`clippy`/`test` is now ticked on the issue. The remaining unticked ones all need the system launchd domain or the tap, which this run deliberately did not touch; they are enumerated under "Not verified in this environment" above.

Issue #341, opened 2026-08-06
chore: remove the dead vendored OpenSSL dependency from Cargo.toml


inureyes:
## Background

`Cargo.toml` carries two target-conditional blocks that pull in a vendored OpenSSL:

```toml
[target.'cfg(target_env = "musl")'.dependencies]
openssl = { version = "0.10.80", features = ["vendored"] }

[target.'cfg(all(target_arch = "aarch64", target_env = "gnu"))'.dependencies]
openssl = { version = "0.10.80", features = ["vendored"] }
```

Nothing in the tree uses OpenSSL. The TLS stack is rustls. These two entries are the only reason OpenSSL is in the dependency graph at all, and the `vendored` feature makes every affected build compile OpenSSL from source.

`.github/workflows/release.yml` builds `x86_64-unknown-linux-musl`, `aarch64-unknown-linux-gnu`, and `aarch64-unknown-linux-musl`, which are exactly the targets those two `cfg`s cover. **Every release therefore compiles OpenSSL from source three times for a dependency nothing links against.** That is the cost this issue removes, and it has nothing to do with containers.

## Rescope: the Docker image is being removed (#342)

This issue was originally filed as a container bug. `docker build --platform linux/arm64` failed because vendored OpenSSL needs perl's `FindBin`, which `rust:1.96-slim` does not ship, and `docker-check` never caught it because it builds `linux/amd64` only, where the aarch64 `cfg` does not apply.

That framing is obsolete. The maintainer has decided to **delete** the Docker image rather than fix it, tracked in #342. The survey behind that decision: no workflow publishes the image (no `docker/login-action`, no `docker push`, no `ghcr.io` reference anywhere under `.github/`), there is no `k8s/`, `helm/`, or `charts/` directory, `README.md` documents no container usage path, and the one `examples/docker-compose.yml` is referenced nowhere in the tree.

The container evidence is kept below, because it is how the dead dependency was found and because it remains a correct description of the mechanism. The container-related acceptance criteria have been removed from this issue: they refer to artifacts #342 deletes, so they cannot be satisfied and should not be pretended to be. Removing the OpenSSL blocks stands on its own merit as a release-build change.

## How it surfaced

Reproduced at `d171d54` on a native arm64 host:

```
$ docker build --platform linux/arm64 -t all-smi-arm64-check .
...
#17 15.22    Compiling openssl-sys v0.9.117
#17 16.50 error: failed to run custom build command for `openssl-sys v0.9.117`
[...]

PR #344, opened 2026-08-06
chore: remove the Docker image and its CI job

### LINKED ISSUE #342: chore: remove the Docker image and its CI job
## Decision

**The Docker image is being removed, not fixed.**

This issue was originally filed as "fix: the Docker image builds but the binary cannot start", proposing to repair the image. The maintainer decided against that. The defect evidence below is unchanged and is exactly what motivated the decision, but the title, scope, and acceptance criteria have been rewritten. A comment on this issue records the change of direction so it is visible in the timeline rather than only in a silently rewritten body.

The reasoning: the image has never worked on any architecture, nothing in the repository or its release process consumes it, and nobody noticed in the months it has been broken. Fixing it would mean adopting and maintaining a container deployment path no one has asked for, and paying its CI cost on every pull request. Removing it costs nothing anyone currently has.

This issue stays open as the record of why there is no `Dockerfile`.

## The defects (kept, they are the evidence)

`docker build` succeeded and the resulting image could not run the binary. CI was green on it the whole time, because `docker-check` built the image and never started it.

Reproduced at `d171d54` on `linux/amd64`, the only platform `docker-check` built:

```
$ docker build --platform linux/amd64 -t all-smi-runcheck .
... exit 0, image built successfully

$ docker run --rm --platform linux/amd64 all-smi-runcheck all-smi --version
/usr/local/bin/all-smi: error while loading shared libraries: libdrm.so.2: cannot open shared object file: No such file or directory
exit=127

$ docker run --rm --platform linux/amd64 --entrypoint sh all-smi-runcheck -c 'ldd /usr/local/bin/all-smi'
/usr/local/bin/all-smi: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.38' not found (required by /usr/local/bin/all-smi)
/usr/local/bin/all-smi: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.39' not found (required by /usr/local/bin/all-smi)
	libdrm.so.2 => not found
	libdrm_amdgpu.so.1 => not found
	...
```

### Defect 1: glibc mismatch between the build and runtime stages

`Dockerfile:2` built on `rust:1.96-slim` (Debian 13 trixie, glibc 2.41) and `Dockerfile:37` ran on `debian:bookworm-slim` (Debian 12 bookworm, glibc 2.36). The binary picked up symbols versioned `GLIBC_2.38` and `GLIBC_2.39` that bookworm cannot resolve.

### Defect 2: runtime shared libraries were never installed

`Cargo.toml` declares `libamdgpu_top` under `cfg(all(target_os = "linux", not(target_env = "musl")))`, which is active on every non-musl Linux target, and `libamdgpu_top` links `libdrm`. The builder stage installed `libdrm-dev` and `libdrm-amdgpu1`; the runtime stage installed `ca-certificates` and nothing else, so `libdrm.so.2` and `libdrm_amdgpu.so.1` were absent at run time.

The split was historical. Commit `04c1d29` ("fix: add AMD GPU dependencies to Dockerfile builder stage", 2025-11-25, released in v0.12.0) added exactly those two packages in a two-line diff that touched only the builder stage. The runtime stage was never given the matching runtime libraries.

### Defect 3: arm64 did not build at all

Separately, `docker build --platform linux/arm64` failed outright on a vendored OpenSSL dependency nothing used (#341). So the image was broken on amd64 at run time and broken on arm64 at build time. #341 has since removed that dependency on its own merit, because it also made every release compile OpenSSL from source three times. That change stands independently of this one.

### Defect 4: the build context was never filtered

No `.dockerignore` is tracked in this repository. `.gitignore:69` carries a blanket `.*` rule, so any `.dockerignore` a developer created locally was silently never committed. CI checks out from git, so **every `docker-check` run built with no `.dockerignore` at all**, and the build context therefore included `target/`, `.git/`, and everything else such a file was meant to exclude. `tests/docker_build_context_test.rs` reads it defensively (`.unwrap_or_default()`), so its absence never surfaced as a test failure either.

There is nothing to delete for this one, since the file was never in the repository. It is recorded because it is the sort of thing that would have bitten whoever eventually tried to publish the image, and because it reinforces rather than weakens the case for removal.

### Why CI never caught any of it

`docker-check` checked out, set up Buildx, and ran `docker/build-push-action@v7` with `push: false`. There was no step after the build. The job was green on every run while producing an artifact that has never started. Since #339 it also ran on pull requests, so it spent 313-354s (mean 326s, per the measurements recorded in #339) on every PR without ever executing what it built.

## Consumption survey: nothing uses the image

Every claim re-verified on the implementation branch, not only at `d171d54`:

- **No workflow publishes it.** The only Docker reference under `.github/` was `docker/build-push-action@v7` in `ci.yml` with `push: false`. There is no `docker/login-action`, no `docker push`, and no `ghcr.io` reference anywhere under `.github/`. The other four workflows (`release.yml`, `debian_build.yml`, `launchpad_ppa.yml`, `update_homebrew_formula.yml`) contain no docker reference at all.
- **No orchestration manifests.** There is no `k8s/`, `helm/`, or `charts/` directory.
- **`README.md` documents no container usage path.** `docker` appears in it only in two changelog lines (`README.md:1593`, `README.md:1612`).
- **`examples/docker-compose.yml` is referenced nowhere.** No README link, no CI job, no other document. All three of its services run `image: all-smi:latest` built from this `Dockerfile` via `build.context: ..`, so all three were broken. It is also stale on its own terms: it sets `HOST_PROC_PATH: /host/proc`, and `HOST_PROC_PATH` appears nowhere else in the tree because the binary never reads it. `src/device/container_utils.rs` finds the host procfs by probing a hardcoded list (`/host/proc`, `/hostproc`, `/proc_host`) instead. So the file documents a configuration knob that does not exist, which is hard to square with anyone having run it.
- **The only `docker build` of this repository in the whole tree** is the `docker-build-container` target at `Makefile:72-74`. It was added in `f3745e2` (#31), the same commit that introduced the container test harness, and nothing has referenced it since: it is absent from the `.PHONY` list and from `make help`, and no script, workflow, or document invokes it. `DEVELOPERS.md` documents the raw `docker build -t all-smi:latest .` command rather than the make target, so even the documentation did not know the target existed. It is invocable by anyone who reads the Makefile and effectively undiscoverable otherwise, which makes it dead in practice rather than merely undocumented.
- **No job depends on `docker-check`.** In `ci.yml` the literal string `docker-check` appears only on lines 1124 and 1131 (inside its own comment header) and 1166 (the job key). Every other job is `needs: test`.
- **`main` is not branch-protected** (`gh api repos/lablup/all-smi/branches/main/protection` returns 404), so no required status check names `Docker Build Check`.

## Scope: what gets deleted

- `Dockerfile`
- `examples/docker-compose.yml`
- `tests/docker_build_context_test.rs`. Added by PR #322 specifically to guard the Docker build context. With no `Dockerfile` it has no subject, and it would fail outright: two of its seven tests do `.expect("Dockerfile must exist")`. It is deleted deliberately, not swept along quietly.
- The `docker-check` job in `.github/workflows/ci.yml`, together with the comment header documenting its cost tradeoffs.
- The `docker-build-container` target in `Makefile`.
- In `DEVELOPERS.md`: the `docker build -t all-smi:latest .` instruction, the `Docker Check` line in the CI description, and the `Docker images` line in the release description (the latter was already false, since nothing publishes them).

There is no tracked `.dockerignore` in this repository (see Defect 4), so there is nothing to remove there. A developer with a stray local copy should delete it by hand; it was never part of the build CI ran.

## What is deliberately kept

- **All container-awareness code under `src/`.** all-smi detects when it is *running inside* a container (`/.dockerenv`, cgroup parsing, `ContainerRuntime::Docker`, Docker-aware disk filtering). That is a product feature and has nothing to do with shipping an image.
- **The container test harness.** The shell scripts under `tests/` and the `docker-dev`, `docker-test-container-api`, and `docker-test-container-view` targets in `Makefile` run all-smi inside stock `rust:1.88` containers with the source bind-mounted, to exercise that container-awareness feature. None of them builds this repository's image.
- **Changelog and report history.** `README.md`'s changelog lines, `debian/changelog`, and the `TECHNICAL_REPORTS/` entries for #322 and #339 record what happened and stay true.

## What is lost

Container deployment is no longer supported. Anyone who was building the image locally has these instead, all of which work today and are exercised by CI:

- Pre-built release binaries from the GitHub releases page
- The Homebrew tap (macOS and Linux)
- The Debian package and the Ubuntu PPA
- `cargo install all-smi`
- `all-smi service`, which runs API mode as a supervised background service under systemd, launchd, or the Windows SCM

If container images become a real requirement later, they should be reintroduced as a deliberate, published, smoke-tested artifact rather than restored from this one.

## Acceptance Criteria

- [ ] `Dockerfile`, the `docker-check` job in `.github/workflows/ci.yml`, `examples/docker-compose.yml`, `tests/docker_build_context_test.rs`, and the `docker-build-container` `Makefile` target are removed.
- [ ] `DEVELOPERS.md` no longer instructs anyone to build a container image and no longer claims releases publish one.
- [ ] `.github/workflows/ci.yml` still parses as valid YAML, and no remaining job references `docker-check`.
- [ ] The test suite still passes with `tests/docker_build_context_test.rs` gone.
- [ ] The loss of container deployment is stated somewhere a user will actually look, rather than left to be inferred from an absent file.
- [ ] Container-awareness source code under `src/` and the container test harness under `tests/` are untouched.

inureyes:
## Summary

Deletes the Docker image and everything that existed to build it. The image has never worked on any architecture, and nothing in the repository or its release process consumes it. The maintainer's decision was to remove rather than repair, so this PR does not attempt to fix the glibc mismatch or the missing runtime libraries recorded in #342.

**Container deployment is no longer supported.** Anyone who was building the image locally has the release binaries, the Homebrew tap, the Debian package and Ubuntu PPA, `cargo install all-smi`, and `all-smi service` for running API mode supervised under systemd, launchd, or the Windows SCM. All of those work today and are exercised by CI, which the image never was.

## What changed

Removed:

- `Dockerfile`
- `examples/docker-compose.yml`
- `tests/docker_build_context_test.rs`
- The `docker-check` job in `.github/workflows/ci.yml`, together with the 57 line cost-analysis comment header written for it (lines 1108-1184, leaving the file at 1107 lines)
- The `docker-build-container` target in `Makefile`

Documentation:

- `DEVELOPERS.md`: removed the `docker build -t all-smi:latest .` instruction, the `Docker Check` CI bullet, and the `Docker images` release bullet. Kept the `make docker-dev` / `docker-test-container-api` / `docker-test-container-view` documentation, with a note clarifying that those targets run all-smi inside a stock `rust:1.88` image to exercise container awareness and do not build a project image.
- `README.md`: added a short note at the end of the Installation section. The section lists six install options and previously said nothing about containers, so a reader expecting a seventh was left to infer its absence.

## Deleting the test was deliberate

`tests/docker_build_context_test.rs` is not collateral damage. PR #322 added it a day earlier specifically to guard the Docker build context after #309 broke `main`. With no `Dockerfile` it has no subject, and it would fail outright rather than pass vacuously: two of its seven tests do `.expect("Dockerfile must exist")`. It has to go in the same commit as the `Dockerfile`, and it is called out here so the loss of that guard is visible rather than quietly absorbed.

## Two documentation defects of different kinds

Worth distinguishing, because only one of them is caused by this PR:

- `DEVELOPERS.md` line 408, `3. **Docker Check**: Validates Docker image build`, described a job that this PR deletes. **Invalidated by this change.**
- `DEVELOPERS.md` line 419 listed `Docker images` among the things a release publishes. **This was already false before this change**, and has been for as long as the line existed: no workflow has ever published an image. There is no `docker/login-action`, no `docker push`, and no `ghcr.io` reference anywhere under `.github/`. Removing it corrects a claim that was never true, rather than one this PR made untrue.

Neither removal required renumbering: line 408 was the last item of a three-item ordered list, and line 419 was a bullet in an unordered sub-list.

**Flagged but deliberately not fixed:** that same `DEVELOPERS.md` CI section names three jobs when `ci.yml` actually has seven (`packaging-sync`, `systemd-service`, `launchd-service`, and `windows-service` are all missing). That staleness predates this PR and fixing it is scope creep, so it is left for a docs pass.

## Survey notes worth recording

Three findings from the consumption sweep, all of which strengthen the case that nothing ever used this:

- **`examples/docker-compose.yml` set `HOST_PROC_PATH: /host/proc`, and the binary never reads that variable.** It appears nowhere else in the tree. `src/device/container_utils.rs:147-153` locates the host procfs by probing a hardcoded list (`/host/proc`, `/hostproc`, `/proc_host`) instead. An example configuring a knob that does not exist is hard to square with anyone having run it.
- **No `.dockerignore` was ever tracked.** `.gitignore:69` carries a blanket `.*` rule, so any local copy was silently never committed, and CI checks out from git. Every `docker-check` run therefore built with no exclusions at all, sending `target/` and `.git/` into the build context. Nothing to delete here, since it was never in the repository.
- **The `docker-build-container` Makefile target was dead, not merely undocumented.** It was the only `docker build` of this repository anywhere in the tree, added in `f3745e2` (#31) and never referenced since: absent from both `.PHONY` and `make help`, invoked by no script, workflow, or document. `DEVELOPERS.md` documented the raw `docker build` command rather than the target, so even the documentation did not know it existed.

## Explicitly untouched

- **Container-awareness code under `src/`**: `/.dockerenv` probing, cgroup parsing, `ContainerRuntime::Docker`, and Docker-aware disk filtering. all-smi detecting that it is *running inside* a container is a product feature with nothing to do with shipping an image.
- **The container test harness**: the shell scripts under `tests/` and the three `docker-dev` Makefile targets, which run all-smi inside stock `rust:1.88` containers to exercise that feature. None of them built this repository's image.
- **History**: the `README.md` and `debian/changelog` changelog lines and the `TECHNICAL_REPORTS/` entries for #322 and #339 record what happened and remain true.

## Test plan

- [x] `actionlint .github/workflows/ci.yml` reports the same 5 pre-existing findings as before the change, with no new ones (baseline captured before editing)
- [x] PyYAML parse of `ci.yml` succeeds; jobs are now `test`, `packaging-sync`, `systemd-service`, `launchd-service`, `windows-service`, `build-check`; `docker-check` is absent and every remaining `needs:` target resolves to an existing job
- [x] No job referenced `docker-check` even before removal, so the graph needed no repair: the string appeared only on lines 1124 and 1131 inside its own comment header and on line 1166 as the job key
- [x] `main` is not branch-protected (`gh api .../branches/main/protection` returns 404), so no required status check names `Docker Build Check`
- [x] All 19 integration test targets pass, run individually
- [x] 1363 library unit tests pass
- [x] `cargo fmt --check`
- [x] `cargo clippy --bin all-smi -- -D warnings`
- [x] No dangling references to `Dockerfile`, `docker-check`, `docker-build-container`, `docker_build_context_test`, or `docker-compose` remain outside `TECHNICAL_REPORTS/`, `debian/changelog`, and the `README.md` changelog

## Ordering

Depends on #341 only in narrative, not in content: #341 touches `Cargo.toml` and `Cargo.lock`, this PR touches none of those, so the two do not conflict. #341 removed the vendored OpenSSL dependency on its own merit, since it made every release compile OpenSSL from source three times.

Closes #342


inureyes:
## CI status on this PR, and what was verified locally instead

GitHub Actions has been in a **major outage** since 2026-08-06T15:22Z (still `investigating` at the time of writing). Its own status update reads: "Workflow runs are still failing or delayed in starting, and some queued jobs may time out." That is exactly what happened here, so the check rollup on this PR should not be read as a verdict on the change.

**On the final commit `7635eb1`, CI attempt 2 produced:**

| Job | Result |
|---|---|
| Test Suite | pass (2m0s) |
| Packaging Sync Check | pass (9s) |
| launchd Service Smoke Test | pass (1m10s) |
| Build Check | **cancelled** by runner starvation, never ran |
| systemd Service Smoke Test | **cancelled** by runner starvation, never ran |
| Windows Service Smoke Test | skipped (self-hosted runner; also skipped on the fully green run for #343) |

Neither cancelled job ever failed. Subsequent reruns and a close/reopen of this PR produced no new run at all, because Actions has stopped dispatching.

**Build Check was replicated locally on this exact commit**, since it is two plain cargo invocations:

- `cargo build --release --bin all-smi` -> exit 0 (finished in 1m47s)
- `cargo build --no-default-features --lib` -> exit 0

**The systemd smoke test is the one gap.** It cannot run on this macOS host. Two things bound the risk: this diff touches no file it exercises (it reads `packaging/systemd/*` and the `all-smi service` machinery, none of which appears in the seven changed files), and its direct macOS analogue, the launchd smoke test, passed on this exact commit.

**Verification actually performed locally, which is a superset of what CI runs:**

- `actionlint .github/workflows/ci.yml`: same 5 pre-existing findings as the pre-change baseline, no new ones
- PyYAML parse: `docker-check` absent, remaining jobs are `test`, `packaging-sync`, `systemd-service`, `launchd-service`, `windows-service`, `build-check`, and every `needs:` target resolves
- All 19 integration test targets pass, run individually
- 1363 library unit tests pass
- `cargo fmt --check`, `cargo clippy --lib --tests -- -D warnings`, `cargo clippy --bin all-smi -- -D warnings`

Recording this rather than merging on a rollup that the outage made meaningless. Worth a fresh CI run once Actions recovers.

PR #348, opened 2026-08-07
feat: add vendor-neutral Windows GPU metrics via DXGI and PDH

### LINKED ISSUE #346: feat: vendor-neutral Windows GPU metrics via DXGI and PDH (utilization, VRAM, per-process memory)
## Problem

On Windows, both the AMD and Intel GPU readers are WMI-only baselines. `src/device/readers/amd_windows.rs` issues a single query (`SELECT Name, AdapterRAM, DriverVersion, VideoProcessor, PNPDeviceID, Status, AdapterDACType FROM Win32_VideoController`), filters by vendor name substrings (amd/radeon/ati), and fills only `name`, `uuid` (from `PNPDeviceID`), `total_memory`, and a detail map. `utilization`, `temperature`, `used_memory`, `frequency`, and `power_consumption` are hardcoded to 0 (lines 162-170), and `get_process_info()` returns an empty Vec (lines 204-208). `src/device/readers/intel_gpu_windows.rs` explicitly mirrors this shape (same WMI query, same zeroed fields); Intel only partially escapes via the opt-in Level Zero augmentation from #248, which overwrites the WMI zeros for utilization and power when `ze_loader.dll` is present.

On top of the zeroed metrics, total VRAM is wrong for cards with more than 4GB because `Win32_VideoController.AdapterRAM` is a uint32 in WMI, so values cap or wrap at 4GB. `amd_windows.rs` already warns about this (lines 116-130) instead of fixing it.

These gaps are vendor-neutral: they affect both readers identically, and both can be closed with OS facilities that need no vendor SDK.

## Proposed Solution

Add a shared, vendor-neutral Windows GPU metrics layer built on two OS facilities:

**1. DXGI for VRAM.** Enumerate adapters via `IDXGIFactory1::EnumAdapters1` and read `DXGI_ADAPTER_DESC1.DedicatedVideoMemory`, which is 64-bit and fixes the 4GB cap. Query `IDXGIAdapter3::QueryVideoMemoryInfo` with `DXGI_MEMORY_SEGMENT_GROUP_LOCAL` for current usage and budget. Match DXGI adapters to WMI entries by parsing `VEN_xxxx&DEV_xxxx` from `PNPDeviceID` and comparing with `DXGI_ADAPTER_DESC1.VendorId`/`DeviceId`, falling back to name matching. Carry the adapter LUID forward for PDH correlation.

**2. PDH performance counters for utilization and per-process data.** Read `\GPU Engine(*)\Utilization Percentage` and aggregate per adapter LUID: instance names encode `pid_<pid>_luid_<hi>_<lo>_phys_<n>_engtype_<type>`, and summing the 3D and Compute engine types per LUID yields the device utilization figure. Read `\GPU Process Memory(*)\Dedicated Usage` to finally populate `get_process_info()` with per-process dedicated GPU memory. This is the same source Task Manager uses, and the `get_process_info` comment in `intel_gpu_windows.rs` (around line 265) already notes that per-process GPU memory on Windows requires PDH or D3DKMT.

## Integration

- Field-level precedence: vendor-specific source (Level Zero for Intel today, ADL for AMD in the follow-up issue) > PDH/DXGI > WMI baseline zero.
- Update `detail["Metrics Source"]` to reflect the composed sources, following the #248 pattern ("WMI" vs "WMI + Level Zero").
- Remove the 4GB warning path in `amd_windows.rs:116-130` once DXGI totals are wired.
- The layer lives in one shared module (for example `src/device/readers/windows_gpu_perf.rs`) consumed by both vendor readers, not duplicated per vendor.

## Dependencies

Promote `windows` or `windows-sys` (with the `Win32_Graphics_Dxgi` and `Win32_System_Performance` features) to a direct dependency under `[target.'cfg(target_os = "windows")'.dependencies]`. Both crates are already in Cargo.lock transitively (via `wmi` and `windows-service`), so this adds no new supply-chain entries. `libloading = "0.9"` and `wmi = "0.18"` are already direct Windows dependencies.

## Verification Constraints

GitHub-hosted Windows runners expose only the Microsoft Basic Display Adapter and typically no GPU Engine counter instances, so runtime behavior must degrade gracefully to the current WMI baseline on such machines. Unit-test the pure parts everywhere: PDH instance-name LUID/pid parsing, `PNPDeviceID` VEN/DEV parsing, and the precedence logic. Add a `windows.gpu.perf_counters` doctor check (`src/doctor/checks/windows.rs` exists); since there is no AMD Windows hardware in CI, `all-smi doctor` output from real machines is the main field-verification path.

## Acceptance Criteria

- [ ] Utilization reports non-zero on real AMD and Intel Windows machines with a busy GPU
- [ ] Total VRAM is correct for a card with more than 4GB
- [ ] `get_process_info()` returns per-process dedicated GPU memory
- [ ] Graceful no-op on machines without GPU counter instances (falls back to the current WMI baseline)
- [ ] Both `amd_windows.rs` and `intel_gpu_windows.rs` consume the shared layer
- [ ] Doctor check added for GPU performance counter availability
- [ ] Integrated into the actual reader flow and visible in view/api output, not a standalone module

## Related Work

- Related to but strictly separate from #345: that issue is a Linux link-time problem (the glibc target links `libdrm.so.2` unconditionally through `libamdgpu_top`), while this is Windows runtime feature work. Neither blocks the other.
- A follow-up issue, #347, tracks AMD ADL sensor augmentation (temperature, power, fan, clocks); when present, its values take precedence over the PDH/DXGI values from this layer.

inureyes:
## Summary

Closes #346.

The AMD and Intel Windows readers were WMI-only baselines: `utilization`, `used_memory`, `frequency`, and `power_consumption` were hardcoded to `0`, `get_process_info()` returned an empty `Vec`, and `total_memory` was wrong for any card above 4 GB because `Win32_VideoController.AdapterRAM` is a `uint32` in the WMI schema.

This adds `src/device/readers/windows_gpu_perf.rs`, a shared layer built on two OS facilities that need no vendor SDK, consumed by both vendor readers rather than duplicated.

| Source | Supplies |
|---|---|
| DXGI `DXGI_ADAPTER_DESC1` | true 64-bit `DedicatedVideoMemory`, adapter LUID, PCI vendor/device ids |
| PDH `\GPU Engine(*)` | device utilization |
| PDH `\GPU Adapter Memory(*)` | system-wide used VRAM |
| PDH `\GPU Process Memory(*)` | per-process VRAM, which populates `get_process_info()` |

Level Zero keeps precedence over this layer for Intel. Per-field provenance goes in the existing `Source: *` detail keys, and `Metrics Source` composes as backends contribute (`WMI + DXGI + PDH`), following the pattern #248 established.

## Design decisions worth reviewing

**The PDH query is persistent.** `Utilization Percentage` is a rate counter: a single `PdhCollectQueryData` establishes a baseline and yields nothing usable. Rather than sleeping inside the reader to manufacture a second sample, the query is opened once and each poll contributes one collection. The first poll after start-up reports no utilization; every poll after that reports the rate over the real interval. `get_process_info()` reuses that sample via `latest()` instead of collecting again, which would halve the interval the rate is computed over.

**Utilization sums within an engine but maxes across engines.** Each sample is one process's share of one engine, so summing across processes gives the engine's busy fraction. Summing across a card's several 3D and Compute engines yields figures well above 100% and would peg every gauge. The maximum is what Task Manager's headline GPU percentage reports. This is a deliberate deviation from the issue text, which said "summing the 3D and Compute engine types per LUID"; flagging it explicitly.

**DXGI `QueryVideoMemoryInfo` is process-scoped.** MSDN defines `CurrentUsage` and `Budget` as this process's view, not the system's. Reading either as the device's used memory would understate a busy GPU by whatever other processes hold, so both are exposed as clearly labelled diagnostic detail fields (`VRAM Usage (this process)`) and `used_memory` comes from the PDH adapter counter instead.

**Video engines are excluded from utilization.** A compositor decoding video keeps `VideoDecode` busy while the shader cores idle; folding it in would make an idle desktop report a large non-zero load.

**Counters are added with `PdhAddEnglishCounterW`.** Counter path components are localized on non-English Windows, so the literal English path only resolves through the English-specific entry point.

## Verification

**No CI job in this repository compiles all-smi for Windows.** The only Windows job (`windows-service` in `ci.yml`) is gated behind `vars.ENABLE_WINDOWS_SERVICE_SMOKE`, which is unset, and its own comment states it has never executed. Windows-only code therefore ships with zero automated coverage by default. Two things were done about that:

1. The module is gated `cfg(any(target_os = "windows", test))`, matching the existing `intel_gpu_sysfs` pattern, so the counter-instance parsing, the utilization aggregation, the PNPDeviceID matching, and the field application are all exercised by the Linux test runner. 28 new tests.
2. The FFI itself was verified by cross-compiling to the real release target from macOS.

| Gate | Result |
|---|---|
| `cargo fmt --check` | pass |
| `cargo clippy --all-targets -- -D warnings` | pass |
| `cargo test` | 3118 pass, 0 fail |
| `cargo xwin check --target x86_64-pc-windows-msvc` | pass |
| `cargo xwin clippy --target x86_64-pc-windows-msvc -- -D warnings` | pass for this change |

The Windows cross-check needed `cargo-xwin` plus `llvm-lib` (rustup's `llvm-tools` provides `llvm-ar`, which acts as `llvm-lib` under that name); `zstd-sys` otherwise blocks any Windows cross-compile from macOS.

## Pre-existing findings, not fixed here

Windows clippy reports three lints in files this PR does not touch, surfaced only because nothing had ever linted that target:

- `src/device/cpu_windows.rs:49` and `:63`, collapsible `if`
- `src/device/windows_temp/amd_ryzen.rs:228`, collapsible `if`

Left alone to keep this diff scoped. Worth a follow-up together with a CI job that actually compiles the Windows target, which `windows-latest` runners would do for free on this public repository.

## Not verified

Acceptance criteria that need real hardware: non-zero utilization on a busy AMD and Intel Windows machine, correct VRAM on a card above 4 GB, and populated per-process rows. The `windows.gpu.perf_counters` doctor check exists to produce exactly that evidence from an operator's machine; it distinguishes "no DXGI adapters", "DXGI works but PDH publishes no instances" (normal on VMs and RDP), and the full path.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -104,6 +104,7 @@ dependencies = [
  "tracing-subscriber",
  "url",
  "whoami",
+ "windows",
  "windows-service",
  "wmi",
  "zstd",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -83,6 +83,18 @@ libamdgpu_top = "=0.11.5"
 [target.'cfg(target_os = "windows")'.dependencies]
 wmi = "0.18"
 libloading = "0.9"
+# DXGI (true 64-bit VRAM size, adapter LUID / PCI ids) and PDH (GPU
+# Engine utilization, per-adapter and per-process dedicated memory) for
+# the vendor-neutral Windows GPU metrics layer, issue #346. Version
+# pinned to the one `wmi` already resolves, so this promotes an existing
+# transitive dependency to a direct one rather than adding a new entry
+# to the supply chain.
+windows = { version = "0.62", features = [
+    "Win32_Foundation",
+    "Win32_Graphics_Dxgi",
+    "Win32_Graphics_Dxgi_Common",
+    "Win32_System_Performance",
+] }
 # Service Control Manager integration for `all-smi service` (issue #311).
 # Mullvad's crate, MIT OR Apache-2.0, MSRV 1.71 against this project's
 # rust-version = 1.96. Chosen over shelling to sc.exe so failures arrive
--- a/src/device/readers/amd_windows.rs
+++ b/src/device/readers/amd_windows.rs
@@ -239,14 +239,17 @@ impl GpuReader for AmdWindowsGpuReader {
 
     fn get_process_info(&self) -> Vec<ProcessInfo> {
         // Per-process GPU memory comes from the PDH `GPU Process
-        // Memory` counter. `latest()` reuses the sample `get_gpu_info`
-        // already took: collecting again here would consume a second
-        // PDH collection and halve the interval the utilization rate is
-        // computed over.
+        // Memory` counter, reusing the sample `get_gpu_info` already
+        // took. The closure covers one-shot callers that never call
+        // `get_gpu_info` at all, such as
+        // `all-smi snapshot --include process`.
         let Ok(adapter_index) = self.adapter_index.lock() else {
             return Vec::new();
         };
-        windows_gpu_perf::process_rows(&adapter_index)
+        windows_gpu_perf::process_rows_with(&adapter_index, || {
+            let mut gpus = self.query_amd_gpus();
+            windows_gpu_perf::augment_gpus(&mut gpus)
+        })
     }
 }
 
--- a/src/device/readers/amd_windows.rs
+++ b/src/device/readers/amd_windows.rs
@@ -12,18 +12,26 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//! AMD GPU reader for Windows using WMI
+//! AMD GPU reader for Windows.
 //!
-//! This module provides basic AMD GPU information on Windows via WMI.
-//! Note: Detailed metrics like utilization and temperature require AMD ADL SDK,
-//! which is not currently implemented.
+//! `Win32_VideoController` supplies the card's identity (name, PCI ids,
+//! driver version) and nothing else of substance: no utilization, no
+//! temperature, and a `uint32` `AdapterRAM` field that saturates at
+//! 4 GB. The vendor-neutral [`windows_gpu_perf`] layer is applied on top
+//! to fill in the true VRAM size (DXGI), device utilization and
+//! system-wide used VRAM (PDH), and per-process GPU memory.
+//!
+//! Temperature, power, fan speed, and clocks are still absent here:
+//! WDDM does not publish them, and they need AMD's own ADL library.
 
 use crate::device::GpuReader;
+use crate::device::readers::windows_gpu_perf::{self, ids::AdapterLuid};
 use crate::device::types::{GpuInfo, ProcessInfo};
 use crate::utils::get_hostname;
 use chrono::Local;
 use serde::Deserialize;
 use std::collections::HashMap;
+use std::sync::Mutex;
 use wmi::WMIConnection;
 
 // Thread-local WMI connection for reuse within the same thread
@@ -69,7 +77,17 @@ struct VideoControllerName {
     name: Option<String>,
 }
 
-pub struct AmdWindowsGpuReader {}
+pub struct AmdWindowsGpuReader {
+    /// Adapter LUID to `(device index, GPU uuid)`, recorded by
+    /// `get_gpu_info` so `get_process_info` can attribute a PDH
+    /// per-process row to the right card without re-querying WMI or
+    /// consuming a second PDH collection.
+    ///
+    /// Empty until the first `get_gpu_info` call, so a
+    /// `get_process_info` that runs first simply reports nothing and the
+    /// next poll fills it in.
+    adapter_index: Mutex<HashMap<AdapterLuid, (usize, String)>>,
+}
 
 impl Default for AmdWindowsGpuReader {
     fn default() -> Self {
@@ -79,7 +97,9 @@ impl Default for AmdWindowsGpuReader {
 
 impl AmdWindowsGpuReader {
     pub fn new() -> Self {
-        Self {}
+        Self {
+            adapter_index: Mutex::new(HashMap::new()),
+        }
     }
 
     fn query_amd_gpus(&self) -> Vec<GpuInfo> {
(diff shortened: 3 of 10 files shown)


inureyes:
## Review round applied

All two HIGH and five MEDIUM findings are fixed in `000f7b9`. Summary of what changed and why, since several were behavioural rather than mechanical:

**H1, multiple readers consuming separate PDH collections.** Confirmed real: `reader_factory.rs` registers AMD and Intel with independent `if`s, so a hybrid machine has both readers, and each `get_gpu_info` consumed a collection. The second reader always measured a microsecond interval. `snapshot()` now reuses a cached sample younger than 500 ms (below the 1 s fastest poll, above the time to walk the readers). Fixes the `get_gpu_info_by_uuid` loop case and the cache-clobbering second-order issue at the same time.

**H2, `PDH_CSTATUS_NEW_DATA`.** Fixed. Both 0 and 1 are documented success statuses and NEW_DATA is precisely the loaded-GPU case, so this would have silently degraded the whole feature while the doctor check reported healthy.

**M1** bitflag now masked. **M5** `match_adapter` requires vendor agreement for every fallback below exact PCI match and declines to guess without a parseable vendor, since a wrong attribution is worse than the honestly-empty baseline it replaced; empty descriptions no longer match everything. **M6** integrated GPUs now fall back to `SharedSystemMemory` with `DXGI (shared)` / `PDH (dedicated carve-out only)` provenance so the source string does not overclaim. **M2** `get_process_info` now merges against the system process table like the nvidia, jetson, and tenstorrent readers, which fixes the empty names and zeroed CPU/RSS on the api and snapshot paths. **M3** an empty adapter index consults a refresh closure so `snapshot --include process` works. **L1/L2** retry once on `PDH_MORE_DATA`, null-check `szName`.

**M4** is inherent to rate counters and is now documented on the module rather than fixed: one-shot callers see baseline utilization, polling paths are correct from the second poll.

Verified after the fixes:

| Gate | Result |
|---|---|
| `cargo fmt --check` | pass |
| `cargo clippy --all-targets -- -D warnings` | pass |
| `cargo test` | 3128 pass, 0 fail (31 in this module, up from 26) |
| `cargo xwin check --target x86_64-pc-windows-msvc` | pass, 0 warnings |
| `cargo xwin clippy --target x86_64-pc-windows-msvc -- -D warnings` | pass for this change |

The three pre-existing Windows clippy lints (`cpu_windows.rs:49`, `:63`, `amd_ryzen.rs:228`) are still untouched and still worth a follow-up alongside a CI job that actually compiles the Windows target.

Issue #345, opened 2026-08-06
fix: AMD backend links libdrm unconditionally, so binaries fail to start without AMD userspace libraries


inureyes:
## Problem

`all-smi` declares `libamdgpu_top` as a **non-optional, target-gated** dependency:

```toml
[target.'cfg(all(target_os = "linux", not(target_env = "musl")))'.dependencies.libamdgpu_top]
version = "=0.11.5"
```

(verified in the published `all-smi` 0.25.0 `Cargo.toml`)

No cargo feature gates it. `all-smi` 0.25.0's full feature set is `cli` (default), `furiosa`, `level_zero`, and `mock`, so there is no way for a downstream crate to opt out. Declaring `all-smi = { version = "0.25.0", default-features = false }` (which `backend.ai-go` already does) does not remove it.

Dependency chain, resolved with `cargo tree -i libdrm_amdgpu_sys --target x86_64-unknown-linux-gnu`:

```
libdrm_amdgpu_sys v0.8.16
└── libamdgpu_top v0.11.5
    └── all-smi v0.25.0
        └── backend-ai-go
```

`libdrm_amdgpu_sys` 0.8.16 links `libdrm.so.2` and `libdrm_amdgpu.so.1` dynamically and unconditionally.

## Impact

Any Linux binary that depends on `all-smi` inherits `libdrm.so.2` and `libdrm_amdgpu.so.1` as hard `NEEDED` entries. A host with **no AMD GPU at all**, which is the overwhelming majority of deployments, cannot start the binary unless AMD's userspace DRM libraries happen to be installed. The failure is a loader error before `main` runs, so the program cannot detect the situation and degrade gracefully, and cannot print anything useful.

Measured with `objdump -p` against the published `backend.ai-go` v1.11.1 server binaries:

- `aigo-server-linux-x86_64`: `libdrm.so.2`, `libdrm_amdgpu.so.1`, `libssl.so.3`, `libcrypto.so.3`, `libgcc_s.so.1`, `libm.so.6`, `libc.so.6`
- `aigo-server-linux-aarch64`: `libdrm.so.2`, `libdrm_amdgpu.so.1`, `libgcc_s.so.1`, `libm.so.6`, `libc.so.6`

The OpenSSL entries have a separate cause that `backend.ai-go` is fixing on its own side. The two `libdrm` entries come from this crate and cannot be fixed downstream.

[...]

PR #349, opened 2026-08-07
feat: add AMD ADL sensor augmentation on Windows via PMLog

### LINKED ISSUE #347: feat: AMD ADL sensor augmentation on Windows (temperature, power, fan, clocks via PMLog)
## Problem

Even with the vendor-neutral DXGI/PDH layer proposed in #346, temperature, board power, fan RPM, and core/memory clocks on Windows AMD GPUs are only reachable through AMD's vendor API. `src/device/readers/amd_windows.rs` currently hardcodes `temperature`, `power_consumption`, and `frequency` to 0 (lines 162-170) and points at the vendor SDK via `detail["Note"] = "Detailed metrics require AMD ADL SDK"`.

## Proposed Solution

Add an opt-in ADL augmentation that runtime-loads the 64-bit `atiadlxx.dll` via `libloading` from the absolute path `C:\Windows\System32\atiadlxx.dll` only, with no search-path loading, matching the DLL-hijacking stance documented in `src/device/windows_temp/amd_ryzen.rs`. No import library is linked, so there is no load-time NEEDED entry: absence of the DLL degrades to the WMI/PDH baseline. This is deliberately the same shape as the resolution direction of #345, where absence of a vendor library must be a runtime observation rather than a startup failure, but this issue is Windows-only and independent of #345. The repo already has runtime-dynamic-loading precedents to follow: `src/device/windows_temp/amd_ryzen.rs`, `src/device/readers/intel_gpu_level_zero/loader.rs`, and `src/device/readers/tpu_pjrt.rs`. `libloading = "0.9"` is already a direct Windows dependency in Cargo.toml.

## Scope Discipline

Implement only the modern path:

- Use the ADL2 context APIs (`ADL2_Main_Control_Create` with an `ADL_Main_Memory_Alloc` callback, `ADL2_Main_Control_Destroy`), never the global-state `ADL_*` variants, because all-smi polls from background tasks.
- Read sensors exclusively via `ADL2_New_QueryPMLogData_Get`, which on modern (Vega/RDNA and newer) cards returns edge/hotspot/memory temperature, GFX and memory activity, board power, fan RPM, GFX/MEM clocks, and VRAM usage in a single call. Gate with `ADL2_Overdrive_Caps`; if PMLog is unsupported (pre-Vega), leave the fields at the baseline values instead of implementing legacy paths.
- Explicitly out of scope: Overdrive 5/6/7 fallbacks, ADLX, and per-process data (covered by PDH in #346).

Sensor IDs to map: `PMLOG_TEMPERATURE_EDGE`, `PMLOG_TEMPERATURE_HOTSPOT`, `PMLOG_TEMPERATURE_MEM`, `PMLOG_ASIC_POWER` (or `PMLOG_GFX_POWER`), `PMLOG_FAN_RPM`, `PMLOG_CLK_GFXCLK`, `PMLOG_CLK_MEMCLK`, `PMLOG_INFO_ACTIVITY_GFX`, `PMLOG_INFO_ACTIVITY_MEM`.

## FFI Safety

Re-declare only the needed function signatures and structs in Rust, with no vendored AMD headers, keeping the project Apache-2.0 clean. Use `#[repr(C)]` structs with compile-time `size_of` assertions exactly as `src/device/windows_temp/amd_ryzen.rs` does (its 56-byte `RmQuickStats` assertion is the template). `ADLPMLogDataOutput` is a fixed array of `{supported, value}` i32 pairs indexed by sensor ID, so a layout mistake corrupts memory and the assertion is mandatory. Cache the loaded library and the ADL2 context in a OnceCell/RwLock, following `amd_ryzen.rs`.

## Integration

- ADL values overwrite the WMI zeros and take precedence over PDH utilization when both are present; this precedence logic interacts with the shared layer from #346.
- Set `detail["Metrics Source"]` to reflect the composed sources, mirroring the Intel "WMI + Level Zero" pattern from #248.
- Wire into `AmdWindowsGpuReader` behind a runtime check, not a Cargo feature. Unlike the `level_zero` feature gate, nothing is linked at build time, so there is no link-time cost to gate away; the always-compiled `amd_ryzen.rs` model fits better than the `intel_gpu_level_zero` feature-gate model. This choice is stated here explicitly so it can be debated in review.

## Diagnostics and Verification

No AMD Windows hardware exists in CI, and cross-compiled builds never execute this code, so field diagnosis via `all-smi doctor` is the main verification path. Add `amd.adl.*` doctor checks in `src/doctor/checks/amd.rs` (currently ROCm/libamdgpu_top/DRI focused and skipped on non-Linux): DLL present at the expected path, context creation succeeds, PMLog capability reported, one sample read successfully. Field verification via `all-smi doctor` output from a real AMD Windows machine is a required acceptance step. Unit tests cover the struct-size assertions and the sensor-ID mapping table.

## Acceptance Criteria

- [ ] Temperature, power, fan RPM, and clocks populated on a PMLog-capable AMD card on Windows
- [ ] Graceful baseline behavior when the DLL is absent, when context creation fails, and when PMLog is unsupported
- [ ] No new load-time imports: `dumpbin /imports` (or equivalent) on the built exe shows no `atiadlxx` entry
- [ ] `amd.adl.*` doctor checks added in `src/doctor/checks/amd.rs`
- [ ] `detail["Metrics Source"]` reflects the composed sources
- [ ] Integrated into the actual reader flow and view/api output, not a standalone module

## Related Work

- Depends functionally on nothing, but the field-precedence logic interacts with #346 (vendor-neutral DXGI/PDH layer); neither blocks the other.
- Related to but strictly separate from #345: #345 is a Linux link-time problem, this is Windows runtime feature work, and neither should block the other.

inureyes:
## Summary

Closes #347. Builds on #346.

The DXGI and PDH layer from #346 covers everything WDDM publishes. It cannot cover temperature, board power, fan speed, or clocks, because Windows does not expose them at all. This adds `src/device/readers/amd_adl.rs`, which reads them from AMD's own library.

`atiadlxx.dll` is loaded at runtime via `libloading` from the absolute path `C:\Windows\System32\atiadlxx.dll`, never by bare name, matching the DLL-hijacking stance already documented in `windows_temp/amd_ryzen.rs`. **No import library is referenced**, so the executable carries no `atiadlxx` entry in its import table and a machine without AMD's driver starts normally and simply reports no ADL data. That is deliberately the same shape #345 asks for on Linux, where a missing `libdrm` is a loader error before `main`.

Precedence is WMI < DXGI/PDH < ADL. ADL reads the hardware's own telemetry rather than the OS's accounting of it, so it overwrites PDH utilization and is the sole source for temperature, power, fan, and clocks.

## Two deliberate scope limits, both about refusing to declare an ABI blind

**Legacy Overdrive paths are not implemented.** Sensors come only from `ADL2_New_QueryPMLogData_Get`, gated on `ADL2_Overdrive_Caps`. OD5/6/7 would be three more ABI surfaces to get right with no way to test them, for hardware predating the cards all-smi targets. A pre-Vega card keeps the baseline.

**`AdapterInfo` is not declared, so augmentation requires exactly one AMD GPU.** This is the important one. `AdapterInfo` is the 1568-byte struct that carries the PCI bus/device/function and PNP string that would let an ADL adapter index be tied to a specific card. ADL sizes its write by *its own* `sizeof`, so a layout mistake overflows our buffer rather than failing cleanly. Worse, a single card exposes several adapter indices (one per display output) reporting identical telemetry, which cannot be deduplicated without that struct either.

Rather than guess, `can_attribute()` requires a single AMD GPU. A multi-AMD-GPU host gets the honest DXGI/PDH baseline instead of one card's temperature reported against another. This is the same conclusion the #346 review reached about adapter matching: declining to attribute beats attributing wrongly. Multi-GPU support is a follow-up that would need `AdapterInfo` and real hardware to validate against.

## The sensor indices are the weakest point, and are treated as such

`ADLSensorType` indices are transcribed from AMD's public `adl_structures.h`. Nothing in CI compiles all-smi for Windows and no test can call the real library, so if AMD renumbered an entry this would read the wrong sensor and report a *plausible but wrong* number, which is worse than reporting nothing. Two mitigations:

1. **Range guards.** Every value is checked against a physically sensible band (temperature 0-150 C, power 0-1000 W, activity 0-100%, and so on). A misindexed read almost always lands outside its target band, so the guard turns a silent wrong number into an absent one. Tested with a case that feeds a clock value into the temperature slot and asserts the whole readout comes back empty.
2. **`amd.adl.sensors` doctor check dumps the raw `index=value` table**, unfiltered and unnamed. That makes the mapping confirmable from real hardware without a code change shipping first, and if every sensor fails the range guard the check says explicitly that this is what a shifted enum looks like and asks for the dump.

`amd.adl.library` separately distinguishes: DLL absent, DLL present but context creation failed, loaded but no PMLog-capable adapter, and healthy with the selected adapter index.

## Also in this PR: DXGI factory caching

Prompted by a question about polling overhead. `CreateDXGIFactory1` is COM object creation that can pull in graphics driver DLLs, and #346 ran it on every poll (as often as once per second). It is now created once and rebuilt only when `IDXGIFactory1::IsCurrent` reports the adapter set changed, which also keeps hot-plug correct.

For the record on cost, since it came up: **nothing in either layer submits GPU work.** DXGI reads adapter descriptors, PDH reads counters the OS already maintains, ADL reads a telemetry block the driver already collects. The load is CPU-side and small; the library load, ADL context creation, and the PMLog capability scan each happen once per process, so a steady-state poll is a single ADL call. One caveat now documented in the module: very aggressive sensor polling (the 100 ms rates desktop monitoring tools use) can hold an AMD GPU out of its deepest idle state. all-smi's intervals start at one second, well clear of that.

## Verification

Same constraint as #346: **no CI job compiles all-smi for Windows.** The module is gated `cfg(any(target_os = "windows", test))` so the sensor mapping, the range guards, and the field application all run on the Linux test runner. 34 new tests.

| Gate | Result |
|---|---|
| `cargo fmt --check` | pass |
| `cargo clippy --all-targets -- -D warnings` | pass |
| `cargo test` | 3162 pass, 0 fail (was 3128 on main) |
| `cargo xwin check --target x86_64-pc-windows-msvc` | pass, 0 warnings |
| `cargo xwin clippy --target x86_64-pc-windows-msvc -- -D warnings` | pass for this change |

Layout is pinned by compile-time assertions on `ADLSingleSensorData` (8 bytes, 4-aligned) and `ADLPMLogDataOutput` (2052 bytes), plus a test asserting the sensor array actually strides by one 8-byte record from offset 4. Those assertions are the only automated check this ABI can have, which is why they are there rather than being decorative.

## Not verified, needs real hardware

- Temperature, power, fan, and clocks populated on a PMLog-capable card
- The sensor index mapping itself (`amd.adl.sensors` exists to produce this evidence)
- `dumpbin /imports` showing no `atiadlxx` entry. The design guarantees it (nothing is linked, `libloading` only), but it has not been observed on a built exe.

## Pre-existing, still not fixed

The same three Windows clippy lints noted in #348, in files neither PR touches: `cpu_windows.rs:49`, `:63`, `amd_ryzen.rs:228`. Worth a follow-up together with a CI job that compiles the Windows target, which `windows-latest` runners would do for free on this public repo.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -94,6 +94,7 @@ windows = { version = "0.62", features = [
     "Win32_Graphics_Dxgi",
     "Win32_Graphics_Dxgi_Common",
     "Win32_System_Performance",
+    "Win32_System_Memory",
 ] }
 # Service Control Manager integration for `all-smi service` (issue #311).
 # Mullvad's crate, MIT OR Apache-2.0, MSRV 1.71 against this project's
--- a/src/device/readers/amd_adl.rs
+++ b/src/device/readers/amd_adl.rs
@@ -1,0 +1,349 @@
+// Copyright 2025 Lablup Inc. and Jeongkyu Shin
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! AMD ADL sensor augmentation for the Windows GPU reader.
+//!
+//! The vendor-neutral DXGI and PDH layer added in #346 covers everything
+//! WDDM publishes: VRAM size, utilization, per-process memory. It cannot
+//! cover temperature, board power, fan speed, or clocks, because Windows
+//! does not expose those at all. Those come from AMD's own library.
+//!
+//! `atiadlxx.dll` is loaded at runtime from an absolute System32 path,
+//! so nothing is linked and a machine without AMD's driver simply gets
+//! no ADL data. See [`loader`] for the loading and hijacking stance.
+//!
+//! ## Scope: modern parts only
+//!
+//! Sensors are read exclusively through `ADL2_New_QueryPMLogData_Get`,
+//! the Overdrive 8 style sensor table that Vega and later expose. The
+//! legacy Overdrive 5, 6, and 7 temperature entry points are
+//! deliberately not implemented: they are three more ABI surfaces to get
+//! right blind, for hardware that predates the datacentre and workstation
+//! cards all-smi targets. A pre-Vega card keeps the WMI and PDH baseline.
+//!
+//! ## Scope: one AMD GPU
+//!
+//! ADL identifies adapters by an index that does not map to a card
+//! without `AdapterInfo`, a 1568-byte struct whose layout this module
+//! deliberately refuses to declare blind (see [`ffi`]). Worse, one card
+//! exposes several adapter indices, one per display output, all
+//! reporting identical telemetry, so an index cannot even be
+//! deduplicated without it.
+//!
+//! Rather than guess, augmentation only runs when the reader found
+//! exactly one AMD GPU. That covers the overwhelming majority of real
+//! machines, and on a multi-AMD-GPU host the result is the honest DXGI
+//! and PDH baseline instead of one card's temperature reported against
+//! another. This mirrors the conclusion the #346 review reached about
+//! adapter matching: declining to attribute beats attributing wrongly.
+//!
+//! ## Cost
+//!
+//! Nothing here submits GPU work; PMLog reads a telemetry block the
+//! driver already maintains. The library load, context creation, and
+//! capability scan each happen once for the process, so a steady-state
+//! poll is a single call. Note that very aggressive sensor polling (at
+//! the 100 ms rates desktop monitoring tools use) can hold an AMD GPU
+//! out of its deepest idle power state; all-smi's intervals start at one
+//! second, which is well clear of that.
+
+pub mod ffi;
+pub mod sensors;
+
+#[cfg(target_os = "windows")]
+pub mod loader;
+
+use crate::device::readers::windows_gpu_perf::note_metrics_source;
+use crate::device::types::GpuInfo;
+use sensors::AdlReadout;
+
+/// Whether an ADL readout can be attributed to a specific card.
+///
+/// See the module docs: without `AdapterInfo` an ADL adapter index
+/// cannot be tied to a GPU, so attribution is only safe when there is
+/// nothing to confuse it with.
+pub fn can_attribute(amd_gpu_count: usize) -> bool {
+    amd_gpu_count == 1
+}
+
+/// Layer an ADL readout onto a GPU that already carries the WMI, DXGI,
+/// and PDH baseline.
+///
+/// ADL outranks the vendor-neutral layer: it reads the hardware's own
+/// telemetry rather than the OS's accounting of it. Utilization is
+/// therefore overwritten when ADL reports graphics activity, and
+/// temperature, power, fan, and clocks are filled in for the first time.
+///
+/// Fields ADL did not produce are left exactly as they were, so a card
+/// that publishes only some sensors keeps the baseline for the rest.
+pub fn apply_to_gpu_info(gpu: &mut GpuInfo, readout: &AdlReadout) {
+    if readout.is_empty() {
+        return;
+    }
+
+    if let Some(temperature) = readout.primary_temperature_c() {
+        gpu.temperature = temperature;
+        gpu.detail
+            .insert("Source: Temperature".to_string(), "ADL".to_string());
+    }
+    // Hotspot and memory temperatures have no dedicated `GpuInfo` field
+    // but are the numbers that actually throttle a modern card, so they
+    // are surfaced as details rather than dropped.
+    if let Some(hotspot) = readout.temperature_hotspot_c {
+        gpu.detail
+            .insert("Hotspot Temperature (C)".to_string(), hotspot.to_string());
+    }
+    if let Some(memory) = readout.temperature_mem_c {
+        gpu.detail
+            .insert("Memory Temperature (C)".to_string(), memory.to_string());
+    }
+
+    if let Some(power) = readout.power_w {
+        gpu.power_consumption = power;
+        gpu.detail
+            .insert("Source: Power".to_string(), "ADL".to_string());
+    }
+
+    if let Some(clock) = readout.clock_gfx_mhz {
+        gpu.frequency = clock;
+        gpu.detail
+            .insert("Source: Frequency".to_string(), "ADL".to_string());
+    }
+    if let Some(clock) = readout.clock_mem_mhz {
+        gpu.detail
+            .insert("Memory Clock (MHz)".to_string(), clock.to_string());
+    }
+
+    if let Some(rpm) = readout.fan_rpm {
+        // `GpuInfo` has no fan field; the Intel reader already
+        // advertises fan provenance the same way.
+        gpu.detail
+            .insert("Fan Speed (RPM)".to_string(), rpm.to_string());
+        gpu.detail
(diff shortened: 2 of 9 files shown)


inureyes:
## Review round applied

All findings addressed in `d7e1afb`. The reviewer cross-checked the ABI against LibreHardwareMonitor's transcription of AMD's headers, which is the closest thing to an independent reference this API has, and that turned up two things I could not have found from the headers alone.

**H1, `PMLOG_ASIC_POWER` is the wrong sensor on RDNA3+.** Real, and it would have produced a wrong number on the exact card class this targets. `PMLOG_BOARD_POWER` (73) is now declared and preferred, falling back to ASIC then GFX. Worth noting explicitly: **no range check could ever have caught this**, since all three are watts and all three are plausible. That makes the preference order load-bearing rather than cosmetic.

**H2, the range guard's docs overclaimed.** Correct criticism. The guard catches unit and scale errors well; it does not catch a small index shift, which is the likeliest failure mode because the enum is grouped by quantity (reading `FAN_PERCENTAGE` as `FAN_RPM` yields 0-100 and passes `0..=20_000`). The docs now say precisely that and name the doctor dump as the mitigation the design actually depends on. No behaviour change, but the honesty of the claim matters given it was the stated justification for shipping an unverifiable mapping.

**M3** the scan no longer latches: a service starting before the display driver came up would previously have disabled ADL for the whole process lifetime and blamed a pre-Vega card. Now retries on a slow interval, and a stale cached index is dropped after a driver update or TDR reset. **M4** `ADL2_Overdrive_Caps` is advisory, not a gate, since workstation SKUs and APUs report `iSupported = 0` while still serving the table. **M5** the PMLog buffer is boxed with headroom and its size validated, applying the same reasoning that made me refuse `AdapterInfo`. **M6** the `SendFactory` justification was factually wrong and is corrected. **M7** `TEMPERATURE_GFX` (28) added and preferred over hotspot, and the chosen sensor is now named in `Source: Temperature`. **M8** the `Note` lists only fields that were actually applied.

**L9** entry points are `extern "C"` (cdecl), only the callback is `extern "system"`. **L10** `adl_malloc` uses `HeapAlloc(GetProcessHeap(), ...)` so a future custom global allocator cannot turn a bounded leak into heap corruption. **L11** `LOAD_LIBRARY_SEARCH_SYSTEM32`, which closes the dependency-chain hole the module claimed to close but did not (atiadlxx pulls in atiadlxy). **L12** temperature is signed and accepts down to -40 C.

**L13 not fixed:** fan RPM and the extra temperatures land in `detail`, which reaches `snapshot` JSON but not the TUI or Prometheus output. `GpuInfo` has no fan field. Temperature, power, and gfx clock do land on real fields and reach both surfaces, so the acceptance criterion is partly met; wiring fan through properly means touching `GpuInfo` or the metrics exporter and is better as its own change.

Verified after the fixes:

| Gate | Result |
|---|---|
| `cargo fmt --check` | pass |
| `cargo clippy --all-targets -- -D warnings` | pass |
| `cargo test` | 3172 pass, 0 fail (22 in amd_adl, up from 17) |
| `cargo xwin check --target x86_64-pc-windows-msvc` | pass, 0 warnings |
| `cargo xwin clippy --target x86_64-pc-windows-msvc -- -D warnings` | pass for this change |

**Flaky test note.** `tests/sse_events_test.rs` has two load-sensitive tests unrelated to this work. `fifty_concurrent_clients_do_not_stall_the_publisher` asserts `FrameBus::publish` averages under 1 ms; on my machine it measures 1.0-1.6 ms and fails intermittently **on `main` as well** (2/5 runs on `main` at 55c6a1a, 3/5 on this branch, same ballpark). `events_lag_event_emitted_for_slow_receiver` failed once on #348's CI and passed on rerun. Both are timing assertions in the file that already needed a flake fix in #338. Worth a follow-up issue; neither is caused by these PRs.

PR #360, opened 2026-08-07
feat: promote GPU fan speed to a first-class GpuInfo field

### LINKED ISSUE #352: feat: promote GPU fan speed to a first-class GpuInfo field
## Problem / Background

Fan speed is published only through the untyped `detail: HashMap<String, String>` map. Four readers write it, all now using the key `Fan Speed` with a value like `"1450 RPM"` after PR #351:

- `src/device/readers/amd.rs:418` (Linux AMD)
- `src/device/readers/intel_gpu_linux/sources.rs:66`
- `src/device/readers/intel_gpu_level_zero/apply.rs:127`
- `src/device/readers/amd_adl.rs` (Windows AMD, from #349)

Being a detail string rather than a field has four consequences:

1. It reaches `all-smi snapshot` JSON/CSV (because `detail` is serde-serialized) but **not** the TUI and **not** the Prometheus exporter. `src/api/metrics/gpu.rs` reads a fixed key list and does not render arbitrary `detail` keys.
2. The provenance already escapes but the value does not. `Source: Fan` is exported as the label `source__fan` (see `src/parsing/common.rs:192` `sanitize_label_name`, and the assertion in `src/api/metrics/gpu.rs:557`). A Prometheus consumer can see *where* fan data would come from while never seeing the number.
3. Consumers must parse `"1450 RPM"` back out of a string, and every reader must agree on spelling by convention. PR #351 had to add a test to hold that line precisely because one reader had drifted.
4. Issue #347's acceptance criterion "temperature, power, fan RPM, and clocks populated ... integrated into view/api output" is therefore only partly met. Temperature, power, and gfx clock land on real `GpuInfo` fields and reach both surfaces; fan does not.

## Proposed Solution

Add `fan_speed_rpm: Option<u32>` to `GpuInfo` in `src/device/types.rs` with `#[serde(default)]` (the struct already uses that attribute in at least 8 places for wire-compatibility with older snapshots and remote nodes). Then:

- Populate it from all four readers, keeping the `detail` entry for now or removing it, whichever review prefers.
- Export `all_smi_gpu_fan_speed_rpm` from `src/api/metrics/gpu.rs`, gated on presence the way other optional metrics are.
- Add the reverse parse in `src/network/metrics_parser.rs` so remote monitoring round-trips it, matching how other GPU metrics are reconstructed.
- Render it in the TUI GPU view where space allows (`src/ui/renderers/gpu_renderer.rs`).

## Cost Estimate

`GpuInfo` has roughly 78 construction sites (measured by counting references to the existing `gpu_core_count:` field across `src/`), so a new non-defaulted field means about that many mechanical edits. All are compiler-checked. There is no FFI and no ABI risk; this is plumbing.

## Technical Considerations

`ChassisInfo.fan_speeds: Vec<FanInfo>` and the `all_smi_chassis_fan_speed_rpm` metric already exist (`src/api/metrics/chassis.rs:311`). Decide during design whether GPU fan speed belongs on `GpuInfo` or should feed the existing chassis fan model. The former is probably right since it is per-device, but the overlap should be explicit rather than accidental.

## Acceptance Criteria

- [x] `fan_speed_rpm: Option<u32>` added to `GpuInfo` with `#[serde(default)]`
- [x] Populated by all four readers (Linux AMD, Intel GPU Linux, Intel Level Zero, Windows AMD ADL)
- [x] Exported to Prometheus as `all_smi_gpu_fan_speed_rpm`, gated on presence
- [x] Parsed back in `src/network/metrics_parser.rs` so remote monitoring round-trips it
- [x] Visible in the TUI GPU view
- [x] Existing `detail`-based behavior does not regress
- [x] Remote monitoring of a mixed old/new fleet still deserializes

inureyes:
## Summary

Fan speed was published only through the untyped `detail` map, so it reached `all-smi snapshot` output but never the TUI or the Prometheus exporter, and consumers had to parse `"1450 RPM"` back out of a string that four readers agreed on by convention alone. This promotes it to `GpuInfo::fan_speed_rpm` and wires it through the exporter, the remote-metrics parser, and the GPU view.

The typed field is `Option<u32>` with `#[serde(default)]`, so snapshots and remote payloads produced before it existed still deserialize. `None` means the device reports no tachometer (passively cooled datacenter cards, drivers that expose only a duty cycle), and every consumer renders nothing rather than substituting `0`, which would be indistinguishable from a seized fan.

## What changed

- `src/device/types.rs`: added `fan_speed_rpm: Option<u32>` to `GpuInfo` with `#[serde(default)]`, documenting that `None` means "no tachometer" and must never be rendered as `0`. The other 48 construction sites take `None`.
- Readers now set the typed field from the same source value, in the same branch, as the existing `Fan Speed` detail write: `src/device/readers/amd.rs` (Linux amdgpu sensors), `src/device/readers/intel_gpu_linux.rs` (hwmon `fan1_input`), `src/device/readers/intel_gpu_level_zero/apply.rs` (Sysman), and `src/device/readers/amd_adl.rs` (Windows PMLog).
- The detail writes are deliberately kept. Snapshots depend on them, and `intel_gpu_level_zero::apply_fan` uses `detail.contains_key("Fan Speed")` as the cross-reader overwrite guard that lets a Linux hwmon reading outrank a later Level Zero sample. Both writes sit behind the same early returns, so the field and the string always describe the same sample.
- A Level Zero readout carrying only a duty cycle (`rpm == None`, `percent == Some(40)`) leaves the typed field unset. A percentage stored in a field named `_rpm` would be exported as a wildly wrong RPM; the percentage still reaches snapshots through the detail string.
- `src/api/metrics/gpu.rs`: exports `all_smi_gpu_fan_speed_rpm` as a gauge, gated on presence, following the `all_smi_gpu_performance_state` block. It prefers the typed field and falls back to parsing the legacy detail string, so mock servers and remote nodes running older builds keep exporting the series. The parser handles both `"1450 RPM"` and the Level Zero `"1600 RPM (40%)"` shape and rejects a bare `"40%"`.
- `src/network/metrics_parser.rs`: reconstructs `fan_speed_rpm` from `all_smi_gpu_fan_speed_rpm`, rejecting fractional and out-of-range values the way the P-state and NUMA handlers already do (new `MAX_GPU_FAN_RPM` ceiling of 100000).
- `src/ui/renderers/gpu_renderer.rs`: renders `Fan:1450rpm` on the existing thermal/P-state secondary row. The "does this row exist" predicate is now a shared helper used by both the renderer and the layout line count, so the reserved line count and what is actually drawn cannot drift apart. That matters here because AMD and Intel cards report none of the NVML thresholds, making fan speed the only thing that opens the row for them.
- `API.md` and `docs/LIB_mode.md`: documented the metric and the new field. `all_smi_gpu_fan_speed_rpm` moved out of the AMD-only table into the shared GPU metrics table with the real label set, since Intel emits it too.

The chosen metric name matches what `src/mock/templates/amd_gpu.rs` already emits, so mock fleets now populate the field end to end without any mock change.

## Test plan

- [x] `cargo clippy --lib --tests -- -D warnings` (also with `--features level_zero`)
- [x] `cargo fmt --check`
- [x] `cargo test --lib api::metrics::gpu` (15 passed): metric present when `Some`, absent when `None` with no detail, present via the detail fallback, structured field wins over a conflicting detail string, duty-cycle-only detail emits nothing, plus a table-driven parser test over every reader value shape
- [x] `cargo test --lib network::metrics_parser` (54 passed): genuine exporter to parser round-trip through `GpuMetricExporter`, absence stays `None`, and out-of-range/fractional values are rejected
- [x] `cargo test --lib device::readers::amd_adl` (24 passed): typed field asserted alongside the existing `Fan Speed` detail assertions, and an empty readout leaves both unset
- [x] `cargo test --lib device::readers::intel_gpu_linux` (31 passed): new fixtures assert that one hwmon `fan1_input` read reaches both the field and the detail string, and that a card without the file leaves both unset
- [x] `cargo test --lib --features level_zero device::readers::intel_gpu_level_zero` (47 passed): hwmon priority now covers the typed field, duty-cycle-only leaves it unset, and L0 fills a gap the hwmon baseline left
- [x] `cargo test --lib ui::renderers::gpu_renderer` (41 passed): fan rendered when present, omitted when absent, no double leading space when fan is the only field on the row, correct separation from its neighbours, and fan independently bumps the layout line count
- [x] `cargo test --test snapshot_test` (13 passed), `cargo test --test thermal_pstate_integration_test` (4 passed)

Not verified: `cargo check --target x86_64-pc-windows-gnu --lib` fails on this host in `zstd-sys` because the mingw cross-compiler (`x86_64-w64-mingw32-gcc`) is not installed, which is unrelated to this change. The ADL code path that changed (`apply_to_gpu_info`) is not `cfg`-gated and its tests do run on Linux; the Windows-gated `GpuInfo` literals in `amd_windows.rs`, `intel_gpu_windows.rs`, and `windows_gpu_perf.rs` only gained `fan_speed_rpm: None` and were reviewed by hand.

Closes #352


inureyes:

### Intent

Promote GPU fan speed from an untyped `detail["Fan Speed"]` string to `GpuInfo::fan_speed_rpm: Option<u32>`, and wire it through the Prometheus exporter, the remote metrics parser, and the TUI (issue #352).

### Findings Addressed (1af7963)

- [x] **The rolling-upgrade fallback sat where it could never run.** (HIGH) `GpuMetricExporter` only ever renders locally read `GpuInfo` (`api::collection_loop.rs:143`, `snapshot/serializers/prometheus.rs`), and the mock server emits exposition text without building a `GpuInfo`, so the detail-string fallback at `src/api/metrics/gpu.rs:373-391` could not reach either case its comment named. Meanwhile an older node *does* put the reading on the wire: `export_device_info` exports every `detail` key as a label on `all_smi_gpu_info`, and `sanitize_label_name("Fan Speed")` is `fan_speed`. `MetricsParser`'s label allowlist dropped it, so a new viewer scraping an old node showed no fan. The `gpu_info` handler now recovers it, reusing the exporter's `parse_fan_speed_detail` (now `pub(crate)`, following the `parse_start_time_seconds_public` precedent) and repeating the metric handler's integer and `MAX_GPU_FAN_RPM` bounds. Recovery is gap-filling only; the dedicated metric stays authoritative in either line order, asserted both ways. A drift-guard test ties `sanitize_label_name(FAN_SPEED_DETAIL_KEY)` to the label constant.
- [x] **The dual-write invariant was asserted but not enforced.** (MEDIUM) `intel_gpu_level_zero::apply_fan` wrote the detail string unconditionally but the typed field only for `rpm == Some(_)`. On the Windows `overwrite_existing` path a duty-cycle-only readout would replace the string with a percentage while leaving an RPM from an earlier sample in the field, and the exporter would publish that stale number. Unreachable today (the Windows Intel reader rebuilds each `GpuInfo` per refresh and `windows_gpu_perf::augment_gpus` never touches fan), but the comment above it claimed otherwise. Assignment is now unconditional, with a regression test.
- [x] **Overlap with the chassis fan model was left implicit.** (LOW) Issue #352 asked for it to be explicit. The `fan_speed_rpm` doc now states it is the per-device tachometer and is deliberately separate from `ChassisInfo::fan_speeds` / `all_smi_chassis_fan_speed_rpm`.

### Remaining Items (follow-ups, not defects of this PR)

- [ ] `intel_gpu_sysfs::read_fan_rpm` filters `rpm > 0` (`src/device/readers/intel_gpu_sysfs.rs:196-206`), so an Intel card idling at a genuine 0 RPM reports `None`, which the new contract reserves for "no tachometer". `amd_adl` distinguishes the two correctly via the ADL `supported` flag and does report `Some(0)`. Pre-existing and not touched here; flipping it would newly publish `0` for cards that expose `fan1_input` without a working fan, so it needs its own issue. (LOW)
- [ ] A Level Zero card reporting only a duty cycle leaves the number reachable solely through the detail string. A companion `fan_speed_percent` field would close that, but it is outside this issue's scope. (LOW)
- [ ] NVIDIA never populates the field because NVML exposes a duty-cycle percentage, which the reader now documents in place. `nvmlDeviceGetFanSpeedRPM` would be a separate change. (LOW)
- [ ] The thermal / fan / P-state row is not width-aware (pre-existing). Adding fan moves the all-fields-populated worst case from roughly 69 to 83 columns. Not reachable with real readers, since NVIDIA populates the thresholds but no fan and AMD / Intel the reverse. (LOW)

### Verification

- [x] All stated requirements implemented (all 7 acceptance criteria verified in code, not just claimed)
- [x] Integrated into project code flow: `print_gpu_info` is called from `view/frame_renderer.rs:379,401`; the metric flows through the shared `render_prometheus_exposition` used by both `/metrics` and `snapshot --format prometheus`; CSV and JSON snapshots pick the field up through the generic dot-path resolver
- [x] Project conventions followed (inline format args, `#[serde(default)]` matching the 8 sibling optional fields, presence-gated metrics, bounded parser input like the P-state and NUMA handlers)
- [x] Existing modules reused where applicable; the layout predicate is shared rather than duplicated, and the fan-string parser is now shared between exporter and parser
- [x] No unintended structural changes: 81 `GpuInfo` literals were verified complete field by field, including the three Windows-gated files that do not compile on a Linux host, and every reader whose module docs say it has no fan source correctly passes `None`
- [x] Tests pass: `cargo fmt --check`, `cargo check --lib --tests`, `cargo clippy --lib --tests -- -D warnings` (and with `--features level_zero`, the only build that compiles the Level Zero apply path), `network::metrics_parser` 58, `api::metrics::gpu` 15, `ui::renderers::gpu_renderer` 41, `device::readers::intel_gpu_level_zero` 48 with the feature, plus the `thermal_pstate_integration_test`, `snapshot_test`, and `hardware_details_integration_test` binaries

Wire-compatibility holds in both directions: `#[serde(default)]` is present, `GpuInfo` sets no `deny_unknown_fields`, and the Prometheus format is additive. With the parser fix, an old node's reading now also survives the trip to a new viewer.

inureyes:

Focus was the remote-metrics ingestion path (`src/network/metrics_parser.rs`), the newly shared `parse_fan_speed_detail`, exposition-format safety, and the per-frame render cost. **No CRITICAL or HIGH findings. Nothing auto-fixed, nothing committed.**

### Ingestion path: clean

`f64 -> u32` on both fan paths is guarded three times over, so no overflow, truncation, NaN, or negative can land in the field:

1. The production scrape regex (`src/view/data_collection/remote_collector.rs:67`) captures the value as `[\d\.]+`, so `-1`, `NaN`, `inf`, and `1e300` never even produce a parsed line.
2. Both handlers require `(0.0..=MAX_GPU_FAN_RPM).contains(&value) && value.fract() == 0.0`. `Range::contains` is false for NaN, so that case is covered independently of the regex. Enforced on the dedicated metric (`metrics_parser.rs:615`) and on the legacy label recovery (`metrics_parser.rs:578`), as intended.
3. `saturating_u32` (`metrics_parser.rs:1079`) rejects NaN and negatives and saturates at `u32::MAX`, and by that point the value is already in `[0, 100000]` with no fractional part, so the `as u32` cast is exact.

`parse_fan_speed_detail` (`src/api/metrics/gpu.rs:37`) cannot panic and allocates nothing: `split_once` and `trim` return slices, `parse::<f64>` allocates nothing, and `is_finite() && rpm >= 0.0` rejects `inf`/`NaN`/negative. Making it `pub(crate)` weakened no caller: the remote parser re-applies the full `MAX_GPU_FAN_RPM` and integer bounds on top of it, and the exporter is the only other caller.

Allocation and complexity on adversarial input are bounded by existing caps that the new code inherits: 10 MB total scrape text, 1024 bytes for the entire label section per line (`parse_prometheus!`), 100 labels of 1024 bytes each (`parse_labels`), 256 GPU rows. The new work per `all_smi_gpu_info` line is one hash lookup plus one `split_once`, all O(1) amortized.

### Exposition-format injection: clean

The fan value reaches the exposition only as a number (`u32`, or an `f64` whose `Display` never emits a newline), so it cannot break the line format. The `Fan Speed` detail key still goes through `sanitize_label_name`, which maps every non-alphanumeric byte to `_` and yields `fan_speed`, so no quote or newline can reach a label name; the value goes through `sanitize_label_value` (control-char strip, 1024-byte truncation) and then `MetricBuilder::metric`'s escaping. `legacy_fan_speed_label_matches_the_exporter_spelling` pins that contract. No other detail key in the tree sanitizes to `fan_speed`, so there is no duplicate-label-name hazard.

### Performance: no regression

Nothing new runs per frame that did not already. The renderer adds one `format!("{rpm}rpm")` per GPU per frame when the field is present, matching the four sibling threshold fields on the same row. Folding the row predicate into `gpu_has_thermal_pstate_row` is a net win: it removes the possibility of the reserved line count and the drawn row disagreeing. The exporter's `detail.get` fallback runs only when the typed field is `None`.

### MEDIUM

**1. Fan RPM is the only reader-side physical quantity without a validation cap.** `src/device/readers/intel_gpu_linux.rs:431-440` explicitly clamps temperature, frequency, power, and memory with the comment "so that a garbled sysfs file can never propagate `u32::MAX` into the exporter", but `fan_rpm` (read at line 430) skips that block and flows straight to `fan_speed_rpm` at line 549. Same gap in `src/device/readers/amd.rs:426`, `src/device/readers/amd_adl.rs:158`, and `src/device/readers/intel_gpu_level_zero/apply.rs:136` (`read_fan_speed` in `point.rs:367` bounds only to `speed >= 0`, so up to `i32::MAX`).

Before this PR an implausible value only reached the untyped detail string. It now reaches `all_smi_gpu_fan_speed_rpm` and the TUI row, where it renders as `Fan:4294967295rpm` (17 columns) and produces a gauge that remote all-smi viewers silently drop because it exceeds `MAX_GPU_FAN_RPM`, so Prometheus and the viewer disagree. Not attacker-reachable (sysfs and the vendor drivers are kernel-owned, and there is no sysfs-root override), which is why this is MEDIUM rather than HIGH. Suggested fix: promote `MAX_GPU_FAN_RPM` to a shared constant and clamp at each reader boundary, matching the pattern already applied to the four neighbouring quantities.

**2. The exporter's detail fallback does not apply the bounds the parser does.** `src/api/metrics/gpu.rs:392-402` emits whatever non-negative finite `f64` `parse_fan_speed_detail` returns, with no ceiling and no integer check, while `metrics_parser.rs:578` rejects anything fractional or above 100000. A `"1450.5 RPM"` detail is therefore published as `1450.5` and then dropped by every all-smi viewer. Cheapest fix is to reuse the same guard in the exporter so the two sides agree by construction.

### LOW

**3. The "dedicated metric always wins" comment is stronger than the code.** `metrics_parser.rs:565-574` states the `gpu_fan_speed_rpm` handler "assigns unconditionally, so it overwrites anything recovered here". That holds only when the handler's guard passes. If a node publishes both an out-of-range or fractional dedicated metric and an in-range legacy `fan_speed` label, the rejected metric leaves the legacy value in place, so the legacy label wins. Behaviour is safe (the surviving value already passed the same bounds), but the comment should say "when it passes its bounds check".

**4. Readers still hardcode the `"Fan Speed"` literal.** `FAN_SPEED_DETAIL_KEY` is defined in `src/api/metrics/gpu.rs:23` but `amd.rs:427`, `amd_adl.rs:160`, `intel_gpu_linux/sources.rs:66`, and `intel_gpu_level_zero/apply.rs:127` all write the string literal. `legacy_fan_speed_label_matches_the_exporter_spelling` pins the const against `sanitize_label_name`, not against the readers, so it would not catch a reader-side rename. Reader unit tests would, but importing the const is the tighter guarantee.

**5. Zero RPM means different things per backend.** `intel_gpu_sysfs::read_fan_rpm` filters `rpm > 0`, so a stopped hwmon fan becomes `None`, while `point.rs:367` returns `Some(0)`, so the same stopped fan under Level Zero becomes `Some(0)` and renders `Fan:0rpm`. That is exactly the reading the field doc says must never appear. Pre-existing for the detail string, but promotion to a typed field and an exported gauge makes it visible.

**6. The `overwrite_existing` guard keys on the detail map, not the field.** `intel_gpu_level_zero/apply.rs:118` still gates on `detail.contains_key("Fan Speed")` while line 136 writes `fan_speed_rpm`. Correct today because all four readers write both in lockstep, but a future reader that sets only the typed field would silently opt out of the guard and have its value clobbered.

**7. Pre-existing, noted for completeness.** `MetricBuilder::help`/`type_` push unconditionally inside the per-GPU loop, so an N-GPU node emits N duplicate `# HELP`/`# TYPE` lines for `all_smi_gpu_fan_speed_rpm`. Strict Prometheus text parsers reject a repeated HELP line for a metric family. The new metric follows the house pattern used by `all_smi_gpu_performance_state` and the four thresholds, so this is not a regression and is out of scope here.

### Verification

`cargo check --lib --tests`, `cargo clippy --lib --tests -- -D warnings` (clean), `cargo test --lib network::metrics_parser` (58 passed), `cargo test --lib api::metrics::gpu` (15 passed).

### Verdict

Low risk. The fan ingestion path is more tightly bounded than the surrounding pre-existing handlers (the thermal-threshold arms at `metrics_parser.rs:587-598` still accept any non-negative `u32` with no ceiling). Safe to merge as is; the two MEDIUM items are worth a follow-up.

inureyes:
## PR Finalization Complete

### Security review follow-up
Fixed both MEDIUM findings from the security review:

1. **Reader-side fan RPM clamp**: `intel_gpu_linux.rs`, `amd.rs`, `amd_adl.rs`, and `intel_gpu_level_zero/apply.rs` now clamp the raw tachometer reading before it reaches either `GpuInfo::fan_speed_rpm` or the `Fan Speed` detail string, the same defence already applied to temperature, frequency, power, and memory. The bound (`MAX_GPU_FAN_RPM`) now lives once in `device::types` alongside `GPU_METRIC_UNAVAILABLE` instead of being duplicated as a private constant in `network::metrics_parser`, so all five sites agree on one limit.
2. **Exporter detail-fallback bounds**: `parse_fan_speed_detail` (shared by the exporter and `network::metrics_parser`'s legacy-node recovery) now rejects a fractional or out-of-range RPM itself, so a garbled or hand-edited `"1450.5 RPM"` detail string can no longer be published for downstream parsers to silently drop. This also let `network::metrics_parser` remove its now-redundant duplicate of the same check.

The dual-write invariant (typed field and `Fan Speed` detail string never disagreeing) is preserved: every clamp is applied once, upstream of both writes.

### Tests
Added regression coverage for both fixes: a garbled/`u32::MAX` sensor reading test in each of the four readers (`intel_gpu_linux`, `amd`, `amd_adl`, `intel_gpu_level_zero::apply`), plus `parse_fan_speed_detail` and exporter-level tests for fractional and out-of-range detail strings in `api::metrics::gpu`. All pre-existing tests continue to pass.

### Documentation
`API.md` and `docs/LIB_mode.md` already documented `all_smi_gpu_fan_speed_rpm` and `fan_speed_rpm` accurately; no changes were needed. No Korean translations of these docs exist in the repository, so there was nothing to update there.

### Lint/Format
`cargo fmt --check` and `cargo clippy --lib --tests -- -D warnings` (both with and without `--features level_zero`) are clean.

### Validation
- `cargo test --lib api::metrics::gpu` (18 passed)
- `cargo test --lib network::metrics_parser` (58 passed)
- `cargo test --lib device::readers::intel_gpu_linux` (32 passed)
- `cargo test --lib device::readers::amd` (30 passed, covers `amd` and `amd_adl`)
- `cargo test --lib --features level_zero device::readers::intel_gpu_level_zero` (49 passed)

Issue #357, opened 2026-08-07
test: verify the Windows SCM service backend on a Windows machine (#311)


inureyes:
## Background

#311 (PR #320) landed the Windows Service Control Manager backend of the `all-smi service` subcommand. The code is written, integrated, and dispatched from the shipped binary. What is not done is verification: all seven acceptance criteria from #311 were deliberately left unticked because they need a Windows machine, which was not available during implementation. Every one of those gaps is recorded honestly in PR #320's body, and none of them is a code defect.

This is group D of the four groups split out of #332. It carries `priority:high` where the other three carry `priority:medium`, because nothing in the Windows backend has ever been executed or even linked, which makes it the largest unverified surface of the three platforms.

## Why this was split out

The four groups in #332 are unblocked by four unrelated environments and are independently actionable. Holding them in one issue meant none of them could be scheduled, assigned, or closed independently: whoever gained access to one environment could not close anything. #332's own first acceptance criterion permits exactly this action, and its closing note says to shrink it as groups get split out. #332 stays open as the umbrella until all four children are resolved.

## Scope

Three of the original 19 unticked items across #309, #310 and #311 are already tracked by #330, which covers the `systemd-service` CI job's system-scope lifecycle path never having executed. Those three are **excluded** here and from the other three split issues so the trackers do not overlap:

- The `kill -9` crash-restart check.
- The tarball system-scope install flow (`sudo all-smi service install --now`, then `uninstall` leaving no unit).
- `/etc/all-smi/config.toml` being honored by a running daemon.

None of the three is a Windows item; the exclusion is stated here for the same reason #332 stated it globally.

## Needs a Windows machine (all of #311)

- [ ] On Windows 10/11 or Server, from an elevated prompt: `all-smi.exe service install --now` registers and starts the service; `curl http://localhost:9090/metrics` returns GPU/CPU/memory metrics; `all-smi.exe service status` shows running with pid; the service survives reboot without login.
- [ ] `service stop` and an SCM-initiated stop (services.msc) both produce a clean shutdown: SERVICE_STOPPED within the wait hint, the energy WAL flush line in the log, no orphan process.
- [ ] Recovery works: `taskkill /F` on the process leads to an automatic SCM restart per the configured failure actions.
- [ ] Non-elevated install/uninstall/start/stop exits 1 with the elevation message.
- [ ] `service run` from a normal console fails gracefully with the SCM-entry-point explanation.
[...]

Issue #356, opened 2026-08-07
test: verify the launchd system scope on macOS with root and a reboot (#310)


inureyes:
## Background

#310 (PR #321) landed the launchd backend of the `all-smi service` subcommand. The code is written, integrated, and dispatched from the shipped binary on macOS. What is not done is verification of the system scope: four acceptance criteria from #310 were deliberately left unticked because they need root on a macOS host plus a reboot, and no authorization for root-level macOS changes was available during implementation. Every one of those gaps is recorded honestly in PR #321's body, and none of them is a code defect.

This is group C of the four groups split out of #332.

## Why this was split out

The four groups in #332 are unblocked by four unrelated environments and are independently actionable. Holding them in one issue meant none of them could be scheduled, assigned, or closed independently: whoever gained access to one environment could not close anything. #332's own first acceptance criterion permits exactly this action, and its closing note says to shrink it as groups get split out. #332 stays open as the umbrella until all four children are resolved.

## Scope

Three of the original 19 unticked items across #309, #310 and #311 are already tracked by #330, which covers the `systemd-service` CI job's system-scope lifecycle path never having executed. Those three are **excluded** here and from the other three split issues so the trackers do not overlap:

- The `kill -9` crash-restart check.
- The tarball system-scope install flow (`sudo all-smi service install --now`, then `uninstall` leaving no unit).
- `/etc/all-smi/config.toml` being honored by a running daemon.

## Needs root on a macOS host plus a reboot (from #310)

- [ ] `sudo brew services start all-smi` bootstraps into the system domain; the service persists across reboot without login.
- [ ] From a zip or tarball install: `sudo all-smi service install --now` creates the LaunchDaemon; `all-smi service status` reports running with a pid; `sudo all-smi service uninstall` boots it out and removes the plist, leaving logs in place.
- [ ] The boot-time daemon context reports the full Apple Silicon metric set, or degradations are documented with a follow-up issue.
- [ ] `/Library/Application Support/all-smi/config.toml` is honored and listed by `all-smi config path`.

Notes:

- Only the user scope was verified, and deliberately so. A LaunchAgent was installed into `gui/$UID` on an M1 Ultra running macOS 26.6, checked end to end (install, start, restart, stop, uninstall, idempotent reinstall, marker refusal, `launchctl disable` handling, 72 `all_smi_*` lines served, energy WAL flush on bootout), and removed. `/Library/LaunchDaemons` was never touched and `sudo launchctl bootstrap system` was never run.
[...]

Issue #355, opened 2026-08-07
test: apply and verify the service block in lablup/homebrew-tap (#310)


inureyes:
## Background

#310 (PR #321) landed the macOS launchd backend and prepared the Homebrew service integration. This follow-up verifies the two Homebrew-specific acceptance criteria against the production tap and the current `all-smi` implementation.

This is group B of the four groups split out of #332.

## Scope

Three lifecycle checks remain owned by #330 and are excluded here: crash restart, tarball system-scope install/uninstall, and system configuration loading. This issue covers only the Homebrew service block and the Homebrew-path refusal behavior from #310.

## Homebrew verification (from #310)

- [x] The tap formula gains the service block; `brew style` passes; a fresh/reinstalled formula starts with `brew services start all-smi`; `curl localhost:9090/metrics` returns data; the log file receives tracing output.
- [x] A Homebrew-path current binary refuses subcommand installation without `--force` and prints the `brew services` guidance.

Evidence from 2026-08-24 KST (2026-08-23 UTC):

- The service block was merged in [lablup/homebrew-tap#1](https://github.com/lablup/homebrew-tap/pull/1) (`b6874cfd9371`). It includes `process_type :background` so the brew-managed service uses the same QoS class as the subcommand-installed service.
- `brew style lablup/tap/all-smi` passed: `1 file inspected, no offenses detected`.
- After refreshing the installed formula receipt with `brew reinstall lablup/tap/all-smi`, `brew services start lablup/tap/all-smi` succeeded. `/metrics` became ready after 7 seconds and returned 69 `all_smi_*` metric lines.
- The generated LaunchAgent used `/opt/homebrew/opt/all-smi/bin/all-smi api` and `ProcessType=Background`; `/opt/homebrew/var/log/all-smi.log` received API startup and request tracing.
- A current-main executable placed under a disposable Homebrew prefix exited 1 from `service install --user --now`, advised `brew services start all-smi` (and the sudo system-daemon variant), and offered `--force` as the explicit override.
- The brew service was stopped after verification and the prior stopped state was restored.

## Scope decisions and follow-up

- Existing pre-service-block 0.25.0 receipts do not learn the service DSL until the formula is reinstalled. Per the release preparation decision, compatibility for those existing receipts is intentionally out of scope: the imminent release will carry the current formula metadata, so no tap revision or follow-up compatibility issue is being added here.
[...]

Issue #354, opened 2026-08-07
test: verify the systemd service path on a Linux host with dpkg (#309)


inureyes:
## Background

#309 (PR #319) landed the systemd backend of the `all-smi service` subcommand. The code is written, integrated, and dispatched from the shipped binary on Linux. What is not done is verification: three acceptance criteria from #309 were deliberately left unticked because they need hardware or an environment that was not available during implementation (no systemd host, no dpkg). Every one of those gaps is recorded honestly in PR #319's body, and none of them is a code defect.

This is group A of the four groups split out of #332.

## Why this was split out

The four groups in #332 are unblocked by four unrelated environments and are independently actionable. Holding them in one issue meant none of them could be scheduled, assigned, or closed independently: whoever gained access to one environment could not close anything. #332's own first acceptance criterion permits exactly this action, and its closing note says to shrink it as groups get split out. #332 stays open as the umbrella until all four children are resolved.

## Scope

Three of the original 19 unticked items across #309, #310 and #311 are already tracked by #330, which covers the `systemd-service` CI job's system-scope lifecycle path never having executed. Those three are **excluded** here and from the other three split issues so the trackers do not overlap:

- The `kill -9` crash-restart check.
- The tarball system-scope install flow (`sudo all-smi service install --now`, then `uninstall` leaving no unit).
- `/etc/all-smi/config.toml` being honored by a running daemon.

## Needs a systemd Linux host with dpkg (from #309)

- [ ] A deb built from this tree installs the unit disabled; `sudo systemctl enable --now all-smi` followed by `curl -s localhost:9090/metrics | head` returns metrics; `journalctl -u all-smi` shows tracing output; `systemctl stop all-smi` takes the graceful SIGTERM path (WAL flush logged).
- [ ] A non-systemd environment (for example a container without systemd) produces the `NotSupported` message, exit 1.
- [ ] A dpkg-managed binary plus `all-smi service install` without `--force` refuses with the guidance message.

Notes:

- The `NotSupported` reporting machinery (message rendering, exit code) was exercised on macOS during #309, when `backend()`'s macOS arm was still a placeholder returning `NotSupported`. That arm no longer exists: #310 replaced it with the launchd backend. The Linux branch that matters here, `SYSTEMD_RUNTIME_MARKER` absent at `src/service_cmd/systemd.rs:47-56`, has only ever been compile-checked and unit-tested against fixtures. It has never been hit on a real host lacking `/run/systemd/system`.
[...]

PR #361, opened 2026-08-07
feat: attribute ADL telemetry per card on multi-GPU AMD Windows hosts

### LINKED ISSUE #353: feat: support multi-GPU AMD hosts in the ADL reader via AdapterInfo
## Problem / Background

`src/device/readers/amd_adl.rs` only augments when the machine has exactly one AMD GPU (`can_attribute(amd_gpu_count) == amd_gpu_count == 1`). On a host with two or more AMD GPUs, temperature, power, fan, and clocks are all absent and the reader falls back to the DXGI/PDH baseline from #348. The most common affected configuration is a Ryzen APU laptop with a Radeon dGPU, which presents two AMD rows.

## Why the limit exists

ADL identifies adapters by an index that cannot be tied to a physical card without `AdapterInfo`, a 1568-byte struct carrying the PCI bus/device/function numbers and the PNP string. Worse, a single card exposes several ADL adapter indices, one per display output, all reporting identical telemetry, so indices cannot even be deduplicated without it. `src/device/readers/amd_adl/ffi.rs` deliberately declines to declare that struct, on the grounds that a hand-transcribed layout is memory corruption rather than a compile error. Rather than guess, the reader declines to attribute, matching the conclusion the #346 review reached about adapter matching: declining beats attributing wrongly.

## Why this is more tractable than the original note suggested

Two corrections to the reasoning recorded in `ffi.rs` and the #349 technical report:

1. `ADL2_Adapter_AdapterInfo_Get(ADL_CONTEXT_HANDLE, LPAdapterInfo, int iInputSize)` takes the buffer size **as a parameter**. The claim that "ADL sizes its write by its own sizeof, so getting it wrong overflows our buffer" is therefore overstated. Passing a correct `iInputSize` together with the padded-buffer pattern already proven for `AdlPmLogDataBuffer` in `ffi.rs` largely removes the overflow risk. The residual risk is reading fields at wrong offsets and getting garbage, which is bad but is not memory corruption.
2. The struct is **self-verifying**. It contains several 256-byte `char` arrays (`strUDID`, `strAdapterName`, `strDisplayName`, `strPNPString`) at known offsets. If the layout is right they read as legible ASCII; if wrong they read as garbage. A doctor check that dumps them confirms the layout empirically from real hardware, exactly the pattern `amd.adl.sensors` already uses to make the PMLog sensor index mapping field-verifiable.

## Payoff

`strPNPString` is the same `PNPDeviceID` that `src/device/readers/amd_windows.rs` already stores as the GPU uuid, so the join becomes exact rather than ordinal. That satisfies rather than circumvents the constraint the #346 review imposed on `match_adapter`, which now requires PCI vendor agreement and refuses to guess.

## Proposed Solution

- Declare `AdapterInfo` in `src/device/readers/amd_adl/ffi.rs` with `#[repr(C)]`, compile-time `size_of` assertions (expected 1568 bytes: 4 + 256 + 4*4 + 256 + 256 + 4 + 4 + 256 + 256 + 256 + 4, all 4-byte aligned), and a test asserting the string field offsets.
- Call `ADL2_Adapter_AdapterInfo_Get` with an over-allocated buffer and a correct `iInputSize`.
- Group adapter indices per physical card by bus/device/function, then match to the reader's GPUs by `strPNPString` against the GPU uuid, falling back to bus/device/function against the PCI ids DXGI already provides.
- Add an `amd.adl.adapters` doctor check dumping index, BDF, `strAdapterName`, and `strPNPString` per adapter, so the layout and the matching can both be confirmed from a real machine.
- Relax `can_attribute` once matching is exact.

## Blocking prerequisite

This needs a real multi-AMD-GPU Windows machine to validate against before it can be trusted. Nothing in CI compiles all-smi for Windows (the only Windows job is gated behind an unset repository variable and has never executed), so the layout cannot be checked automatically. Do not merge this on cross-compile evidence alone.

## Acceptance Criteria

- [x] `AdapterInfo` declared in `ffi.rs` with `size_of` and field-offset assertions
- [x] Adapter indices deduplicated per physical card by bus/device/function
- [x] Exact match to reader GPUs via `strPNPString`, falling back to bus/device/function
- [x] `can_attribute` relaxed to allow two or more AMD GPUs
- [x] `amd.adl.adapters` doctor check added, dumping index, BDF, `strAdapterName`, `strPNPString`
- [ ] Confirmed on a real host with two or more AMD GPUs, with the doctor output attached to the PR
- [x] Single-GPU behavior unchanged

inureyes:
## Status: layout CONFIRMED on real hardware, merging with the two-GPU observation tracked in #370

**This supersedes the "do not merge" header this PR originally carried.** That header existed because the `AdapterInfo` layout was transcribed blind from AMD's public `adl_structures.h`, and nothing in CI compiles all-smi for Windows. That risk is retired. On 2026-08-18 the branch ran on a real AMD host (Radeon(TM) 8060S Graphics, Strix Halo APU `DEV_1586`, driver 32.0.31035.1003, Windows 11 Pro 26200, native `x86_64-pc-windows-msvc`, rustc 1.97.1). The declared 1572-byte layout, its stride across five rows, and every load-bearing offset were evaluated under the real target ABI and confirmed against the live driver; `strPNPString` came back byte-identical to the WMI `PNPDeviceID`, so the exact join this PR rests on does resolve on real hardware. The run also surfaced a bug that would have made the feature dead on arrival on the very hosts it targets, fixed here by `0bebb9a`. Whole-workspace `cargo build --release` and 61 `--lib` tests passed natively. The two hardware comments below carry the full dumps.

What remains is an observation, not a code question: whether two *different physical* cards report *different* telemetry. No host with one AMD GPU can establish that, since `plan_attribution` takes its `SoleGpu` arm before the inventory is ever consulted. Holding validated code out of `main` indefinitely, waiting on hardware that may not appear, costs more than it protects here, because every failure path in this PR declines attribution and keeps the DXGI/PDH baseline rather than guessing. The remaining acceptance criterion is therefore split out to #370 as a `type:test` issue, the same way #309, #310, and #311 were split into #354, #355, #356, and #357.

**For whoever has a two-AMD-GPU Windows host:** the validation procedure, the expected output shape, and what to report all live in #370.

## Summary

The Windows ADL reader bailed out on any host with two or more AMD GPUs (`can_attribute(count) == (count == 1)`), because without `AdapterInfo` an ADL adapter index cannot be tied to a physical card, and one card exposes several indices (one per display output) reporting identical telemetry. This PR declares `AdapterInfo`, groups adapter indices per physical card, matches cards to the reader's GPUs exactly, and relaxes attribution to any number of AMD GPUs while keeping every failure path a decline. The motivating configuration is a Ryzen APU plus Radeon dGPU laptop, which presents two AMD rows and previously lost temperature, power, fan, and clocks entirely.

## Struct size: 1572 bytes, not the issue's 1568

The issue body computes 1568 via `4 + 256 + 4*4 + 256 + 256 + 4 + 4 + 256 + 256 + 256 + 4`, which drops one `int`: the formula goes straight from `iSize` to `strUDID` and omits `iAdapterIndex`. Derived field by field from `adl_structures.h` (Windows branch), the layout is: `iSize` at 0, `iAdapterIndex` at 4, `strUDID` at 8, `iBusNumber` 264, `iDeviceNumber` 268, `iFunctionNumber` 272, `iVendorID` 276, `strAdapterName` 280, `strDisplayName` 536, `iPresent` 792, then the `#if defined(_WIN32) || defined(_WIN64)` tail: `iExist` 796, `strDriverPath` 800, `strDriverPathExt` 1056, `strPNPString` 1312, `iOSDisplayIndex` 1568, total 1572. That is 9 ints (36 bytes) plus 6 char[256] arrays (1536 bytes) with no padding, since every field is naturally 4-aligned. The declaration is Windows-shaped and documented as such: the Linux variant has a different tail and size, and nothing outside `cfg(target_os = "windows")` calls the entry point. Compile-time assertions pin the 1572-byte size, the 4-byte alignment, and every load-bearing offset, and they run unconditionally so the Linux test runner checks them.

## Safety design (why declaring this struct became acceptable)

- `ffi.rs`'s original refusal reasoned that ADL sizes its write by its own `sizeof` with no length parameter. That is true of `ADL2_New_QueryPMLogData_Get` but not here: `ADL2_Adapter_AdapterInfo_Get` takes the buffer size as `iInputSize`, which this PR passes correctly (the size of the requested rows).
- The padded-buffer pattern proven on `AdlPmLogDataBuffer` is preserved anyway: `AdapterInfoArray` allocates twice the requested rows plus slack, so a driver that ignores `iInputSize` and writes by a larger `sizeof` still lands inside our allocation.
- The residual risk (reading fields at wrong offsets) is caught at runtime: the struct is self-verifying because `strUDID`, `strAdapterName`, `strDisplayName`, and `strPNPString` are 256-byte char arrays at known offsets that read as NUL-terminated legible ASCII only if the layout is right. The buffer is pre-filled with a `0xAA` poison pattern and `AdapterInfo::classify` sorts each returned row into `Untouched` (poison intact, the driver never wrote it), `Blank` (all zero, the driver memset without populating, which AMD's own SDK sample treats as normal `iPresent`-style filtering), `Populated` (passes the string, `iSize`, and `iAdapterIndex` checks), or `Garbled` (written bytes contradicting the layout). `AdapterInfoArray::validated` returns only the populated rows and rejects the table when any row is garbled, when row 0 is not populated, when nothing is populated, or when populated rows repeat an `iAdapterIndex`. The two stride errors point in opposite directions and both are covered: a declared size larger than the driver's makes the driver stride shorter, so our later rows read misaligned and classify garbled; a declared size smaller than the driver's makes it write fewer rows and leave a trailing untouched run, which is accepted and under-attributes the missing cards but can never mis-attribute, since the exact PNP join in `plan_attribution` is the real gate and a garbled read cannot produce a `strPNPString` equal to a WMI `PNPDeviceID` by accident.
- Verification failure, a missing entry point (old drivers are looked up leniently), a failed call, or a failed match all produce the same outcome: decline attribution and keep the DXGI/PDH baseline, matching the conclusion the #346 review reached for `match_adapter`. A single-GPU host never consults `AdapterInfo` at all and keeps the pre-#353 behavior unchanged, pinned by a regression test.

## Known gaps in the verification, deliberately left open for hardware evidence

- **Resolved: blank rows no longer veto the table.** An earlier revision rejected any all-zero row as short-write evidence, which would have disabled multi-GPU attribution on drivers that follow AMD's own SDK pattern (memset the array, populate present adapters, let the caller filter by `iPresent`), exactly the hosts the pending hardware validation needs to work on. The poison pre-fill now separates the two facts the zero fill conflated: an all-zero row proves the driver wrote zeros (`BLANK`, accepted and dropped), while a row still carrying the poison proves the driver never wrote it (`UNTOUCHED`, accepted only past a populated row 0). Interleaved blanks, the signature of healthy filtering, are deliberately not rejected; a stride mismatch produces garbled rows or a trailing untouched run instead. The worst case of accepting non-populated rows is under-attribution, never mis-attribution. What the hardware dump should now show: every row tagged with its state, ideally all rows populated or a mix of populated and `BLANK`; a run of `UNTOUCHED` rows after a populated row 0 would be the signature of a declared struct smaller than the driver's and should be reported. See `AdapterInfo::classify` and `AdapterInfoArray::validated` in `ffi.rs`.
- **Only 4 of the struct's 6 string fields are checked by `looks_sane`.** `strUDID`, `strAdapterName`, `strDisplayName`, and `strPNPString` are verified; `strDriverPath` (offset 800) and `strDriverPathExt` (offset 1056) are not. Both sit strictly between `strDisplayName` (536) and `strPNPString` (1312), so a `strPNPString` parse that succeeds already means the stride carried the two skipped fields through intact, and the marginal detection gain of checking them too is small. Each additional strictness arm is also a new way for this check to decline on real hardware nobody has tested yet. Revisit after a real host produces a passing dump. See the doc comment on `AdapterInfo::looks_sane` in `ffi.rs`.

## What changed

- `src/device/readers/amd_adl/ffi.rs`: `AdapterInfo` declared `#[repr(C)]` with compile-time size/align/offset assertions and offset tests; `adl_string` / `adl_string_lossy` helpers; `AdapterInfo::looks_sane` runtime layout check; `AdapterInfoArray` padded buffer with `validated()`; `Adl2AdapterAdapterInfoGet` entry point type (`extern "C"`, Windows-gated per file convention); module docs rewritten, since they previously explained at length why the struct was absent. The buffer is poison-filled (`AdapterInfo::poisoned`, `0xAA`) rather than zeroed, `AdapterInfo::classify` assigns each row a `RowState` (`Untouched` / `Blank` / `Populated` / `Garbled`), and `validated()` returns the populated rows only, rejecting on any garbled row, a non-populated row 0, an empty result, or repeated `iAdapterIndex` values; the remaining verification gap above is recorded on `looks_sane`.
- `src/device/readers/amd_adl/adapters.rs` (new, platform-independent like `windows_gpu_perf::ids`): `parse_adapters`, `group_by_card` (per-card grouping by bus/device/function with per-row BDF plausibility filtering, dropping a group whose rows carry two different non-empty PNP instance paths rather than merging them), `plan_attribution` (exact `strPNPString` join first, unambiguous PCI vendor+device fallback second, no ordinal or name guessing, injectivity enforced, `SoleGpu`/`PerCard`/`Decline`), and `describe_raw_entry` for the doctor dump. Reuses `parse_pnp_device_id` from `windows_gpu_perf::ids` rather than reimplementing PCI id parsing. Also now the platform-independent home for `MAX_ADAPTER_ROWS`, `plausible_adapter_count`, `clamp_scan_count`, and `describe_layout_failure`, moved out of the Windows-only `loader.rs` and `doctor/checks/amd.rs` so this arithmetic and message-shaping logic is unit-tested on the Linux runner instead of living entirely behind `cfg(target_os = "windows")` with zero coverage. `describe_raw_entry` now tags each dump line with the row's state, and `describe_layout_failure` names the four rejection shapes (garbled rows, a dead call with row 0 still poisoned, an empty memset, repeated adapter indices).
- `src/device/readers/amd_adl.rs`: `can_attribute` replaced by `adapters::plan_attribution`; `augment` attributes per matched card on multi-GPU hosts (trying each of a card's indices until one answers PMLog) and keeps the single-GPU path byte-identical, including not fetching the inventory at all; module docs updated.
- `src/device/readers/amd_adl/loader.rs`: lenient lookup of `ADL2_Adapter_AdapterInfo_Get`; `adapter_inventory()` cached on the existing 60-second rescan interval so topology changes self-heal; `sample_adapter(index)` for per-card PMLog reads; `adapter_info_probe()` returning every raw row plus the accepted populated subset (`AdapterProbe::Rows { rows, accepted }`) for the doctor, with the cached inventory parsing only the populated rows; `scan_for_capable_adapter` and `probe_adapter_info` now call into `adapters::clamp_scan_count` / `adapters::plausible_adapter_count` instead of inlining the bound.
- `src/doctor/checks/amd.rs`: new `amd.adl.adapters` check (template: `amd.adl.sensors`) dumping index, bus/device/function, `strAdapterName`, and `strPNPString` per adapter, printing the raw rows even, and especially, when layout verification fails, since garbage strings are the evidence a wrong transcription produces; each dump line now carries the row's state tag, and the failure-shape message comes from the tested `adapters::describe_layout_failure`.
- `README.md`: the `amd.*` doctor check-ID table now lists `amd.adl.library`, `amd.adl.sensors`, and `amd.adl.adapters` alongside the pre-existing Linux checks; it previously listed none of the three ADL checks.

## Test plan

- [x] `cargo fmt -- --check`
- [x] `cargo test --lib device::readers::amd_adl` (61 passed: layout offsets, string sanity, padded array, row classification into untouched/blank/populated/garbled, trailing-untouched acceptance, interleaved-blank acceptance, non-populated-row-0 rejection, duplicate-index rejection, wrong-stride garble rejection, grouping with several indices per card and two cards, the BDF-disagreement group drop, APU+dGPU exact matching, PCI-id fallback, identical-cards decline, cross-vendor decline, failed-verification decline, single-GPU regression pin, the checked `input_size` conversion, the extracted adapter-count bound/clamp functions, and the four doctor failure-shape messages plus per-row state tags)
- [x] `cargo test --lib doctor` (25 passed)
- [x] `cargo test --lib device::readers::windows_gpu_perf` (31 passed)
- [x] Windows-target compile check of the ADL modules in isolation: `ffi.rs`, `adapters.rs`, and `loader.rs` copied into a throwaway crate with a six-line stub for `windows_gpu_perf::ids`, then `cargo check --target x86_64-pc-windows-gnu`. Passes; flipping the 1572 assertion to 1568 fails the build, so the layout assertions are demonstrably live under a Windows target. Re-run after the poison-fill change with byte-identical copies of the three modules plus a replica of the doctor arm: passes, and a negative control (renaming the `validated` call in the loader copy) fails the build, so the `cfg(windows)` code is demonstrably compiled
- [ ] `cargo check --target x86_64-pc-windows-gnu` on the whole workspace: fails on this host (`zstd-sys` needs `x86_64-w64-mingw32-gcc`, which is not installed), which is why the logic lives in unconditional modules wherever possible
- [ ] Confirmed on a real host with two or more AMD GPUs with doctor output attached: NOT DONE, awaiting hardware (see the status section above)

Closes #353
Refs #370

### DIFF
--- a/README.md
+++ b/README.md
@@ -606,7 +606,7 @@ Stable check IDs (greppable across versions):
 | `privileges.*` | `privileges.user`, `privileges.root`, `privileges.video_render_group`, `privileges.dev_dri`, `privileges.dev_tenstorrent` |
 | `container.*` | `container.runtime`, `container.cgroup`, `container.k8s_serviceaccount` |
 | `nvidia.*` | `nvidia.nvml.loadable`, `nvidia.smi.binary`, `nvidia.driver.version`, `nvidia.env.visible_devices`, `nvidia.mig.mode` |
-| `amd.*` | `amd.rocm.version`, `amd.libamdgpu_top.abi`, `amd.dri.perms`, `amd.build.target_env` |
+| `amd.*` | `amd.rocm.version`, `amd.libamdgpu_top.abi`, `amd.dri.perms`, `amd.build.target_env`, `amd.adl.library`, `amd.adl.sensors`, `amd.adl.adapters` |
 | `apple.*` | `apple.macos.version`, `apple.silicon`, `apple.smc` |
 | `gaudi.*` | `gaudi.hlsmi`, `gaudi.devices`, `gaudi.driver` |
 | `tpu.*` | `tpu.libtpu`, `tpu.env.name`, `tpu.accel.vendor` |
--- a/src/device/readers/amd_adl.rs
+++ b/src/device/readers/amd_adl.rs
@@ -32,21 +32,28 @@
 //! right blind, for hardware that predates the datacentre and workstation
 //! cards all-smi targets. A pre-Vega card keeps the WMI and PDH baseline.
 //!
-//! ## Scope: one AMD GPU
+//! ## Scope: attribution across cards
 //!
-//! ADL identifies adapters by an index that does not map to a card
-//! without `AdapterInfo`, a 1568-byte struct whose layout this module
-//! deliberately refuses to declare blind (see [`ffi`]). Worse, one card
-//! exposes several adapter indices, one per display output, all
-//! reporting identical telemetry, so an index cannot even be
-//! deduplicated without it.
+//! ADL identifies adapters by an index, and one card exposes several
+//! indices, one per display output, all reporting identical telemetry.
+//! `AdapterInfo` (see [`ffi`]) ties each index to a physical card by
+//! PCI bus/device/function and carries the Windows PNP string, which
+//! is the same `PNPDeviceID` the reader stores as the GPU uuid, so on
+//! a multi-GPU host the join is exact. See [`adapters`] for the
+//! grouping and matching rules.
 //!
-//! Rather than guess, augmentation only runs when the reader found
-//! exactly one AMD GPU. That covers the overwhelming majority of real
-//! machines, and on a multi-AMD-GPU host the result is the honest DXGI
-//! and PDH baseline instead of one card's temperature reported against
-//! another. This mirrors the conclusion the #346 review reached about
-//! adapter matching: declining to attribute beats attributing wrongly.
+//! The whole scheme is conditioned on the runtime layout verification
+//! in [`ffi::AdapterInfoArray::validated`]: the `AdapterInfo` layout is
+//! transcribed blind (nothing in CI compiles for Windows), so rows are
+//! only trusted after their string fields prove the layout at runtime.
+//! If verification, the ADL call, or the matching fails, multi-GPU
+//! hosts keep the honest DXGI and PDH baseline; and a machine with
+//! exactly one AMD GPU never consults `AdapterInfo` at all, keeping
+//! the behavior it had before #353. Either way the failure mode stays
+//! the one the #346 review settled on: declining to attribute beats
+//! attributing wrongly. The `amd.adl.adapters` doctor check dumps the
+//! raw rows so both the layout and the matching can be confirmed from
+//! a real machine.
 //!
 //! ## Cost
 //!
@@ -58,6 +65,7 @@
 //! out of its deepest idle power state; all-smi's intervals start at one
 //! second, which is well clear of that.
 
+pub mod adapters;
 pub mod ffi;
 pub mod sensors;
 
@@ -67,15 +75,6 @@ pub mod loader;
 use crate::device::readers::windows_gpu_perf::note_metrics_source;
 use crate::device::types::{GpuInfo, MAX_GPU_FAN_RPM};
 use sensors::AdlReadout;
-
-/// Whether an ADL readout can be attributed to a specific card.
-///
-/// See the module docs: without `AdapterInfo` an ADL adapter index
-/// cannot be tied to a GPU, so attribution is only safe when there is
-/// nothing to confuse it with.
-pub fn can_attribute(amd_gpu_count: usize) -> bool {
-    amd_gpu_count == 1
-}
 
 /// Layer an ADL readout onto a GPU that already carries the WMI, DXGI,
 /// and PDH baseline.
@@ -194,18 +193,52 @@ pub fn apply_to_gpu_info(gpu: &mut GpuInfo, readout: &AdlReadout) {
 /// Sample ADL and layer the result onto the reader's GPUs.
 ///
 /// Called after the vendor-neutral layer so ADL's readings win. A no-op
-/// when ADL is unavailable, when no adapter exposes PMLog, or when the
-/// machine has more than one AMD GPU.
+/// when ADL is unavailable or when no adapter exposes PMLog. On a
+/// single-GPU machine the process-wide sample is attributed directly,
+/// exactly as before #353; on a multi-GPU machine each GPU is
+/// augmented only when [`adapters::plan_attribution`] matched it to a
+/// card exactly, and a failed match or failed `AdapterInfo` layout
+/// verification leaves the DXGI and PDH baseline untouched.
 #[cfg(target_os = "windows")]
 pub fn augment(gpus: &mut [GpuInfo]) {
-    if !can_attribute(gpus.len()) {
-        return;
-    }
-    let Some(output) = loader::sample() else {
-        return;
+    use adapters::AttributionPlan;
+
+    // Only a multi-GPU host needs the adapter inventory; the sole-GPU
+    // path must not depend on `AdapterInfo` in any way, including not
+    // paying for the call.
+    let inventory = if gpus.len() > 1 {
+        loader::adapter_inventory()
+    } else {
+        None
     };
-    let readout = sensors::extract(&output);
-    apply_to_gpu_info(&mut gpus[0], &readout);
+    let uuids: Vec<&str> = gpus.iter().map(|gpu| gpu.uuid.as_str()).collect();
+    match adapters::plan_attribution(&uuids, inventory.as_deref()) {
+        AttributionPlan::SoleGpu => {
+            let Some(output) = loader::sample() else {
+                return;
+            };
+            let readout = sensors::extract(&output);
+            apply_to_gpu_info(&mut gpus[0], &readout);
+        }
+        AttributionPlan::PerCard(matches) => {
+            for card in matches {
+                // Every index of a card reports the same telemetry, so
+                // take the first one that answers; a card where none
+                // answers (a pre-Vega APU next to a modern dGPU) keeps
+                // its baseline.
+                let Some(output) = card
+                    .adl_indices
+                    .iter()
+                    .find_map(|&index| loader::sample_adapter(index))
+                else {
+                    continue;
+                };
+                let readout = sensors::extract(&output);
+                apply_to_gpu_info(&mut gpus[card.gpu_index], &readout);
+            }
+        }
+        AttributionPlan::Decline => {}
+    }
 }
 
 /// Non-Windows builds have no ADL. The stub keeps the surrounding logic
@@ -473,17 +506,42 @@ mod tests {
     }
 
     #[test]
-    fn attribution_is_refused_for_anything_but_a_single_amd_gpu() {
-        assert!(can_attribute(1));
-        // Zero cards is nothing to attribute to; two or more cannot be
-        // told apart without AdapterInfo.
-        assert!(!can_attribute(0));
-        assert!(!can_attribute(2));
-        assert!(!can_attribute(4));
+    fn a_single_gpu_never_depends_on_adapterinfo() {
+        // The regression pin for pre-#353 behavior: with exactly one
+        // AMD GPU the plan is SoleGpu whether the adapter inventory is
+        // present, absent, or failed layout verification, so the
+        // single-GPU machine that worked before AdapterInfo existed
+        // keeps working when it lies. The matching-level counterpart
+        // tests live in `adapters`.
+        use adapters::{AttributionPlan, plan_attribution};
+
+        let gpu = baseline_gpu();
+        let uuids = [gpu.uuid.as_str()];
+        assert_eq!(plan_attribution(&uuids, None), AttributionPlan::SoleGpu);
+        assert_eq!(
+            plan_attribution(&uuids, Some(&[])),
+            AttributionPlan::SoleGpu
+        );
+    }
+
+    #[test]
+    fn multi_gpu_attribution_declines_without_a_validated_inventory() {
+        // Two AMD GPUs and no validated AdapterInfo rows (unavailable
+        // entry point, failed call, or failed layout verification all
+        // arrive here as `None`): the reader must decline rather than
+        // guess, exactly as it did before #353.
+        use adapters::{AttributionPlan, plan_attribution};
+
+        let uuids = ["PCI\\VEN_1002&DEV_744C", "PCI\\VEN_1002&DEV_164E"];
+        assert_eq!(plan_attribution(&uuids, None), AttributionPlan::Decline);
     }
 
     #[test]
     fn augment_is_inert_when_attribution_is_refused() {
+        // On a Linux test runner the stub is inert; on a Windows dev
+        // machine without a validated adapter inventory the multi-GPU
+        // path declines. Either way two baseline GPUs must come back
+        // untouched.
         let mut gpus = vec![baseline_gpu(), baseline_gpu()];
         augment(&mut gpus);
         for gpu in &gpus {
--- a/src/device/readers/amd_adl/adapters.rs
+++ b/src/device/readers/amd_adl/adapters.rs
@@ -236,32 +236,62 @@ fn plausible_bdf(adapter: &AdlAdapter) -> bool {
         && (0..=7).contains(&adapter.function)
 }
 
+/// Whether two ADL `strPNPString` values name the same physical device.
+///
+/// ADL does not repeat one instance path across a card's rows. Observed
+/// on an AMD Radeon(TM) 8060S (driver 32.0.31035.1003): the primary row
+/// carries the base device instance path, the path WMI reports as
+/// `PNPDeviceID`, and each additional display-output row carries that
+/// same path with a `&02`, `&03`, ... suffix appended. So sameness is
+/// "one extends the other at a separator", not equality.
+///
+/// The `&` boundary is load-bearing. Without it two genuinely distinct
+/// paths that merely share a prefix, `...&0&0041` and `...&0&00410`,
+/// would read as one card. Comparison is ASCII-case-insensitive because
+/// WMI and ADL do not agree on case.
+fn shares_device_instance(a: &str, b: &str) -> bool {
+    let (short, long) = if a.len() <= b.len() { (a, b) } else { (b, a) };
+    let (short, long) = (short.as_bytes(), long.as_bytes());
+    if !long[..short.len()].eq_ignore_ascii_case(short) {
+        return false;
+    }
+    long.len() == short.len() || long[short.len()] == b'&'
+}
+
 /// Group adapter rows into one [`CardGroup`] per physical card.
 ///
 /// Groups are sorted by bus/device/function and their indices sorted
 /// ascending, so the output is deterministic regardless of the order
 /// the driver enumerated in.
 ///
-/// A group whose rows carry two different non-empty PNP instance paths
+/// A group whose rows carry two unrelated non-empty PNP instance paths
 /// is dropped rather than returned. Grouping trusts the PCI
 /// bus/device/function to identify a physical card, so a driver that
 /// leaves those fields unfilled collapses every card into one group.
 /// The caller then samples telemetry from the first index of the group
 /// that answers PMLog, which can be a different physical card's index,
 /// and one card's temperature, power, and fan would be reported for
-/// another GPU with nothing in the output saying so. Two different
+/// another GPU with nothing in the output saying so. Two unrelated
 /// instance paths under one PCI address is the observable signature of
 /// that state, and dropping the group is the same answer the rest of
-/// this module gives everywhere else: decline rather than guess. Rows
-/// of one real card share their instance path; a secondary
-/// display-output row may leave it blank, which is backfilled and is
-/// not a conflict, and case differences are not a conflict either,
-/// since WMI and ADL do not agree on case.
+/// this module gives everywhere else: decline rather than guess.
+///
+/// "Unrelated" is judged by [`shares_device_instance`], not by string
+/// equality: a card's display-output rows extend its base path rather
+/// than repeating it (see that function). A row may also leave the path
+/// blank, which is no evidence either way and is skipped. The group's
+/// own `pnp_string` is the shortest path its rows carried, which is the
+/// base path, and therefore the only form the exact join in
+/// [`plan_attribution`] can match against a WMI `PNPDeviceID`.
+///
+/// Agreement is judged once per group against that base rather than
+/// pairwise as rows arrive, so the verdict does not depend on which row
+/// the driver enumerated first.
 pub fn group_by_card(adapters: &[AdlAdapter]) -> Vec<CardGroup> {
     let mut groups: Vec<CardGroup> = Vec::new();
-    // Parallel to `groups`: whether the group saw contradictory PNP
-    // strings and therefore cannot be trusted to be one card.
-    let mut conflicting: Vec<bool> = Vec::new();
+    // Parallel to `groups`: every non-empty instance path the group's
+    // rows carried, checked for agreement once the group is complete.
+    let mut paths: Vec<Vec<String>> = Vec::new();
     for adapter in adapters {
         if !plausible_bdf(adapter) {
             continue;
@@ -271,22 +301,14 @@ pub fn group_by_card(adapters: &[AdlAdapter]) -> Vec<CardGroup> {
                 && group.device == adapter.device
                 && group.function == adapter.function
         });
-        match existing {
+        let position = match existing {
             Some(position) => {
                 let group = &mut groups[position];
                 group.indices.push(adapter.index);
-                // Prefer any row that actually carries the string; a
-                // secondary display-output row can leave one blank.
-                if !adapter.pnp_string.is_empty() {
-                    if group.pnp_string.is_empty() {
-                        group.pnp_string = adapter.pnp_string.clone();
-                    } else if !group.pnp_string.eq_ignore_ascii_case(&adapter.pnp_string) {
-                        conflicting[position] = true;
-                    }
-                }
                 if group.adapter_name.is_empty() && !adapter.adapter_name.is_empty() {
                     group.adapter_name = adapter.adapter_name.clone();
                 }
+                position
             }
             None => {
                 groups.push(CardGroup {
@@ -294,17 +316,35 @@ pub fn group_by_card(adapters: &[AdlAdapter]) -> Vec<CardGroup> {
                     device: adapter.device,
                     function: adapter.function,
                     adapter_name: adapter.adapter_name.clone(),
-                    pnp_string: adapter.pnp_string.clone(),
+                    // Resolved below, once every row of this group has
+                    // been seen and the shortest path is known.
+                    pnp_string: String::new(),
                     indices: vec![adapter.index],
                 });
-                conflicting.push(false);
+                paths.push(Vec::new());
+                groups.len() - 1
             }
+        };
+        if !adapter.pnp_string.is_empty() {
+            paths[position].push(adapter.pnp_string.clone());
         }
     }
     let mut groups: Vec<CardGroup> = groups
         .into_iter()
-        .zip(conflicting)
-        .filter_map(|(group, conflicted)| (!conflicted).then_some(group))
+        .zip(paths)
+        .filter_map(|(mut group, paths)| {
+            // No row carried a path: nothing contradicts anything, so
+            // the group survives with an empty `pnp_string` that simply
+            // matches no GPU downstream, exactly as before.
+            let Some(base) = paths.iter().min_by_key(|path| path.len()) else {
+                return Some(group);
+            };
+            if !paths.iter().all(|path| shares_device_instance(base, path)) {
+                return None;
+            }
+            group.pnp_string = base.clone();
+            Some(group)
+        })
         .collect();
     for group in &mut groups {
         group.indices.sort_unstable();
@@ -503,6 +543,96 @@ mod tests {
         assert_eq!(groups.len(), 1);
         assert_eq!(groups[0].indices, vec![0, 1, 2]);
         assert_eq!(groups[0].pnp_string, DGPU_PNP);
+    }
+
+    // Observed on a real host, 2026-08-18: AMD Radeon(TM) 8060S
+    // Graphics (Strix Halo APU), driver 32.0.31035.1003, Windows 11.
+    // ADL enumerated five rows for the single physical card, all
+    // sharing bus 189 / device 0 / function 0, but each secondary
+    // display-output row carried a *distinct* instance path: the base
+    // path with an `&02`..`&05` suffix appended. Not blank, not merely
+    // case-differing, genuinely different -- which is the one shape
+    // `group_by_card` treats as a conflict and drops.
+    const REAL_8060S_BASE: &str = r"PCI\VEN_1002&DEV_1586&SUBSYS_B0261F4C&REV_C1\4&2368981F&0&0041";
+
+    #[test]
+    fn real_display_output_rows_do_not_dissolve_their_card() {
+        let mut adapters = vec![adapter(0, (189, 0, 0), REAL_8060S_BASE)];
+        let suffixed: Vec<String> = (2..=5).map(|n| format!("{REAL_8060S_BASE}&0{n}")).collect();
+        for (offset, pnp) in suffixed.iter().enumerate() {
+            adapters.push(adapter(offset as i32 + 1, (189, 0, 0), pnp));
+        }
+        let groups = group_by_card(&adapters);
+        assert_eq!(
+            groups.len(),
+            1,
+            "five ADL rows of one physical card must yield one group"
+        );
+        assert_eq!(groups[0].indices, vec![0, 1, 2, 3, 4]);
+        // The base path is the one WMI reports as PNPDeviceID, so it is
+        // the only value the exact join in `plan_attribution` can match.
+        assert_eq!(groups[0].pnp_string, REAL_8060S_BASE);
+    }
+
+    #[test]
+    fn the_base_path_wins_however_the_driver_enumerated_the_rows() {
+        // Display-output rows first, base path last. Grouping decides
+        // against the shortest path once the group is complete, so the
+        // verdict does not depend on enumeration order.
+        let second = format!("{REAL_8060S_BASE}&02");
+        let third = format!("{REAL_8060S_BASE}&03");
+        let adapters = vec![
+            adapter(1, (189, 0, 0), &second),
+            adapter(2, (189, 0, 0), &third),
+            adapter(0, (189, 0, 0), REAL_8060S_BASE),
+        ];
+        let groups = group_by_card(&adapters);
+        assert_eq!(groups.len(), 1);
+        assert_eq!(groups[0].indices, vec![0, 1, 2]);
+        assert_eq!(groups[0].pnp_string, REAL_8060S_BASE);
+    }
+
+    #[test]
+    fn a_shared_prefix_that_is_not_a_path_boundary_is_still_a_conflict() {
+        // `...&0&0041` and `...&0&00410` are different devices that
+        // happen to share a prefix. Only an extension at a `&` boundary
+        // is a display-output sibling, so these must not merge.
+        let lookalike = format!("{REAL_8060S_BASE}0");
+        let adapters = vec![
+            adapter(0, (189, 0, 0), REAL_8060S_BASE),
+            adapter(1, (189, 0, 0), &lookalike),
+        ];
+        assert!(group_by_card(&adapters).is_empty());
+    }
+
+    #[test]
+    fn display_output_rows_no_longer_block_multi_gpu_attribution() {
+        // The configuration this feature exists for, in the row shape
+        // real hardware produces: an APU and a dGPU, each enumerating a
+        // base row plus a display-output row. Before grouping keyed on
+        // the base path, both groups were dropped as conflicting and
+        // this returned `Decline`, which made the whole feature inert.
+        let apu_second = format!("{APU_PNP}&02");
+        let dgpu_second = format!("{DGPU_PNP}&02");
+        let adapters = vec![
+            adapter(0, (4, 0, 0), APU_PNP),
+            adapter(1, (4, 0, 0), &apu_second),
+            adapter(2, (8, 0, 0), DGPU_PNP),
+            adapter(3, (8, 0, 0), &dgpu_second),
+        ];
+        assert_eq!(
+            plan_attribution(&[DGPU_PNP, APU_PNP], Some(&adapters)),
+            AttributionPlan::PerCard(vec![
+                CardMatch {
+                    gpu_index: 0,
+                    adl_indices: vec![2, 3],
+                },
+                CardMatch {
+                    gpu_index: 1,
+                    adl_indices: vec![0, 1],
+                },
+            ])
+        );
     }
 
     #[test]
--- a/src/device/readers/amd_adl/adapters.rs
+++ b/src/device/readers/amd_adl/adapters.rs
@@ -146,38 +146,71 @@ fn plausible_bdf(adapter: &AdlAdapter) -> bool {
 /// Groups are sorted by bus/device/function and their indices sorted
 /// ascending, so the output is deterministic regardless of the order
 /// the driver enumerated in.
+///
+/// A group whose rows carry two different non-empty PNP instance paths
+/// is dropped rather than returned. Grouping trusts the PCI
+/// bus/device/function to identify a physical card, so a driver that
+/// leaves those fields unfilled collapses every card into one group.
+/// The caller then samples telemetry from the first index of the group
+/// that answers PMLog, which can be a different physical card's index,
+/// and one card's temperature, power, and fan would be reported for
+/// another GPU with nothing in the output saying so. Two different
+/// instance paths under one PCI address is the observable signature of
+/// that state, and dropping the group is the same answer the rest of
+/// this module gives everywhere else: decline rather than guess. Rows
+/// of one real card share their instance path; a secondary
+/// display-output row may leave it blank, which is backfilled and is
+/// not a conflict, and case differences are not a conflict either,
+/// since WMI and ADL do not agree on case.
 pub fn group_by_card(adapters: &[AdlAdapter]) -> Vec<CardGroup> {
     let mut groups: Vec<CardGroup> = Vec::new();
+    // Parallel to `groups`: whether the group saw contradictory PNP
+    // strings and therefore cannot be trusted to be one card.
+    let mut conflicting: Vec<bool> = Vec::new();
     for adapter in adapters {
         if !plausible_bdf(adapter) {
             continue;
         }
-        match groups.iter_mut().find(|group| {
+        let existing = groups.iter().position(|group| {
             group.bus == adapter.bus
                 && group.device == adapter.device
                 && group.function == adapter.function
-        }) {
-            Some(group) => {
+        });
+        match existing {
+            Some(position) => {
+                let group = &mut groups[position];
                 group.indices.push(adapter.index);
                 // Prefer any row that actually carries the string; a
                 // secondary display-output row can leave one blank.
-                if group.pnp_string.is_empty() && !adapter.pnp_string.is_empty() {
-                    group.pnp_string = adapter.pnp_string.clone();
+                if !adapter.pnp_string.is_empty() {
+                    if group.pnp_string.is_empty() {
+                        group.pnp_string = adapter.pnp_string.clone();
+                    } else if !group.pnp_string.eq_ignore_ascii_case(&adapter.pnp_string) {
+                        conflicting[position] = true;
+                    }
                 }
                 if group.adapter_name.is_empty() && !adapter.adapter_name.is_empty() {
                     group.adapter_name = adapter.adapter_name.clone();
                 }
             }
-            None => groups.push(CardGroup {
-                bus: adapter.bus,
-                device: adapter.device,
-                function: adapter.function,
-                adapter_name: adapter.adapter_name.clone(),
-                pnp_string: adapter.pnp_string.clone(),
-                indices: vec![adapter.index],
-            }),
+            None => {
+                groups.push(CardGroup {
+                    bus: adapter.bus,
+                    device: adapter.device,
+                    function: adapter.function,
+                    adapter_name: adapter.adapter_name.clone(),
+                    pnp_string: adapter.pnp_string.clone(),
+                    indices: vec![adapter.index],
+                });
+                conflicting.push(false);
+            }
         }
     }
+    let mut groups: Vec<CardGroup> = groups
+        .into_iter()
+        .zip(conflicting)
+        .filter_map(|(group, conflicted)| (!conflicted).then_some(group))
+        .collect();
     for group in &mut groups {
         group.indices.sort_unstable();
         group.indices.dedup();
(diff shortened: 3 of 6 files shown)


inureyes:

### Intent

Declare `AdapterInfo`, group ADL adapter indices per physical card, match cards to the reader's GPUs exactly, and relax attribution to multi-GPU AMD Windows hosts, with every failure path declining to the DXGI/PDH baseline.

### Layout transcription: independently re-derived, matches

Re-derived field by field from the Windows branch of AMD's `adl_structures.h` with `ADL_MAX_PATH = 256`. Every offset the PR asserts is correct, and so is the 1572-byte total: 9 `int`s (36 bytes) plus 6 `char[256]` arrays (1536 bytes), with every `int` after an array landing on a 4-aligned offset (264, 792, 1568), so C and `#[repr(C)]` agree and there is no padding. The issue's 1568 figure is wrong for the reason the PR gives. `iExist` at 796, `strDriverPath` at 800, and `strDriverPathExt` at 1056 are not asserted individually, but they are transitively pinned: `iPresent` at 792 and `strPNPString` at 1312 bracket exactly 516 bytes, which is `4 + 256 + 256`.

The declaration also survives a real Windows-target compile. Copying `ffi.rs`, `adapters.rs`, and `loader.rs` into a throwaway crate with a stub for `windows_gpu_perf::ids` and running `cargo check --target x86_64-pc-windows-gnu` passes, which type-checks `loader.rs` against the new `extern "C"` entry point and evaluates the size, alignment, and offset assertions under a Windows target's `int`. Flipping 1572 to 1568 fails that build, so the assertions are demonstrably live and not vacuous. The PR body and test plan have been updated, since they claimed the Windows-gated code could not be compile-checked at all.

### Findings addressed

- [x] **HIGH: `looks_sane` accepted a row the driver never wrote** (`src/device/readers/amd_adl/ffi.rs`). `AdapterInfoArray::for_count` pre-fills the buffer with zeroed rows, and every arm of the check was individually satisfied by that zero row: `iSize == 0` is explicitly accepted, `iAdapterIndex == 0` is in range, and `adl_string(&[0; 256])` returns `Some("")` because a NUL at offset 0 is documented as the empty string rather than garbage. So `validated()` returned `Some` for a buffer ADL never touched.

  This mattered because the runtime verification is the entire substitute for the hardware validation that cannot be run. Two consequences: the `amd.adl.adapters` doctor check, the mechanism by which a real host is supposed to confirm or refute the transcription, would report PASS over unwritten rows; and the documented guarantee that "a wrong stride is caught on the second entry even when the first parses" failed in the one direction where nothing garbles. If the real `sizeof(AdapterInfo)` is larger than 1572, ADL derives its row count as `iInputSize / sizeof`, writes fewer rows than requested, and leaves the tail blank, which is exactly the transcription error the check exists to catch. Fixed in 445900e by `AdapterInfo::is_blank`, with tests pinning both the blank row and the short-write table.

  Attribution itself was never at risk: a blank row groups to an empty PNP string, `parse_pnp_device_id("")` is `None`, so it can match no GPU. This restores the verification rather than fixing a mis-attribution.

- [x] **LOW: borrow ordering around the FFI write** (`src/device/readers/amd_adl/loader.rs`). `buffer.input_size()` took a shared borrow of the buffer after `buffer.as_mut_ptr()` produced the pointer ADL writes through. Sound in practice, since the reborrow covers the `Vec` header rather than the heap block, but hoisted into a local so the question does not arise.

### Verified, no finding

- **Memory safety of the call.** `count` is bounded to `1..=64` before allocation; `for_count(n)` allocates `2n + 4` rows and reports only `n` in `iInputSize`, so the advertised size is genuinely smaller than the allocation and a driver that ignores `iInputSize` lands inside it. `requested_entries()` slices `..n`, always within the allocation. `n * 1572` peaks at 100,608, far from `c_int` overflow. Rows are requested exactly as `ADL2_Adapter_NumberOfAdapters_Get` reported them. No `from_raw_parts` anywhere, and no read of uninitialized memory: the `Vec` is pre-filled, which is why the blank-row hole above was a verification weakness and not UB.
- **String handling cannot overrun.** `adl_string` and `adl_string_lossy` both bound themselves by the slice; a 256-byte array with no NUL returns `None` and renders 256 bytes respectively, never past the end.
- **Every row is checked**, not just the first, and a failed verification declines rather than proceeding with partial data: `validated()` is all-or-nothing, `adapter_inventory()` maps anything but `layout_ok: true` to `None`, and `plan_attribution` turns a `None` inventory into `Decline`.
- **Calling convention.** `Adl2AdapterAdapterInfoGet` is `extern "C"`, matching the file's rule that only `ADL_MAIN_MALLOC_CALLBACK` is `extern "system"`.
- **Non-Windows targets cannot reach the declaration.** The struct stays unconditional so its assertions run on the Linux runner, but the entry point type is `cfg(target_os = "windows")` and nothing outside that cfg calls it.
- **Matching logic.** `group_by_card` merges only on exact bus/device/function, so it cannot fuse two PCI functions; a card split across two groups would carry a duplicate PNP string, which phase 1 rejects as a multi-hit rather than guessing. `plan_attribution` enforces injectivity three ways: a uuid hitting several groups is no match, `has_duplicate_assignment` declines outright after each phase, and phase 2 snapshots both the unmatched-GPU set and the unclaimed-group set before the loop, requiring `peers == 1` and `candidates.len() == 1`. Two GPUs with distinct `PciIds` cannot select the same group because a group has exactly one identity, and two GPUs with the same identity both fail the `peers` test. The fallback requires the full vendor and device pair on both sides, and there is no ordinal or name arm anywhere. `parse_pnp_device_id` is reused from `windows_gpu_perf::ids` rather than reimplemented.
- **Single-GPU behavior is byte-identical**, verified against the diff rather than the test: `gpus.len() > 1` gates the inventory fetch entirely, `plan_attribution` returns `SoleGpu` on `len == 1` before touching `adapters`, and the resulting `loader::sample()` into `apply_to_gpu_info(&mut gpus[0], ...)` is the pre-PR body unchanged. Zero GPUs still no-ops.
- **Every failure path declines.** Missing entry point, failed call, implausible count, failed layout verification, no match, and ambiguous match all reach `Decline` or skip that one GPU, keeping the DXGI/PDH baseline.
- **Doctor check.** Prints the raw rows on PASS and on layout failure, and its rendering matches the expected output in the PR body exactly. `Severity::Info` matches its `amd.adl.library` and `amd.adl.sensors` siblings.
- **Integration.** Reached from `AmdWindowsGpuReader::get_gpu_info` via `amd_adl::augment`, and the doctor check is registered in the `amd.*` list. Nothing orphaned.

### Remaining items, not fixed

- **MEDIUM: `strDriverPath` and `strDriverPathExt` are not part of the string sanity check.** Four of the six string fields are checked. Offsets 800 and 1056 are only transitively pinned by `strPNPString` at 1312, so a tail reordering that preserved the total would be caught by `strPNPString` reading garbage rather than by those two directly. Left alone deliberately: the marginal coverage is small and every additional strictness arm is a new way to decline on hardware nobody has tested yet. Worth revisiting once a real host has produced a passing dump.
- **LOW: `group_by_card` silently keeps the first non-empty `pnp_string` when two rows share a bus/device/function but disagree.** Cannot happen if bus/device/function identity is sound, and a conflict would more likely surface as a multi-hit decline, so this is a note rather than a defect.

### The validation gap

Accurate now, and it was only overstated in the pessimistic direction. The layout is transcribed rather than observed, and no `AdapterInfo` has ever been filled in by a real driver here. What can be established without hardware has been: the transcription matches AMD's header on independent re-derivation, the assertions hold under a Windows target, the buffer accounting is sound, the string checks cannot overrun, and every failure path declines. What cannot: whether the installed driver's `AdapterInfo` actually is this struct. The remaining acceptance criterion from #353 is unchanged and still the gate.

### Verification

- [x] Tests pass (`cargo test --lib device::readers::amd_adl` 48 passed, `doctor` 25, `windows_gpu_perf` 31; `cargo clippy --lib --tests -- -D warnings` and `cargo fmt --check` clean)
- [ ] Confirmed on a real host with two or more AMD GPUs: still open, as intended

inureyes:

Scope: every `unsafe` block in `amd_adl/ffi.rs` and `amd_adl/loader.rs`, the row-count and stride arithmetic, the runtime self-verification helpers, resource handling across the 60-second inventory cache, and the per-refresh cost of the multi-GPU path. There is no untrusted-network input here, so this was read as a memory-safety review of unsafe FFI against a closed-source driver DLL on a target that CI never compiles.

### No CRITICAL or HIGH findings

The properties that matter most all hold, and each was checked rather than assumed:

- **The buffer is never smaller than the size advertised to the driver.** `input_size` is `requested * 1572`; the allocation is `(2 * requested + 4) * 1572`. Advertised is strictly smaller for every input, including the wrapping cases (truncation only engages above ~1.37M rows, where the allocation already exceeds 4 GB and the truncated value cannot reach it).
- **No uninitialized reads.** `for_count` zero-fills via `vec![AdapterInfo::default(); n]`, so every byte of every row is initialized before the call whether or not the driver writes it. `AdapterInfo` is 9 `c_int`s and 6 `[u8; 256]` arrays with no padding (pinned by the const assertions), so every bit pattern the driver can produce is a valid value: reading a partially written or untouched row is defined, not UB.
- **No reference is live across the FFI call.** `input_size` is hoisted into a local before `as_mut_ptr` (commit 445900e), and nothing touches the buffer again until `validated()` after the call returns.
- **The headroom is inside the pointer's provenance on the new path.** `Vec::as_mut_ptr` covers the whole allocation, so a driver ignoring `iInputSize` writes in bounds.
- **No panic reachable from arbitrary driver bytes.** `adl_string` stops at the first NUL and cannot read past 256 bytes; it rejects non-printable bytes before constructing a `str`, so `from_utf8` cannot fail and no `CStr` is built. `looks_sane`, `is_blank`, `parse_adapters`, `group_by_card`, `plan_attribution`, and `describe_raw_entry` are all total on arbitrary input. Every slice index in the new code is provably in range, including `gpus[card.gpu_index]`.
- **Row counts are bounded** to `1..=MAX_ADAPTER_ROWS` (64) before allocation, so the multiplications cannot overflow and a negative or zero count is a declined call.
- **No leak, double free, or use after free.** The context is created once and lives for the process, the library is held for the process, the cached inventory holds owned `String`s and ints rather than pointers, and the malloc callback is never invoked because this reader only calls caller-allocated entry points.
- **Performance is as claimed.** A single-GPU host does not fetch `AdapterInfo` at all (`gpus.len() > 1` gate, plus the `SoleGpu` early return), the inventory is fetched at most once per 60 s, and steady state is one PMLog call per matched card.

### Fixed in 6d2907a

| Severity | Location | Finding |
|---|---|---|
| MEDIUM | `amd_adl/adapters.rs` `group_by_card` | Rows sharing a PCI bus/device/function were merged even when their PNP instance paths disagreed. A driver leaving BDF unfilled collapses every card into one group, and `augment` samples the first index of a group that answers PMLog, so one card's temperature, power, and fan could be reported for another GPU with nothing in the output saying so. Such a group is now dropped, which declines instead of guessing. A blank sibling row is still backfilled and a case difference is still not a conflict. |
| MEDIUM | `amd_adl/loader.rs` `scan_for_capable_adapter` | The pre-existing capability scan looped over an unbounded driver-supplied count, one `ADL2_Overdrive_Caps` call and possibly one 8 KB-allocating PMLog read per index, while holding the process-wide runtime lock. A garbage count wedged the refresh loop. Now clamped to the same `MAX_ADAPTER_ROWS` the new path uses. |
| LOW | `amd_adl/loader.rs` `read_pmlog` | `&raw mut buffer.output` gave a pointer valid only for that field's 2052 bytes, putting the padded headroom out of bounds for exactly the oversized write it exists to absorb. Now derived from the whole box and narrowed by a cast. |
| LOW | `amd_adl/ffi.rs` `AdapterInfoArray::input_size` | Truncating `as c_int` cast guarded only by a doc comment. The invariant held for every reachable input, but this is the one number that would be a driver-side buffer overflow if it ever exceeded the allocation, so it is now a checked conversion that reports 0 rather than wrapping. |
| LOW | `doctor/checks/amd.rs` `amd.adl.adapters` | A failed verification always said "this is what a wrong transcribed struct layout looks like", but blank rows and garbled rows point at opposite corrections (declared struct larger than the driver's, versus wrong offsets or stride). The warning now names which one was seen. |

### Remaining, reported not fixed

- **MEDIUM. Blank-row rejection is all or nothing and may fire on healthy hardware.** AMD's own ADL sample memsets the adapter array before the call and then filters rows by `iPresent` / `ADL_Adapter_Active_Get`, which implies real drivers do leave some rows untouched. `looks_sane` rejecting any blank row would then disable multi-GPU attribution on a host where the layout is actually correct. Not changed, because tolerating blank rows reopens the short-write hole 445900e deliberately closed, and the failure is graceful (decline to the DXGI/PDH baseline). **This is worth checking explicitly during hardware validation:** if the doctor reports "the driver left N of M row(s) untouched" while the non-blank rows show legible device paths, the layout is right and this policy is what needs revisiting, not the struct.
- **LOW. Stale-inventory window.** The ADL index to card mapping is cached for 60 s while the GPU list is re-queried every refresh. If a driver restart or TDR renumbers adapter indices without changing the WMI GPU set, telemetry can land on the wrong GPU for up to a minute. Left alone: shortening the TTL pays a real cost for a rare window, and invalidating on GPU-set change does not cover renumbering.
- **LOW. Residual headroom assumption.** A driver that both ignores `iInputSize` and has a real row size above ~3144 bytes (2x the declared 1572) would still escape the allocation. Not realistic for the Windows layout, but it is the one overflow the padding does not cover.

### Verdict on soundness given that the target cannot be tested

The unsafe code is sound **conditional on the transcribed layout**, and it is structured so that a wrong layout is a wrong reading rather than memory corruption. That conditional is the right place for the risk to sit: `iInputSize` is passed honestly and is always below the allocation, the padding absorbs a driver that ignores it, the buffer is fully initialized so no row read is undefined regardless of what the driver did, and every downstream consumer is total on arbitrary bytes. A misdeclared layout therefore produces garbage that `looks_sane` rejects, and a rejection declines attribution rather than corrupting anything.

Independently of the code, the transcription itself matches AMD's public `adl_structures.h` field for field, including the `#if defined(_WIN32) || defined(_WIN64)` tail, and 1572 is the correct Windows size (the issue's 1568 drops `iAdapterIndex`).

What remains genuinely unverifiable without hardware is whether the installed driver's `sizeof(AdapterInfo)` agrees with the header, and that is exactly what the `amd.adl.adapters` doctor check is built to answer. The hardware-validation gate in the PR description should stay.

### Verification

`cargo check --lib --tests`, `cargo test --lib device::readers::amd_adl` (51 passed), `cargo test --lib doctor` (25 passed), `cargo clippy --lib --tests -- -D warnings`, `cargo fmt --check`, all clean. The Windows-gated changes in `loader.rs` and the doctor check were compile-checked for `x86_64-pc-windows-gnu` in a throwaway crate holding byte-identical copies of the three ADL modules plus a verbatim copy of `check_adl_adapters`, with negative controls confirming the check actually reaches the changed lines.

inureyes:
## PR Finalization

This PR still stays open awaiting validation on a real multi-AMD-GPU Windows host per the status section and issue #353's blocking acceptance criterion. Nothing below changes that; it only closes gaps in the code-review-only evidence this PR can produce on its own.

### Tests
`device::readers::amd_adl` went from 51 to 55 passing tests. The three previously-untested behaviors called out for this pass (the clamped `scan_for_capable_adapter` loop, the `probe_adapter_info` count-plausibility reject, and the doctor check's blank-vs-garbled failure message) were untestable in place because every caller lives behind `cfg(target_os = "windows")` and this repository has no Windows CI job. `MAX_ADAPTER_ROWS`, `plausible_adapter_count`, `clamp_scan_count`, and `describe_layout_failure` moved out of `loader.rs` / `doctor/checks/amd.rs` into the platform-independent `adapters.rs`, the same split already used for grouping and matching, with new tests for each. Blank-row rejection, the BDF-disagreement group drop, and the checked `input_size` conversion already had coverage and needed nothing further.

### Documentation
`README.md`'s doctor check-ID table listed no `amd.adl.*` checks at all; it now lists `amd.adl.library`, `amd.adl.sensors`, and `amd.adl.adapters`. No other doc references the pre-#353 single-AMD-GPU limitation this PR lifts, so there was nothing stale to correct there, and neither `API.md` nor the man page document `doctor` checks at all, which is a pre-existing gap outside this PR's scope.

### The two open, deliberately-unfixed items
Both are now recorded as doc comments on `AdapterInfo::is_blank` and `AdapterInfo::looks_sane` in `ffi.rs`, and in a new "Known gaps in the verification" section in the PR body above:
- Blank-row rejection may be too strict for a driver that legitimately leaves some rows untouched (AMD's own ADL SDK sample filters by `iPresent` / `ADL2_Adapter_Active_Get` rather than assuming every row is populated). Not relaxed here; the specific dump shape that would justify revisiting it (blank rows interleaved with legible ones, not a run of blank rows past some index) is written down for whoever validates on hardware.
- `looks_sane` checks only 4 of the struct's 6 string fields; `strDriverPath` and `strDriverPathExt` are transitively pinned by `strPNPString` passing, so the marginal coverage gain is small and left alone pending hardware evidence.

### Lint/format
`cargo clippy --lib --tests -- -D warnings` and `cargo fmt -- --check` both clean.

### Validation
- `cargo check --lib --tests`
- `cargo clippy --lib --tests -- -D warnings`
- `cargo fmt -- --check`
- `cargo test --lib device::readers::amd_adl` (55 passed)
- `cargo test --lib doctor` (25 passed)
- `cargo test --lib device::readers::windows_gpu_perf` (31 passed)
- Windows-target compile check of `ffi.rs`/`adapters.rs`/`loader.rs` in a throwaway crate via `cargo check --target x86_64-pc-windows-gnu`, plus a negative control confirming the 1572-byte `AdapterInfo` assertion still fails the build when flipped to 1568

Still waiting on the real-hardware doctor dump per the status section; nothing here substitutes for it.

inureyes:
Label state: `status:blocked`, not `status:done`.

Implementation, implementation review, security review, and finalization are all complete and green (1608 tests passing, clippy and fmt clean, three review findings found and fixed). What remains is not review work. This PR is blocked on the one acceptance criterion that cannot be satisfied from a Linux development host or from CI: confirmation on a real Windows host with two or more AMD GPUs, with the `amd.adl.adapters` doctor output attached here.

`status:done` in this project means "implementation complete, ready to merge". That is not true here and labelling it so would invite exactly the merge that #353 warns against. `status:review` would be equally misleading in the other direction, since it implies review is still outstanding when it is finished.

To unblock: run `all-smi doctor --only amd.adl` on a multi-AMD-GPU Windows host and attach the output. See the top of the PR body for what a correct dump looks like and what a failing one tells us.

inureyes:
## Hardware validation: layout CONFIRMED, and a real-hardware bug found

Ran this branch on a real Windows host with an AMD driver. The host has **one** AMD GPU, not two, so it does not close the outstanding acceptance criterion from #353. It does, however, retire this PR's central risk — the blind-transcribed `AdapterInfo` layout — because one physical card exposes one ADL adapter row per display output, and this card produced five. It also surfaced a bug that would have made the multi-GPU feature dead on arrival.

### Test host

| | |
|---|---|
| GPU | AMD Radeon(TM) 8060S Graphics (Strix Halo APU, `DEV_1586`), 1 card |
| Driver | 32.0.31035.1003 |
| OS | Windows 11 Pro 26200 |
| Toolchain | rustc 1.97.1, **`x86_64-pc-windows-msvc` (native, not cross)** |

### 1. `AdapterInfo` layout: CONFIRMED

`all-smi doctor --only amd.adl`:

```
PASS amd.adl.adapters   5 adapter row(s), 5 populated, across 0 physical card(s);
  [0] index=0 bus=189 device=0 function=0 vendor=1002 name="AMD Radeon(TM) 8060S Graphics" pnp="PCI\VEN_1002&DEV_1586&SUBSYS_B0261F4C&REV_C1\4&2368981F&0&0041"
  [1] index=1 bus=189 device=0 function=0 vendor=1002 name="AMD Radeon(TM) 8060S Graphics" pnp="PCI\VEN_1002&DEV_1586&SUBSYS_B0261F4C&REV_C1\4&2368981F&0&0041&02"
  [2] index=2 bus=189 device=0 function=0 vendor=1002 name="AMD Radeon(TM) 8060S Graphics" pnp="PCI\VEN_1002&DEV_1586&SUBSYS_B0261F4C&REV_C1\4&2368981F&0&0041&03"
  [3] index=3 bus=189 device=0 function=0 vendor=1002 name="AMD Radeon(TM) 8060S Graphics" pnp="PCI\VEN_1002&DEV_1586&SUBSYS_B0261F4C&REV_C1\4&2368981F&0&0041&04"
  [4] index=4 bus=189 device=0 function=0 vendor=1002 name="AMD Radeon(TM) 8060S Graphics" pnp="PCI\VEN_1002&DEV_1586&SUBSYS_B0261F4C&REV_C1\4&2368981F&0&0041&05"
PASS amd.adl.library    C:\Windows\System32\atiadlxx.dll loaded, PMLog-capable adapter index 0
PASS amd.adl.sensors    16 supported sensor(s); interpreted: power=Some(21.0)W gfx=Some(739)MHz mclk=Some(1000)MHz activity=Some(2.0)%; raw: 1=739 2=1000 3=653 7=12 19=2 23=21 28=40 29=41 30=1 32=0 33=0 34=2532 38=0 39=0 71=61 72=0

Summary: 3 pass, 0 warn, 0 fail, 0 skipped
```

Reading the requested row-state tags: **all five rows are `POPULATED`. No `GARBLED`, no `UNTOUCHED`, no `BLANK`.** The driver wrote every row we asked for, so this dump exercises the layout rather than the short-write paths.

What that pins down against a real driver:

- `strAdapterName` (offset 280) reads `"AMD Radeon(TM) 8060S Graphics"`, which is byte-identical to the WMI `Name`.
- `strPNPString` (offset 1312) on row `[0]` is byte-identical to the WMI `PNPDeviceID`. The exact join this PR is built on therefore does resolve on real hardware.
- `iAdapterIndex` (4), `iBusNumber` (264), `iDeviceNumber` (268), `iFunctionNumber` (272), `iVendorID` (276) all read correctly — `vendor=1002`, and a consistent bus 189 / device 0 / function 0 across all five rows.
- Legible strings at rows `[1]`–`[4]` mean the **1572-byte stride is right**. A wrong stride would garble rows past the first, which is exactly what `classify` would have reported. **The issue body's 1568 is wrong; this PR's 1572 is correct.**

Two things strengthen this beyond the PR's original cross-compile check:

- The build was **native MSVC**, so the `const _: () = assert!(size_of::<AdapterInfo>() == 1572)` assertions and all ten offset assertions were evaluated under the actual target ABI, not a mingw stand-in.
- `cargo build --release` on the **whole workspace** succeeded (exit 0). That is the test-plan item previously unchecked because `zstd-sys` needed `x86_64-w64-mingw32-gcc`; it is not an issue for a native build.

`cargo test --release --lib device::readers::amd_adl` → **61 passed** natively on Windows.

### 2. Single-GPU regression path: intact

`all-smi snapshot` on this host attributes ADL telemetry correctly, and the `SoleGpu` arm means `AdapterInfo` is never consulted:

```json
"name": "AMD Radeon(TM) 8060S Graphics",
"temperature": 38, "power_consumption": 17.0, "frequency": 773, "utilization": 3.0,
"uuid": "PCI\VEN_1002&DEV_1586&SUBSYS_B0261F4C&REV_C1\4&2368981F&0&0041",
"detail": {
  "Source: Temperature": "ADL (gfx)", "Source: Power": "ADL",
  "Source: Utilization": "ADL", "Source: Frequency": "ADL",
  "Metrics Source": "WMI + DXGI + PDH + ADL"
}
```

Note the GPU `uuid` equals row `[0]`'s `strPNPString` exactly.

### 3. BUG: `group_by_card` drops the only card

Look again at the header line: **`5 adapter row(s), 5 populated, across 0 physical card(s)`**. Five valid rows, zero groups.

`group_by_card` marks a BDF group conflicting when two rows carry different non-empty PNP strings (`adapters.rs:283`) and drops it (`adapters.rs:307`). The assumption is stated in its doc comment:

> Rows of one real card share their instance path; a secondary display-output row may leave it blank

**Real hardware refutes this.** The secondary display-output rows are not blank and do not merely differ in case — they carry a genuinely different instance path: the base path with `&02`…`&05` appended. That is precisely the one shape the guard treats as evidence of two cards collapsed under one PCI address, so the group is discarded.

On this host the impact is nil, because one GPU takes the `SoleGpu` arm before the inventory is consulted. **On the two-AMD-GPU host this PR exists to serve, every group would be dropped, `group_by_card` would return empty, and `plan_attribution` would hit `if groups.is_empty() { return Decline }` — permanently.** The feature would never engage on real hardware, and no test in the suite catches it because every grouping test repeats one identical PNP string per card.

Reproduced deterministically with the observed strings:

```
test device::readers::amd_adl::adapters::tests::real_display_output_rows_do_not_dissolve_their_card ... FAILED
assertion `left == right` failed: five ADL rows of one physical card must yield one group
  left: 0
 right: 1
```

### Fix being prepared

Narrow the conflict test from "not equal" to "neither is a prefix of the other" (case-insensitive), and let a group's canonical `pnp_string` be the shortest member — the base path, which is the only value WMI's `PNPDeviceID` can match.

This preserves the guard's intent. Two genuinely distinct cards diverge at the `DEV_` field, so `APU_PNP` vs `DGPU_PNP` is still a conflict and is still dropped; display-output siblings are strict extensions of their base and now group correctly. All existing grouping tests plus the new regression test are satisfied.

### Acceptance criteria status

- [x] `AdapterInfo` layout (size 1572, stride, and every load-bearing offset) verified against a real AMD driver
- [x] Whole-workspace Windows compile — native MSVC, exit 0
- [x] Single-GPU behavior unchanged, with live ADL telemetry
- [ ] **Still open:** confirmation on a host with two or more AMD GPUs. This host has one. The per-card join, the `PerCard` plan, and distinct per-GPU temperature/power/fan/clocks remain unexercised on hardware.

inureyes:
## Fix pushed: `0bebb9a` — grouping keyed on the base device instance path

Follow-up to the hardware validation above, verified on the same host.

### The change

`group_by_card` judged two rows to be the same card by string equality of `strPNPString`. Real ADL never repeats the path, so sameness is now **"one path extends the other at a `&` boundary"**, via a new `shares_device_instance` helper.

Two further points made it correct rather than merely permissive:

- **The `&` boundary is load-bearing.** A bare prefix test would merge `...&0&0041` with `...&0&00410`, which are different devices. Requiring the remainder to begin with a separator keeps them distinct.
- **Agreement is judged once per group, against the shortest path its rows carried, rather than pairwise as rows arrive.** Pairwise-on-arrival would still fail if the driver enumerated a display-output row before the base one: `...&02` and `...&03` are not prefixes of each other, so the group would dissolve again. Deciding at the end against the shortest path makes the verdict independent of enumeration order. The shortest path is the base path, and therefore the group's identity — it is the only form a WMI `PNPDeviceID` can match.

The original guard's intent is preserved. Two genuinely distinct cards diverge at the `DEV_` field, so `APU_PNP` vs `DGPU_PNP` under one PCI address is still a conflict and is still dropped — `a_bdf_group_with_contradictory_pnp_strings_is_dropped_rather_than_merged` passes unchanged. A group whose rows all left the path blank still survives with an empty identity, exactly as before.

### Verified on hardware

Same host, rebuilt:

```
PASS amd.adl.adapters   5 adapter row(s), 5 populated, across 1 physical card(s); ...
```

`across 0 physical card(s)` → **`across 1 physical card(s)`**. The single-GPU path is unaffected and still reports live ADL telemetry (`temperature: 40`, `power_consumption: 28.0`, `frequency: 696`, `Source: Temperature/Power/Utilization/Frequency = ADL`).

### Tests

`cargo test --lib device::readers::amd_adl` → **65 passed** (was 61), all pre-existing grouping and attribution tests unchanged. Four added:

| Test | What it pins |
|---|---|
| `real_display_output_rows_do_not_dissolve_their_card` | The five observed rows, verbatim, group into one card |
| `the_base_path_wins_however_the_driver_enumerated_the_rows` | Base row last still yields the base path as identity |
| `a_shared_prefix_that_is_not_a_path_boundary_is_still_a_conflict` | `...&0041` vs `...&00410` do not merge |
| `display_output_rows_no_longer_block_multi_gpu_attribution` | APU + dGPU, each with a display-output row, yields `PerCard` instead of `Decline` |

The last one is the regression that matters: on the pre-fix code it returned `Decline`, which is the whole feature going inert.

### Also in this commit

One `format!` argument inlined in the `amd.adl.adapters` check, which `-D warnings` rejects. Worth noting *why* it was missed: that check is `cfg(target_os = "windows")`, so clippy on the Linux runner never linted it. `cargo clippy --lib --tests -- -D warnings` is now clean for every file this PR touches.

### Notes for whoever runs CI on Windows

Independent of this PR, on a native MSVC build (rustc 1.97.1):

- Four `--lib` tests fail on Windows in files this PR does not touch, all pre-existing on `main`: `device::common::validation::tests::test_validate_command_path` (asserts `/bin/ls` is absolute), `metrics::energy_wal::tests::expand_tilde_replaces_home_prefix` and `::torn_final_record_is_discarded` (tilde/HOME assumptions), and `doctor::bundle::tests::bundle_writes_expected_entries` (`NamedTempFile` re-open hits Windows file locking, os error 32).
- Three clippy `-D warnings` failures also pre-exist on `main`, all in `cfg(windows)` code: `utils/command_timeout.rs:190`, `device/cpu_windows.rs:49` and `:63`, `device/windows_temp/amd_ryzen.rs:228`.

Both sets are the same blind spot this bug came from — Linux CI cannot see `cfg(windows)` code. Probably worth their own issue.

### Acceptance criteria

- [x] `AdapterInfo` layout verified against a real AMD driver
- [x] Whole-workspace native Windows build, `cargo fmt`, and clippy clean for PR-touched files
- [x] Single-GPU behavior unchanged, live ADL telemetry
- [x] Grouping produces the correct card count on real hardware
- [ ] **Still open:** a host with two or more AMD GPUs. The per-card join and distinct per-GPU telemetry are now covered by unit tests in the real row shape, but remain unexercised on hardware.

inureyes:
## Per-index PMLog probe: the Windows-only sampling glue works on real hardware

The two reports above validated the `AdapterInfo` layout and the grouping logic. Both live in `adapters.rs`, which is `cfg`-free and therefore unit-testable. This report covers the layer neither tests nor the earlier runs could reach.

### The gap this closes

Three things in the multi-GPU path are `cfg(target_os = "windows")` **and** unreachable on a single-GPU host, because `plan_attribution` takes its `SoleGpu` arm before any of them run:

- `loader::adapter_inventory()` — the 60-second cached inventory, called by `augment` only when `gpus.len() > 1`
- `loader::sample_adapter(index)` — PMLog against an explicit adapter index, rather than the scanned index `sample()` uses
- the `PerCard` arm of `augment`, whose `.find_map(|&index| loader::sample_adapter(index))` takes the first index of a card that answers

No unit test reaches any of them (`augment_is_inert_when_attribution_is_refused` exercises only the `Decline` arm), and nothing in the CLI calls them on this host. So they were shipping on inspection alone.

This host cannot supply two cards, but it does supply **five ADL indices belonging to one card**, which is enough to exercise the mechanism. I drove `adapter_inventory()` and then `sample_adapter()` against every index it returned, via a temporary probe (since reverted — no code in this PR changed as a result).

### Result

`selected_adapter_index()` = `Some(0)`, `adapter_inventory()` = 5 rows. Every index answered:

| ADL index | PNP suffix | sensors | gfx °C | power W | clk_gfx | clk_mem | activity |
|---|---|---|---|---|---|---|---|
| 0 | *(base)* | 16 | 40 | 32.0 | 710 | 1000 | 1.0 |
| 1 | `&02` | 16 | 39 | 32.0 | 710 | 1000 | 1.0 |
| 2 | `&03` | 16 | 39 | 32.0 | 709 | 1000 | 1.0 |
| 3 | `&04` | 16 | 40 | 32.0 | 709 | 1000 | 1.0 |
| 4 | `&05` | 16 | 40 | 32.0 | 709 | 1000 | 1.0 |

Raw, index 0 and index 4 for comparison:

```
[0] 1=710 2=1000 3=642 7=12 19=1 23=32 28=40 29=40 30=1 32=0 33=0 34=2726 38=0 39=0 71=0 72=0
[4] 1=709 2=1000 3=641 7=12 19=1 23=32 28=40 29=40 30=1 32=0 33=0 34=2732 38=0 39=0 71=0 72=0
```

### What this establishes

1. **`adapter_inventory()` works against a real driver.** Its first ever execution; it returned the same five rows the doctor check reports.

2. **`sample_adapter()` works for every index, not just the scanned one.** All five returned a full 16-sensor table — no `None`, no short table. The explicit-index path carries no off-by-one and no context misuse.

   This matters more than it first looks. Because `augment` takes the *first index that answers*, a card whose secondary display-output indices stayed silent would make that selection quietly order-dependent. They all answer, so it is not.

3. **The comment on the `PerCard` arm is correct:** *"Every index of a card reports the same telemetry."* Power is bit-identical at 32.0 W across all five, as are `clk_mem` (1000) and activity (1.0).

   The small spread elsewhere — gfx 40/39/39/40/40, `clk_gfx` 710/710/709/709/709, sensor 34 at 2726 then 2732 — is live sensor jitter between sequential reads on an active APU, not a per-index difference. The magnitude is ±1 unit, and identical power across all five reads confirms a single underlying source.

### What is still unverified

Whether **different cards** yield **different** telemetry. Every index here belongs to one physical card, so per-card differentiation cannot be observed on this host at all.

The nature of the remaining risk has changed, though. Before: the sampling mechanism was entirely unexercised. Now: the mechanism demonstrably works, and the single open question is whether the driver separates readings per card — which is exactly what the outstanding two-AMD-GPU acceptance criterion asks.

### Follow-up

Making this probe a permanent `amd.adl.per_adapter` doctor check would let whoever has a two-GPU host answer that question with one command, instead of reading the source and writing a probe as I did. That is the same field-verification role `amd.adl.sensors` and `amd.adl.adapters` already play.

Filed as #369 rather than added here, to keep this PR's scope where it is. It would also put `adapter_inventory()`, `sample_adapter()`, and the `PerCard` arm on a reachable code path, which is worth having regardless of how the two-GPU question resolves.

inureyes:
## Merging: what changed about the decision, and what did not

This PR opened with an explicit "do not merge on cross-compile or code-review evidence alone" header. That condition has been met, so this records why it is being merged now rather than quietly dropping the header.

### What retired the blocker

The blocker was never the two-GPU count by itself. It was that the `AdapterInfo` layout had been transcribed blind from AMD's public headers, with no execution against a real driver anywhere in the loop. The 2026-08-18 run closed exactly that gap:

- 1572-byte size, stride across five rows, and every load-bearing offset evaluated under the real `x86_64-pc-windows-msvc` ABI and confirmed against driver 32.0.31035.1003.
- `strPNPString` byte-identical to the WMI `PNPDeviceID`, so the join the feature rests on resolves on real hardware.
- All five rows `POPULATED`. No `GARBLED`, no `UNTOUCHED`, no `BLANK`.
- `adapter_inventory()` and `sample_adapter(index)` driven against every index, all returning a full 16-sensor table.
- A real defect found and fixed (`0bebb9a`): `group_by_card` dissolved the only card, which would have made `plan_attribution` return `Decline` permanently on a two-card host. That bug was found by running the code, not by reading it, which is the strongest argument that the hardware pass did its job.

### What is still open, and why it does not block

Whether two *different physical* cards yield *different* telemetry. That is an observation about driver behaviour, not a question about this code, and it is unobtainable on any single-GPU host because `plan_attribution` takes `SoleGpu` before the inventory is consulted.

The risk of merging without it is bounded by design: verification failure, a missing entry point, a failed call, and a failed match all land on the same `Decline` arm, which keeps the DXGI/PDH baseline. The single-GPU path never consults `AdapterInfo` at all and is pinned by a regression test. The realistic worst case on an unverified two-GPU host is that attribution declines and the user sees today's behaviour, not that telemetry is attributed to the wrong card.

Against that, keeping validated code parked outside `main` has a real cost: it blocks #369, which adds the `amd.adl.per_adapter` doctor check that makes this very verification a one-command operation instead of a hand-written probe.

### Where the remaining work lives

- **#370** (`type:test`, `status:ready`): confirm per-card attribution on a Windows host with two or more AMD GPUs. Carries the procedure, the expected output shape, and everything already established, so whoever gets the hardware does not have to reconstruct it. Same split pattern as #354 through #357 out of #309/#310/#311.
- **#369**: the `amd.adl.per_adapter` doctor check, starting next off `main`.

#353 closes with this merge, since what is left there is field verification rather than unwritten code. #370 asks the verifier to post the confirmed dump back to #353 as a comment, so the implementation issue ends up carrying its own evidence.

### Pre-merge state

`cargo test --lib device::readers::amd_adl` on this Windows host: 65 passed, 0 failed. All required CI checks green. `MERGEABLE` / `CLEAN`.

PR #382, opened 2026-08-23
feat(ci): drive the Level Zero backend against a stub loader

### LINKED ISSUE #379: feat: stub the Level Zero loader in CI so the Sysman paths run without hardware
## Problem / Background

PR #365 (merged as `dd64e21`) made the Intel Level Zero backend unconditional for every Linux and Windows target via the `all_smi_level_zero` cfg emitted by `build.rs`, and added the first CI check that touches the real loader. `.github/workflows/ci.yml:56-73` installs the Ubuntu `libze1` package on the Linux `test` job, and `.github/workflows/ci.yml:76-88` runs `the_installed_loader_exports_every_symbol_we_resolve` (`src/device/readers/intel_gpu_level_zero/tests.rs:418-436`) with `ALL_SMI_EXPECT_LEVEL_ZERO_LOADER=1`, then greps the output for the marker `all-smi: level-zero-loader-assertion-ran` so a test that quietly stops asserting turns the job red instead of staying green.

That check stops at `dlopen` plus symbol resolution. It deliberately does not call `zeInit`, and that is the right call: a GitHub-hosted runner has no Intel GPU, so the loader has no driver to hand back, and initialisation failing there is correct behaviour rather than a defect. Asserting on it would make the test fail for the wrong reason.

The consequence is that everything past `try_load_library` is unexercised by any runner this project has. Nothing checks `initialize_runtime` (`src/device/readers/intel_gpu_level_zero/loader.rs:191-247`), including its `zesInit` versus `ZES_ENABLE_SYSMAN=1` branch and its `zeInit` result handling. Nothing checks `enumerate_devices` (`loader.rs:256-315`): the count-then-buffer idiom, the `cap_handle_count` truncation at `loader.rs:50`, the null-handle skips, the `zes_pci_properties_t` read, or the BDF mapping through `format_pci_bdf` (`loader.rs:325`) into `devices_by_pci`. Nothing checks the delta arithmetic and metric refresh reached through `refresh` (`src/device/readers/intel_gpu_level_zero.rs:200`): `populate_engine_samples` and `populate_power_samples` (`refresh.rs:71`, `refresh.rs:126`), `refresh_engines` and `refresh_power` (`refresh.rs:171`, `refresh.rs:225`), the point-in-time families in `point.rs:95-330`, or `apply_to_gpu_info` (`apply.rs:25`).

The FFI struct layouts in `src/device/readers/intel_gpu_level_zero/ffi.rs:112-220` and `ffi/sysman.rs:45-200` are the sharpest edge. The coverage that exists today is size-only: `tests.rs:97-137` and `tests/ffi_layout.rs:42-51` assert `size_of` against numbers transcribed by hand from the Sysman spec, checking our own struct against our own transcription. That catches a mis-sized field, and the class is real: `zes_pci_properties_t` once declared three `ze_bool_t` fields as `u32`, inflating the struct from 56 to 64 bytes and leaving the trailing bytes of a driver-written buffer reading garbage. It does not catch two same-typed fields written in the wrong order, a wrong offset that happens to preserve the total size, or a spec value transcribed wrong in both the struct and its assertion. All three are invisible to the compiler and surface only on real hardware.

## Current Behavior

On the Linux `test` job, three steps touch Level Zero. `ci.yml:56-73` installs `libze1` (falling back to `level-zero`) and fails the job if `libze_loader.so.1` does not appear in `ldconfig -p`. `ci.yml:76-88` runs the single symbol-resolution test under the env gate and greps for its marker. `ci.yml:94-95` runs the full `cargo test`, which now compiles the backend on every Linux target and so includes its pure-logic tests: enum value locks, BDF formatting, size assertions, and the delta arithmetic helpers called with hand-supplied numbers rather than driver-supplied ones.

No step ever obtains an `LzRuntime`. `initialize_runtime` returns `None` on any runner, so `with_runtime` (`loader.rs:184`) short-circuits and `refresh` returns `None`. Two existing tests assert exactly that absence: `enumerated_pci_bdfs_empty_when_runtime_absent` (`tests.rs:439`) and `refresh_returns_none_without_runtime` (`tests.rs:453`).

## Proposed Solution

Build a stub Level Zero loader in CI, bind the backend to it on the Linux runner, and drive the real backend code against it. Six decisions, already made:

**1. Compile the stub against the vendor headers, not against our transcription.** Install the Level Zero development headers alongside the runtime (`libze-dev` on Ubuntu 24.04, `level-zero-dev` on images served by Intel's repository). If neither package resolves, vendor a pinned copy of `ze_api.h` and `zes_api.h` under `tests/fixtures/level_zero/include/` and record the upstream tag in a comment. The stub is then C code filling the vendor's own struct definitions. A value that survives the round trip into a Rust field proves the `#[repr(C)]` layout agrees with the real header field by field, which is strictly stronger than the size assertions in `tests/ffi_layout.rs` and is the point of the whole exercise.

**2. Bind it through `LD_LIBRARY_PATH`, not through a runtime override in shipped code.** `LIBZE_PATHS[0]` is the bare SONAME `libze_loader.so.1` (`loader.rs:72-81`), and `dlopen` on a bare SONAME searches `LD_LIBRARY_PATH` before the default paths, so a stub named `libze_loader.so.1` in a directory placed first on `LD_LIBRARY_PATH` wins over the real `/usr/lib/x86_64-linux-gnu/libze_loader.so.1` that `ci.yml:56-73` installs. This needs no new branch in `loader.rs`, no test-only environment override, and no `cfg(test)` arm in the production loader, which is what keeps the stub out of reach of a shipped build. The rejected alternative is adding an `ALL_SMI_LEVEL_ZERO_LIBRARY` path override to `LIBZE_PATHS`: it would put a "load this arbitrary shared object" switch into every released binary in exchange for nothing the `LD_LIBRARY_PATH` route does not already provide.

**3. Run the stub-driven tests in their own process.** `LZ_RUNTIME` is a process-wide `OnceCell` (`loader.rs:137`) latched by the first caller of `ensure_runtime` (`loader.rs:177`), and two existing unit tests in the same binary already latch it (`tests.rs:439` and `tests.rs:453` both reach `with_runtime`). Whichever test runs first decides which loader the entire test binary sees, so a stub test living in `src/.../tests.rs` would be racy. Put the new tests in their own integration file, `tests/level_zero_stub.rs`, which Cargo compiles into a separate binary and therefore a separate process, and drive them through the public surface the library already exports: `all_smi::device::readers::intel_gpu_level_zero::{LevelZeroState, LevelZeroReadout, refresh, enumerated_pci_bdfs, apply_to_gpu_info, ApplyPlatform}`, reachable via `src/lib.rs:119` then `src/device/mod.rs:63` then `src/device/readers/mod.rs:85-86`.

**4. Export exactly the symbols the backend resolves.** The four mandatory ones from `try_load_library` (`api.rs:59-108`): `zeInit`, `zeDriverGet`, `zeDeviceGet`, `zesDevicePciGetProperties`. Plus `zesInit`, so `initialize_runtime` takes the modern Sysman route at `loader.rs:230-238` rather than the legacy environment-variable route. Plus the optional Sysman families listed in `LzApi` (`api.rs:23-46`): engine (`zesDeviceEnumEngineGroups`, `zesEngineGetProperties`, `zesEngineGetActivity`), power (`zesDeviceEnumPowerDomains`, `zesPowerGetEnergyCounter`), temperature, memory, frequency, and fan. Every enumerator implements the same count-then-buffer contract the real API has (non-null count pointer with a null buffer means count only), because that contract is precisely what `enumerate_devices` and the three `populate_*` helpers depend on.

**5. Return synthetic values chosen to be individually identifiable.** One fake driver with two fake devices at known addresses, for example `0000:03:00.0` and `0000:af:00.0`, so `format_pci_bdf` (`loader.rs:325`) has a non-trivial domain, bus, device, and function to render and `enumerated_pci_bdfs` (`intel_gpu_level_zero.rs:355`) has a sort order to get right. Give every numeric field a value unique within its struct, so a field read at the wrong offset yields an obviously wrong number rather than a plausible one.

**6. Advance the delta counters between calls.** Engine activity and power are delta-based (`refresh.rs:171-305`) and a first sample is only seeded, so a single refresh proves nothing. Have the stub advance `active_time`, `energy`, and `timestamp` by a fixed step per call, making the second `refresh` produce an exact constant that `compute_engine_busy_pct` (`refresh.rs:256`) and `compute_power_watts` (`refresh.rs:281`) can be asserted against end to end instead of only through their already-tested pure arithmetic.

The repository has precedent for this technique. PR #376 type-checked the Windows-only `intel_gpu_windows.rs` through a scratch worktree with a stubbed `wmi` crate, so the real module compiled against the real `GpuInfo` and the real classifier without a Windows host. The method and its most important rule are written down at `src/service_cmd/mod.rs:105-135`: prove the harness actually reaches the code before trusting a clean run, by injecting a deliberate error and confirming it fails. PR #376 did exactly that for its probe. Do the same here.

## Scope

**In scope:** the stub source and its build (`tests/fixtures/level_zero/stub_ze_loader.c` plus a `cc -shared -fPIC -Wl,-soname,libze_loader.so.1` step, or a `cdylib` helper crate if the implementer prefers Rust, though C against the vendor headers is preferred for the reason in decision 1); a new integration test `tests/level_zero_stub.rs` driving the public API; a new step on the Linux `test` job in `.github/workflows/ci.yml`, placed after the existing loader steps at lines 56-88, that builds the stub and runs the new test with `LD_LIBRARY_PATH` pointed at it, with a marker grep in the same shape as `ci.yml:81-88`; and a coverage comment at the top of the stub source.

**Out of scope:** Windows. A `ze_loader.dll` stub would need an MSVC build on the `windows-service` job (`ci.yml:925`), which cannot run `cargo test` or `clippy` today for the reasons recorded in #366, #367, and #368. Linux first. Also out of scope: doctor-level reporting of the Level Zero runtime state, which is #380, and anything that genuinely requires an Intel GPU.

## Implementation Notes

- **Reuse**: the fail-loudly pattern from #365 is the reference and should not be reinvented. It is an env key that turns the test from a no-op into an assertion (`tests.rs:390`), a marker string printed only after the assertion passes (`tests.rs:400`, `tests.rs:435`), and a workflow `grep` that fails the job when the marker is absent (`ci.yml:84-87`). Use the same three-part shape with its own key and its own marker.
- **Reuse**: drive the assertions through the existing public entry points (`refresh`, `enumerated_pci_bdfs`, `apply_to_gpu_info`) rather than adding new test-only accessors. `engine_count` (`intel_gpu_level_zero.rs:343`), `power_domain_count` (`:366`), and `is_bound` (`:372`) already exist as diagnostics and cover most of what a test would otherwise want to reach into.
- **Constraints**: no change to `LIBZE_PATHS` or to the production loader search order. No new runtime dependency and no new default-on cargo feature. Both loaders will be present on the runner, since `ci.yml:56-73` installs the real one, so only `LD_LIBRARY_PATH` ordering separates them: build the stub into a dedicated directory and put that directory first.
- **Edge cases**: every enumerator must handle the count-only call, because `enumerate_devices` (`loader.rs:262-303`) and all three `populate_*` helpers call twice. A driver that writes back a smaller count on the fill call is handled by the `truncate` calls at `loader.rs:280` and `loader.rs:302`, so include one stub device that does this. `cap_handle_count` (`loader.rs:50`) clamps anything above `MAX_L0_HANDLES = 256`, so one enumerator reporting an over-cap count is worth a test. A BDF that matches no stub device must still yield no data, which is the contract `refresh_returns_none_without_runtime` (`tests.rs:453`) already documents.
- **Error handling**: a Sysman family for which the stub returns a non-success code must degrade only that field, leaving the rest of the readout fresh. That per-family independence is the documented contract in the module header (`intel_gpu_level_zero.rs:30-35`) and deserves one negative test.
- **Determinism**: no clock reads and no randomness in the stub. Timestamps advance by a fixed step per call so the derived percentages and watts are exact constants in the assertions, not ranges.

## Acceptance Criteria

- [ ] A stub Level Zero loader lives in the repository under a test-only path and exports at minimum `zeInit`, `zesInit`, `zeDriverGet`, `zeDeviceGet`, and `zesDevicePciGetProperties`, plus the engine, power, temperature, memory, frequency, and fan Sysman symbols named in `LzApi` (`src/device/readers/intel_gpu_level_zero/api.rs:23-46`).
- [ ] The stub is wired into `.github/workflows/ci.yml` on the Linux `test` job and actually runs there. Committing the stub without a workflow step that builds and uses it does not satisfy this.
- [ ] With the stub bound, `enumerated_pci_bdfs()` returns the stub's fake BDF strings in sorted order, which proves `zeDriverGet` then `zeDeviceGet` then `zesDevicePciGetProperties` then `format_pci_bdf` end to end.
- [ ] Two successive `refresh(&mut state, "<stub bdf>")` calls produce a `LevelZeroReadout` whose `primary_engine_utilization` and `power_watts` equal the values implied by the stub's fixed per-call deltas, asserted as exact constants rather than ranges.
- [ ] `apply_to_gpu_info` writes the stub's temperature, power, memory, and frequency into a `GpuInfo`, and the test asserts the resulting field values, so a wrong struct offset changes an observable number rather than passing silently.
- [ ] The tests fail loudly rather than skipping when the stub is absent or the `LD_LIBRARY_PATH` binding stops working, following the #365 pattern: an env key that arms the assertion, a marker printed only after asserting, and a workflow `grep` that fails the job when the marker does not appear.
- [ ] Reachability is proven rather than assumed: a deliberate wrong value injected into the stub makes the new tests fail, and reverting it makes them pass. Record both runs in the PR.
- [ ] A comment at the top of the stub source states explicitly what is verified and what is not. Verified: name each `#[repr(C)]` type in `ffi.rs` and `ffi/sysman.rs` whose fields the stub writes and the test reads back. Not verified: real driver behaviour, realistic value ranges, Windows and `ze_loader.dll`, and any Sysman domain the stub does not implement.
- [ ] The stub cannot be mistaken for a real backend in a shipped build. `LIBZE_PATHS` is unchanged, no library-path environment override is added to production code, and the stub is not compiled into the `all-smi` binary or into any release artifact. State in the PR how each of the three was checked.
- [ ] The pre-existing `the_installed_loader_exports_every_symbol_we_resolve` step still passes against the real `libze1` on the same job, and the two runs do not interfere (they must be separate processes, per the `LZ_RUNTIME` `OnceCell` at `loader.rs:137`).
- [ ] `cargo test`, `cargo fmt --check`, and `cargo clippy --all-targets -- -D warnings` all pass, matching `ci.yml:94-105`.

## Verification

```bash
# Linux host with a C toolchain and the Level Zero runtime plus headers.
sudo apt-get install -y libze1 libze-dev   # or: level-zero level-zero-dev

# Build the stub into its own directory so LD_LIBRARY_PATH ordering is unambiguous.
mkdir -p target/lz-stub
cc -shared -fPIC -Wl,-soname,libze_loader.so.1 \
   -o target/lz-stub/libze_loader.so.1 tests/fixtures/level_zero/stub_ze_loader.c

# Drive the backend against the stub, in its own test process.
LD_LIBRARY_PATH="$PWD/target/lz-stub:$LD_LIBRARY_PATH" \
ALL_SMI_LEVEL_ZERO_STUB=1 \
  cargo test --test level_zero_stub -- --nocapture

# The real-loader check from #365 must still pass, unchanged.
ALL_SMI_EXPECT_LEVEL_ZERO_LOADER=1 \
  cargo test --lib the_installed_loader_exports_every_symbol_we_resolve -- --nocapture

# Full gate.
cargo test
cargo fmt --check
cargo clippy --all-targets -- -D warnings
```

A pass looks like this: the stub tests report the exact synthetic values with no ranges or tolerances, the new marker line appears in the output and the workflow grep finds it, and the `#365` step still reports its own marker against the real `libze1`.

Two manual steps that cannot be automated here. First, inject a wrong constant into the stub, confirm the tests fail, revert it, and confirm they pass. Second, on a machine with a real Intel GPU, confirm the backend still binds the real loader and reports real values, since the stub route must be unreachable outside CI.

## Technical Considerations

- Refs #365 (merged as `dd64e21`), #373, #364, #252, #248.
- Stub and probe precedent: PR #376, and the method write-up at `src/service_cmd/mod.rs:105-135`.
- Blockers for extending this to Windows later: #366, #367, #368.

inureyes:
feat(ci): drive the Level Zero backend against a stub loader

The loader check from #365 stops at `dlopen` plus symbol resolution, and deliberately so: a runner with no Intel GPU has no driver for the loader to return, so `zeInit` failing there is correct rather than a defect. The consequence is that everything past `try_load_library` is unexercised by any runner this project has. Device enumeration, the count-then-buffer idiom and its clamp, the BDF mapping, the delta arithmetic, and the metric refresh all ship on the strength of review alone.

The struct layouts are the sharpest edge. The coverage that exists compares our `#[repr(C)]` struct against our own transcription of the spec, which catches a mis-sized field and did once: `zes_pci_properties_t` declared three `ze_bool_t` fields as `u32`, inflating it from 56 to 64 bytes. It cannot catch two same-typed fields in the wrong order, a wrong offset that preserves the total, or a spec value transcribed wrong in both the struct and its assertion.

## The stub

`tests/fixtures/level_zero/stub_ze_loader.c` plays the driver's part, exporting all 23 symbols the backend resolves: the four mandatory ones, plus `zesInit` so initialisation takes the modern Sysman route, plus the engine, power, temperature, memory, frequency, and fan families.

It is compiled against the **vendor headers**, not against our transcription. That is the whole point: the vendor's own struct is filled on the C side and read back through our type on the Rust side, so a value arriving intact is evidence the two layouts agree field by field. A hand-written stub using our definitions would only prove our definitions agree with themselves.

Values are chosen to be individually identifiable. `zes_freq_state_t` carries five `f64` fields; the stub gives each a different number, so `actual` read one field early reports 1200, 2200, or 2300 rather than a plausible 2100. Counters advance by a fixed step per call, no clock reads and no randomness, so the second refresh is exactly 25.00% compute, 10.00% render, and 45.00 W, asserted as constants rather than ranges.

## How it binds, and why nothing shipped can reach it

`LIBZE_PATHS[0]` is the bare SONAME `libze_loader.so.1`, and `dlopen` on a bare SONAME searches `LD_LIBRARY_PATH` before the default paths. A stub built under that name in a directory placed first therefore wins over the real `libze1` the same job installs, with no new branch in the loader.

The rejected alternative was an `ALL_SMI_LEVEL_ZERO_LIBRARY` path override, which would have put a "load this arbitrary shared object" switch into every released binary in exchange for nothing this route does not provide. So: `LIBZE_PATHS` is unchanged, no environment override exists in production code, and the stub is neither compiled into the `all-smi` binary nor into any release artifact. `LD_LIBRARY_PATH` is set on one workflow step, so every other step still sees the real loader.

## The tests

`tests/level_zero_stub.rs` is its own integration binary, therefore its own process, because `LZ_RUNTIME` is a process-wide `OnceCell` and the library's unit tests already latch it. Whoever ran first would otherwise decide which loader the entire binary sees.

Five tests, driving only the public API: sorted BDF enumeration end to end; the exact delta-derived percentages and watts; the point-in-time families landing in a real `GpuInfo`; a device whose enumerator reports 4096 handles and then fills one, exercising both the `MAX_L0_HANDLES` clamp and the post-fill truncate; and an unknown BDF binding to nothing.

They fail loudly rather than skipping, in the same three-part shape as #365: an env key arms the assertions, a marker prints only after they pass, and the CI step greps for it.

## Verification

- The C stub cannot be compiled from this macOS host: it needs the vendor headers, which have no macOS package. CI is the compile check, and the build step also asserts with `nm` that all 23 symbols are exported.
- Rust side verified locally: the integration binary compiles both with the backend on (through the scratch probe) and off (macOS, where the file compiles away), and runs green unarmed.
- `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` clean.

Reachability proof and the full CI evidence follow in a PR comment once the first run reports.

Closes #379

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -122,6 +122,69 @@ jobs:
               exit 1
               ;;
           esac
+      # Everything past `try_load_library` needs a driver to answer, and no
+      # runner has an Intel GPU. A synthetic loader supplies the answers so
+      # the rest of the backend runs here: device enumeration, the
+      # count-then-buffer idiom and its clamp, the BDF mapping, the delta
+      # arithmetic, and above all the struct layouts.
+      #
+      # The headers are the point. The stub is compiled against the vendor's
+      # own `ze_api.h` and `zes_api.h`, so a value that survives into an
+      # all-smi `#[repr(C)]` field proves the two layouts agree field by
+      # field. The size assertions we already have compare our struct to our
+      # own transcription of the spec, which cannot catch a field-order swap
+      # or an offset error that preserves the total.
+      - name: Install the Level Zero headers
+        run: sudo apt-get install -y libze-dev || sudo apt-get install -y level-zero-dev
+
+      # Built into its own directory so LD_LIBRARY_PATH ordering is
+      # unambiguous against the real libze1 installed above, which the
+      # previous step still needs.
+      - name: Build the stub Level Zero loader
+        run: |
+          set -euo pipefail
+          mkdir -p target/lz-stub
+          cc -shared -fPIC -Wall -Wextra -Werror \
+             -Wl,-soname,libze_loader.so.1 \
+             -o target/lz-stub/libze_loader.so.1 \
+             tests/fixtures/level_zero/stub_ze_loader.c
+          # Every symbol the backend resolves must be exported, or the test
+          # below fails later and less clearly.
+          for sym in zeInit zesInit zeDriverGet zeDeviceGet \
+                     zesDevicePciGetProperties zesDeviceEnumEngineGroups \
+                     zesEngineGetProperties zesEngineGetActivity \
+                     zesDeviceEnumPowerDomains zesPowerGetEnergyCounter \
+                     zesDeviceEnumTemperatureSensors zesTemperatureGetProperties \
+                     zesTemperatureGetState zesDeviceEnumMemoryModules \
+                     zesMemoryGetProperties zesMemoryGetState \
+                     zesDeviceEnumFrequencyDomains zesFrequencyGetProperties \
+                     zesFrequencyGetState zesDeviceEnumFans \
+                     zesFanGetProperties zesFanGetState; do
+            if ! nm -D --defined-only target/lz-stub/libze_loader.so.1 | grep -q " $sym\$"; then
+              echo "::error::the stub does not export $sym"
+              exit 1
+            fi
+          done
+
+      # `LIBZE_PATHS[0]` is the bare SONAME, and dlopen searches
+      # LD_LIBRARY_PATH before the default paths, so the stub wins here
+      # without any production code knowing it exists. The variable is set
+      # on this step alone, so every other step still sees the real loader.
+      #
+      # Its own test binary, therefore its own process: `LZ_RUNTIME` is a
+      # process-wide OnceCell and the library's unit tests already latch it.
+      - name: Drive the backend against the stub loader
+        env:
+          ALL_SMI_LEVEL_ZERO_STUB: "1"
+        run: |
+          set -euo pipefail
+          log="${RUNNER_TEMP:-/tmp}/level-zero-stub.log"
+          LD_LIBRARY_PATH="$PWD/target/lz-stub:${LD_LIBRARY_PATH:-}" \
+            cargo test --test level_zero_stub -- --nocapture | tee "$log"
+          if ! grep -q 'all-smi: level-zero-stub-assertions-ran' "$log"; then
+            echo "::error::the stub tests did not assert. Either ALL_SMI_LEVEL_ZERO_STUB no longer reaches them, or the LD_LIBRARY_PATH binding stopped working and every test returned early."
+            exit 1
+          fi
 
       # `cargo test` with no flags now covers the Level Zero backend:
       # build.rs turns `all_smi_level_zero` on for every Linux target, so
--- a/tests/fixtures/level_zero/stub_ze_loader.c
+++ b/tests/fixtures/level_zero/stub_ze_loader.c
@@ -476,11 +476,8 @@ zesFrequencyGetState(zes_freq_handle_t hFrequency, zes_freq_state_t *pState) {
     pState->currentVoltage = 1.05;
     pState->request = 2200.0;
     pState->tdp = 2300.0;
-    /* REACHABILITY PROOF, reverted in the next commit: `actual` and
-     * `efficient` swapped. This is exactly the defect class the size
-     * assertions cannot see, so the test must go red. */
-    pState->efficient = STUB_FREQ_ACTUAL_MHZ;
-    pState->actual = 1200.0;
+    pState->efficient = 1200.0;
+    pState->actual = STUB_FREQ_ACTUAL_MHZ;
     pState->throttleReasons = 0;
     return ZE_RESULT_SUCCESS;
 }
--- a/tests/fixtures/level_zero/stub_ze_loader.c
+++ b/tests/fixtures/level_zero/stub_ze_loader.c
@@ -1,0 +1,541 @@
+/*
+ * Copyright 2025 Lablup Inc. and Jeongkyu Shin
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/*
+ * A synthetic Level Zero loader, built and used only by CI.
+ *
+ * ## Why it exists
+ *
+ * Every runner this project has lacks an Intel GPU, so `zeInit` has no
+ * driver to return and everything past `try_load_library` goes unexercised:
+ * device enumeration, the count-then-buffer idiom, the BDF mapping, the
+ * delta arithmetic, and the struct layouts the driver writes into. This
+ * library plays the driver's part so that code runs on an ordinary runner.
+ *
+ * ## What it verifies
+ *
+ * It is compiled against the **vendor headers**, not against all-smi's own
+ * transcription of them. That is the entire point. The existing size
+ * assertions in `tests/ffi_layout.rs` compare our struct to our own reading
+ * of the spec, which catches a mis-sized field but not two same-typed
+ * fields in the wrong order, nor a wrong offset that preserves the total,
+ * nor a spec value transcribed wrong in both places. Here the vendor's own
+ * struct definition is filled in on the C side and read back through
+ * all-smi's `#[repr(C)]` type on the Rust side, so a value arriving intact
+ * proves the two layouts agree field by field.
+ *
+ * Types verified this way, all in `src/device/readers/intel_gpu_level_zero`:
+ *   ffi.rs         zes_pci_address_t, zes_pci_speed_t, zes_pci_properties_t,
+ *                  zes_engine_properties_t, zes_engine_stats_t,
+ *                  zes_power_energy_counter_t
+ *   ffi/sysman.rs  zes_temp_properties_t, zes_mem_properties_t,
+ *                  zes_mem_state_t, zes_freq_properties_t,
+ *                  zes_freq_state_t, zes_fan_properties_t
+ *
+ * ## What it does NOT verify
+ *
+ * Real driver behaviour, plausible value ranges, error paths a real driver
+ * would take, Windows and `ze_loader.dll` (out of scope, see the issue),
+ * and any Sysman domain not implemented below. A green run here means the
+ * plumbing and the layouts are right, not that an Intel GPU would report
+ * anything in particular.
+ *
+ * ## Determinism
+ *
+ * No clock reads and no randomness. Counters advance by a fixed step per
+ * call so the derived percentages and watts are exact constants that the
+ * test asserts on directly, never ranges.
+ */
+
+#include <level_zero/ze_api.h>
+#include <level_zero/zes_api.h>
+
+#include <stdint.h>
+#include <string.h>
+
+/* ------------------------------------------------------------------ *
+ * Handles
+ *
+ * Every handle is the address of a distinct static object, so a handle
+ * mixed up between calls dereferences to visibly wrong data rather than
+ * to a plausible neighbour.
+ * ------------------------------------------------------------------ */
+
+static int stub_driver;
+
+/* Two devices at addresses that exercise a non-trivial bus byte and a
+ * sort order: 0xaf sorts after 0x03 numerically and as a string. */
+static int stub_device_a;
+static int stub_device_b;
+
+static int stub_engine_compute;
+static int stub_engine_render;
+static int stub_power;
+static int stub_temp;
+static int stub_mem;
+static int stub_freq;
+static int stub_fan;
+
+/* Device B owns nothing but its PCI address. It exists to prove that a
+ * second device is enumerated and sorted, and to exercise the two count
+ * edge cases below without disturbing device A's exact values. */
+
+/* ------------------------------------------------------------------ *
+ * Fixed per-call steps
+ *
+ * The backend derives engine busy from delta(activeTime)/delta(timestamp)
+ * and power from delta(energy in microjoules)/delta(timestamp in
+ * microseconds), which is microjoules per microsecond, that is watts. With
+ * the steps below the second refresh must produce exactly 25.00% compute,
+ * 10.00% render, and 45.00 W.
+ * ------------------------------------------------------------------ */
+
+#define STUB_TICK_US            1000000ULL
+#define STUB_COMPUTE_ACTIVE_US   250000ULL   /* 25.00% of a tick */
+#define STUB_RENDER_ACTIVE_US    100000ULL   /* 10.00% of a tick */
+#define STUB_ENERGY_UJ         45000000ULL   /* 45.00 W over a tick */
+
+static uint64_t compute_calls;
+static uint64_t render_calls;
+static uint64_t power_calls;
+
+/* Point-in-time values. Each is unique within its struct so a field read
+ * at the wrong offset yields an obviously wrong number. */
+#define STUB_TEMP_C            61.0
+#define STUB_MEM_SIZE      12884901888ULL   /* 12 GiB */
+#define STUB_MEM_FREE       4294967296ULL   /*  4 GiB, so 8 GiB used */
+#define STUB_FREQ_ACTUAL_MHZ    2100.0
+#define STUB_FREQ_MIN_MHZ        300.0
+#define STUB_FREQ_MAX_MHZ       2400.0
+#define STUB_FAN_RPM            1800
+
+/* ------------------------------------------------------------------ *
+ * Count-then-buffer helper
+ *
+ * The real API contract, which `enumerate_devices` and every `populate_*`
+ * helper depend on: a non-null count pointer with a null buffer means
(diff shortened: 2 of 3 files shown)


inureyes:
## Reachability proven, not assumed

The acceptance criteria ask for a deliberate wrong value to make the tests fail. A plain value change would only prove the assertion runs, so the injection was the defect class this PR exists to catch instead: `actual` and `efficient` swapped in the stub's `zes_freq_state_t` write. Same type, same total struct size, wrong order, which is exactly what the size assertions in `tests/ffi_layout.rs` cannot see.

**Injected** (`b8c6f6b`), run [32623903057](https://github.com/lablup/all-smi/actions/runs/32623903057), Test Suite **fail** in 2m40s:

```
thread 'the_point_in_time_families_land_in_gpu_info' panicked at tests/level_zero_stub.rs:197:5:
assertion `left == right` failed: frequency came back wrong
  left: 1200
 right: 2100
test result: FAILED. 4 passed; 1 failed
```

`1200` is `efficient`. It arrived at `actual`'s offset, and the test named the field and the number. The other four tests stayed green, so the failure was scoped to the struct that was actually broken.

**Reverted** (`d2d36ae`), and the run before the injection ([32623254357](https://github.com/lablup/all-smi/actions/runs/32623254357)) is what a clean pass looks like:

```
running 5 tests
all-smi: level-zero-stub-assertions-ran     (x5, one per test)
test result: ok. 5 passed; 0 failed
```

Five markers for five tests: every one armed and asserted rather than returning early.

## The C stub compiled against the vendor headers on the first attempt

`libze-dev` resolved on `ubuntu-latest`, `cc -Wall -Wextra -Werror` produced no diagnostics, and the `nm -D` gate confirmed all 23 symbols are exported. So the round trip is real: the vendor's own struct definitions are filled on the C side and read back through all-smi's `#[repr(C)]` types, for all twelve types listed in the stub's header comment.

Worth stating plainly, since it shaped how this was built: the stub cannot be compiled from the macOS development host, which has no vendor headers available. The Rust half was verified locally (the integration binary compiles with the backend on through the scratch probe and compiles away with it off, and runs green unarmed), and CI is the compile check for the C half.

## Nothing shipped can reach the stub

Each of the three claims, checked rather than asserted:

- `LIBZE_PATHS` is unchanged. `git diff main -- src/device/readers/intel_gpu_level_zero/loader.rs` is empty.
- No library-path override exists in production code. The binding is `LD_LIBRARY_PATH`, set on one workflow step, because `LIBZE_PATHS[0]` is the bare SONAME and `dlopen` searches that variable first. The rejected alternative, an `ALL_SMI_LEVEL_ZERO_LIBRARY` override, would have put a "load this arbitrary shared object" switch into every released binary.
- The stub is not compiled into any artifact. It is a `.c` file under `tests/fixtures/`, built only by the workflow step, and `tests/level_zero_stub.rs` is a Cargo integration target that never links into the `all-smi` binary.

The `#365` real-loader check still passes on the same job, against the real `libze1`, in its own process.

inureyes:
## Final review summary

- Review result: approved for merge. The implementation matches issue #379 and keeps the synthetic loader confined to CI and integration-test paths.
- Correctness fixes: commit 2443b4e replaces shared non-atomic C counters with relaxed atomics so parallel Rust tests cannot trigger a data race, and changes Device B to return a real unsupported-feature error for power enumeration while asserting that engine data remains fresh.
- Security: no production loader search path, runtime environment override, dependency, credential, or external input boundary was added. LD_LIBRARY_PATH remains scoped to one CI step.
- Performance: production runtime behavior is unchanged. The only added synchronization is three relaxed counters in the test fixture, and handle allocation remains capped at 256.
- Local validation: cargo fmt --check passed; cargo test passed with 1,859 tests and 2 ignored; cargo clippy --all-targets -- -D warnings -A clippy::result_large_err passed. The unmodified local clippy command is blocked by two pre-existing Rust 1.98 warnings in src/view/data_collection/ssh_strategy.rs, outside this PR.
- CI validation: run 32653173247 passed Test Suite, Packaging Sync, systemd smoke, launchd smoke, Build Check, and CLA. The vendor-header C fixture compiled with -Wall -Wextra -Werror, the armed stub target printed five assertion markers, and count_edges_and_a_failing_family_are_isolated passed.
- Remaining scope limitation: real Intel driver behavior and Windows ze_loader.dll remain intentionally unverified, as documented in #379.

No unresolved blocking findings remain.

PR #393, opened 2026-08-24
fix(doctor): act on what real Windows hardware showed

inureyes:
## Summary

Four corrections that came out of running `all-smi` on real Windows hardware: first the CI runner (a VMware VM), then a Ryzen AI MAX+ 395 with a Radeon 8060S and a detached NVIDIA eGPU. Three of the four are the same defect in different places: **a check held the value that answered the question and did not print it.**

## 1. The elevation guard gave the wrong advice

The service smoke test told the reader to reconfigure the runner as an administrator. `windows-on-macmini02-x64` reports `elevated: False`, and the decision was to leave it that way: an elevated persistent runner executes whatever lands on `main` with Administrator rights, and a one-off check on a dedicated machine is the cheaper risk. #357 carries that decision; the workflow now agrees with it. Also drops two references to `ENABLE_WINDOWS_SERVICE_SMOKE`, gone since #392.

## 2. The Windows capacity did not say which pool it came from

`resolve_adapter_memory` returns the dedicated pool above a 1 GiB floor and the shared aperture otherwise. `windows.gpu.perf_counters` printed the number without the distinction, and I misread the VM's output as a result. It now prints `95.8 GiB dedicated` or `... shared aperture`.

For the record, the question that ambiguity blocked is now settled by the AMD host, with nothing left to interpret:

| Source | Same adapter, same moment |
|---|---|
| WMI `Win32_VideoController.AdapterRAM` | `4293918720` (3.999 GiB) |
| DXGI, via this check | **95.8 GiB** |
| Firmware carve-out, confirmed by the operator | 96 GB of 128 GB |

`4293918720` is `0xFFF00000`, the largest MiB-aligned value below `2^32`. The identical number appeared on the VM's 8 GiB adapter, which is what shows it to be a saturation artifact rather than a configured size. A `uint32` cannot carry 96 GB, so the WMI baseline was wrong by a factor of 24 on this machine. That is the defect #348 replaced.

## 3. The ADL readout hid the temperature it actually used

`describe_readout` printed edge, hotspot and mem and left out gfx, which is the second link in `primary_temperature_c`'s chain. A Strix Halo APU reports none of edge, mem or hotspot and does report gfx, so `amd.adl.sensors` said

```
edge=NoneC hotspot=NoneC mem=NoneC
```

while the TUI correctly showed **41 C**, taken from the gfx sensor at raw index 28. A field-verification dump should not need the source read to reconcile it with the product. The clock keeps its value under `gfxclk` so the two `gfx` quantities stay apart. Nothing string-matches this format.

## 4. A host with no NVIDIA GPU was reported as failing

`check_nvml_loadable` turned every `Nvml::init()` error into `Fail` with no branch for the commonest cause, and `check_smi` had the same shape once the binary was found. A driver survives its hardware being unplugged, so finding `nvidia-smi` says nothing about a device being present.

A machine whose only NVIDIA GPU is external and detached reported two failures and exited 2 while healthy. Unelevated it printed `NVML_ERROR_NO_PERMISSION`, sending the operator after an elevation problem that does not exist; elevated, the same host reports the real condition, that it cannot reach the driver.

Both checks now stand down when `has_nvidia()` is false, which is the position `level_zero.init` and `level_zero.devices` already take for Intel, for the reason #381 records. The skip message does not assert "no GPU": `has_nvidia()` asks `nvidia-smi -L`, so a false answer covers no hardware, an unreachable driver, and an unelevated Windows shell. It names all three.

## Test plan

- [x] `cargo fmt --check`, `cargo clippy --lib -- -D warnings`, `cargo check --lib` clean
- [x] `cargo test --lib doctor`: 47 passed
- [x] The NVIDIA gate does not suppress the positive path: on a Linux host that has an NVIDIA GPU, all five `nvidia.*` checks still pass and `doctor --only nvidia` exits 0
- [x] Workflow parses; jobs and step count unchanged
- [ ] The two Windows-gated changes (2 and 3) are not compiled locally: this host has no `x86_64-pc-windows-msvc` toolchain and both sit behind `cfg(target_os = "windows")` with no `test` arm. Field accesses were verified against the struct definitions. The `windows-checks` job from #392 compiles them on merge.

### DIFF
--- a/.github/workflows/windows-selfhosted.yml
+++ b/.github/workflows/windows-selfhosted.yml
@@ -167,8 +167,15 @@ jobs:
   #
   # THIS JOB HAS NEVER EXECUTED. It was written alongside the backend on
   # a macOS host with no access to Windows or to the runner, so treat its
-  # first real run as part of the work, not as a regression signal. That
-  # first run is what settles issue #357.
+  # first real run as part of the work, not as a regression signal.
+  #
+  # It will not be this runner that runs it first. `windows-on-macmini02-x64`
+  # reports `elevated: False`, measured in run 32705270378, and the decision
+  # was to leave it that way: an elevated persistent runner executes whatever
+  # lands on main with Administrator rights, and a one-off check on a
+  # dedicated machine is the cheaper risk. Issue #357 records that, and the
+  # guard below states it rather than telling the next reader to go and
+  # elevate the box.
   #
   # The cleanup step is `if: always()` on purpose: a half-finished run
   # that left `all-smi` registered would poison every later run on a
@@ -203,9 +210,9 @@ jobs:
       # feature, so a plain build already includes the backend. Passing the
       # flag would compile the same thing while testing a configuration we
       # do not ship; leaving it off means this job builds exactly what the
-      # release artifact is. The job is opt-in via
-      # ENABLE_WINDOWS_SERVICE_SMOKE, so this is coverage when the runner
-      # is available rather than a guarantee.
+      # release artifact is. The job runs only when dispatched with
+      # `run_service_smoke`, so this is coverage when someone asks for it
+      # rather than a guarantee.
       - name: Build all-smi
         run: cargo build --bin all-smi
 
@@ -214,7 +221,7 @@ jobs:
           $id = [Security.Principal.WindowsIdentity]::GetCurrent()
           $principal = New-Object Security.Principal.WindowsPrincipal($id)
           if (-not $principal.IsInRole([Security.Principal.WindowsBuiltInRole]::Administrator)) {
-            Write-Output "::error::the runner service is not elevated; every mutating Service Control Manager action needs Administrator rights. Reconfigure the runner to run as an administrator, or unset ENABLE_WINDOWS_SERVICE_SMOKE."
+            Write-Output "::error::the runner service is not elevated, so every mutating Service Control Manager action here would fail. This is expected on windows-on-macmini02-x64 and is not a defect to work around: elevating a shared persistent runner was considered and declined (see #357). Run this on a dedicated Windows machine instead."
             exit 1
           }
 
--- a/src/doctor/checks/amd.rs
+++ b/src/doctor/checks/amd.rs
@@ -210,10 +210,20 @@ fn check_adl_adapters(_ctx: &CheckCtx) -> CheckResult {
 /// so the formatting lives here rather than at each call site.
 #[cfg(target_os = "windows")]
 fn describe_readout(readout: &crate::device::readers::amd_adl::sensors::AdlReadout) -> String {
+    // `temperature_gfx_c` is printed because it is the second link in
+    // `primary_temperature_c`'s chain (edge, then gfx, then hotspot), and
+    // omitting it made this dump contradict the reading all-smi actually
+    // publishes. A Strix Halo APU reports none of edge, mem or hotspot and
+    // does report gfx, so the summary said `edge=None hotspot=None mem=None`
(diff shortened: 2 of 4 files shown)


inureyes:
## Where these four changes came from: a manual run on real hardware

Every commit here was written because a real machine printed something that did not match what the code claimed. Recording the evidence so the reasoning survives the diff.

**Host**: Ryzen AI MAX+ 395 with a Radeon 8060S (Strix Halo), Windows 11 Pro 26200, 128 GB physical with **96 GB allocated to the GPU in firmware**, leaving 31.6 GiB to the OS. An NVIDIA eGPU is normally attached and was **detached** during the run. Source build of `main` at v0.26.1.

### Commit 2, the memory pool: settled, with a 24x margin

| Source | Same adapter, same moment |
|---|---|
| WMI `Win32_VideoController.AdapterRAM` | `4293918720` (3.999 GiB) |
| DXGI, via `windows.gpu.perf_counters` | **95.8 GiB** |
| Actual firmware carve-out | 96 GB of 128 GB |

`4293918720` is `0xFFF00000`, the largest MiB-aligned value below `2^32`. The **identical number** appeared earlier on the CI VM's 8 GiB adapter, and two adapters of wildly different size reporting the same figure is what shows it to be a saturation artifact rather than a configured size. I had guessed the opposite from the VM data alone, which is exactly why this commit makes the check print which pool it used.

### Commit 3, the gfx temperature: the check contradicted the product

```
PASS amd.adl.sensors  ... edge=NoneC hotspot=NoneC mem=NoneC power=Some(22.0)W ...
                      raw: 1=641 2=1000 3=609 7=12 19=1 23=22 28=41 29=41 ...
```

All three printed temperatures are `None`, and the TUI showed **41 C** at the same time. The value is at raw index 28, which is `PMLOG_TEMPERATURE_GFX`, the second link in `primary_temperature_c`'s chain. This APU reports none of edge (8), mem (9) or hotspot (27), so the chain correctly falls through to gfx, and the summary simply did not print the field it used.

### Commit 4, the NVIDIA verdict: the error message was hiding the cause

Two runs on the same host, same binary:

```
unelevated:  NVIDIA-SMI has failed because you do not have sufficient permissions.
elevated:    NVIDIA-SMI has failed because it couldn't communicate with the NVIDIA driver.
```

The permission error is a mask. Underneath it is a detached eGPU. `doctor` reported `FAIL nvidia.nvml.loadable` with `NVML_ERROR_NO_PERMISSION` and exited 2 on a healthy host, sending the reader after an elevation problem that does not exist. That two-run split is why the skip message names all three states instead of asserting "no GPU".

### Commit 1, the elevation note

The CI runner reports `elevated: False` and stays that way by decision (#357). The guard previously advised the opposite.

## Also validated on the same run, outside this PR

- **#361's `AdapterInfo` layout is correct.** `amd.adl.adapters` returned 5 rows, **all populated**, all strings legible, all `bus=189 device=0 function=0`, grouping to 1 physical card. Had the transcribed 1572 bytes been wrong, ADL would have written fewer rows and left the tail blank, which is precisely what the blank-row rejection was added to catch.
- **#371's verdict logic behaves.** `amd.adl.per_adapter` reported per-card differentiation as **unobservable** on a one-card host rather than claiming IDENTICAL.
- **#357 is closed.** The SCM service backend passed all 11 lifecycle assertions on its first ever execution.

## What this PR still has not proved

Commits 2 and 3 touch `cfg(target_os = "windows")` code with no `test` arm, and this development host has no `x86_64-pc-windows-msvc` toolchain. They are not compiled here. The `windows-checks` job from #392 compiles them on merge, which is that job's first useful catch opportunity.

Issue #366, opened 2026-08-18
fix: four --lib tests fail on native Windows


inureyes:
## Problem / Background

`cargo test --release --lib` on `origin/main` (`b92ea0a`) gives `1417 passed; 4 failed` on a native Windows host.

These failures reproduce on `main` itself, so they are pre-existing debt and were **not** introduced by any open PR.

### Verification environment

- Host: Windows 11 Pro 26200, native `x86_64-pc-windows-msvc` (not cross-compiled)
- Toolchain: rustc 1.97.1 / cargo 1.97.1
- Commit verified against: `origin/main` at `b92ea0a` ("fix: report the pinned libamdgpu_top version and level_zero in doctor (#363)")

### Discovery context

These surfaced while doing hardware validation of PR #361 on a real AMD Windows host. See https://github.com/lablup/all-smi/pull/361#issuecomment-5323899954

### 1. `device::common::validation::tests::test_validate_command_path` — `src/device/common/validation.rs:140`

`assertion failed: validate_command_path(&PathBuf::from("/bin/ls"))`, with stdout `Non-absolute command path rejected: "/bin/ls"`.

Cause: `/bin/ls` has no drive letter, so `Path::is_absolute()` is false on Windows and the validator rejects it. The test asserts a Unix path shape.

### 2. `metrics::energy_wal::tests::expand_tilde_replaces_home_prefix` — `src/metrics/energy_wal.rs:940`

`left: "C:\Users\<user>\.cache/all-smi/energy-wal.bin"`, `right: "/tmp/fake-home/.cache/all-smi/energy-wal.bin"`.

Cause: the test overrides `HOME`, but home resolution on Windows uses `USERPROFILE`, so the override is ignored and the real profile directory is returned.

### 3. `metrics::energy_wal::tests::torn_final_record_is_discarded` — `src/metrics/energy_wal.rs:771`

`called Result::unwrap() on an Err value: Os { code: 32, kind: Uncategorized, message: "The process cannot access the file because it is being used by another process." }`

Cause: Windows does not permit the file operation while an existing handle to the file is still open. Unix allows it, so the test passes on Linux.

### 4. `doctor::bundle::tests::bundle_writes_expected_entries` — in `src/doctor/bundle.rs`

`bundle ok: failed to create bundle file "C:\Users\<user>\AppData\Local\Temp\.tmpXXXXXX"`, `Caused by: The process cannot access the file because it is being used by another process. (os error 32)`

[...]

PR #394, opened 2026-08-24
fix: clear the clippy denials on native Windows

### LINKED ISSUE #367: fix: clippy -D warnings fails at four sites on native Windows
## Problem / Background

`cargo clippy --release --lib --tests -- -D warnings` fails on `origin/main` (`b92ea0a`) on a native Windows host.

There are four findings across three files. Every one of them is inside `cfg(target_os = "windows")` or Windows-only code, which is why the Linux CI clippy job has never reported them.

These failures reproduce on `main` itself, so they are pre-existing debt and were **not** introduced by any open PR.

### Verification environment

- Host: Windows 11 Pro 26200, native `x86_64-pc-windows-msvc` (not cross-compiled)
- Toolchain: rustc 1.97.1 / cargo 1.97.1
- Commit verified against: `origin/main` at `b92ea0a` ("fix: report the pinned libamdgpu_top version and level_zero in doctor (#363)")

### Discovery context

These surfaced while doing hardware validation of PR #361 on a real AMD Windows host. See https://github.com/lablup/all-smi/pull/361#issuecomment-5323899954

## Findings

1. `src/utils/command_timeout.rs:190:9` — `error: unused import: super::*`
2. `src/device/cpu_windows.rs:49:9` — `error: this if statement can be collapsed` (`clippy::collapsible_if`)
3. `src/device/cpu_windows.rs:63:9` — `error: this if statement can be collapsed` (`clippy::collapsible_if`)
4. `src/device/windows_temp/amd_ryzen.rs:228:9` — `error: this if statement can be collapsed` (`clippy::collapsible_if`)

All four are mechanical fixes.

## Proposed Solution

Remove the unused import and collapse the three nested `if` statements. No behavioural change is expected or intended.

## Acceptance Criteria

- [ ] The unused `super::*` import at `src/utils/command_timeout.rs:190` is removed
- [ ] The three `clippy::collapsible_if` sites in `src/device/cpu_windows.rs` and `src/device/windows_temp/amd_ryzen.rs` are collapsed
- [ ] `cargo clippy --release --lib --tests -- -D warnings` passes on a native `x86_64-pc-windows-msvc` host
- [ ] Linux and macOS clippy remain clean

## Technical Considerations

The reason these accumulated is that no PR-time CI job ever compiles or lints Windows-only code. That gap is tracked separately in #368; fixing these four sites is a prerequisite for landing that job green.

## Related

Root cause of why these accumulated: #368

inureyes:
## Summary

Makes `cargo clippy --all-targets -- -D warnings` pass on native Windows. Two commits, because clearing the four sites this issue names turned out not to be sufficient.

Closes #367

## The four sites (first commit)

All confirmed still present on `ecc60a4` by run [32716362643](https://github.com/lablup/all-smi/actions/runs/32716362643), the first CI run ever to lint this code.

Three are `clippy::collapsible_if`, each collapsing to a let-chain, which this crate already uses elsewhere and edition 2024 supports at the pinned MSRV. The `amd_ryzen.rs` one is worth a note: `take()` runs whether or not the second condition holds, so the lock is cleared even when the library was never initialised and only the uninit call is conditional. A let-chain preserves that order exactly. The two comments explaining why a failed WMI connection is deliberately left as `None` moved above their `if` rather than disappearing with the nesting.

The fourth is the unused `use super::*` in `command_timeout.rs`'s test module. **Deleting it breaks Linux**, where it is used, so the fix has to be a gate. The module is now `#[cfg(all(test, unix))]` rather than gating the import alone: both tests drive Unix programs (`yes`, `printf`), and gating only the import would leave `output_cap_leaves_small_outputs_alone` compiled on Windows with its entire body behind `#[cfg(unix)]`. That is a test reporting `ok` while asserting nothing, and an empty pass is a worse signal than an absent test. Windows loses one vacuous test as a result.

That does leave the output cap, which is cross-platform product code, with no Windows coverage. Restoring it needs a Windows program that emits an unbounded stream, which is a test to write rather than a lint to silence, so it is left for its own change. Both constants it uses are consumed by product code, so gating the module orphans nothing.

## Why a second commit was needed

The Windows run stopped at the lib, so **the bin was never linted**. Its log shows only `could not compile all-smi (lib)` and `(lib test)`. `src/view/data_collection/ssh_strategy.rs` lives in the bin, so as soon as the lib compiles, clippy reaches it and raises `result_large_err` on the two collector signatures. Fixing only the four sites would have moved the failure rather than removed it.

The conventional fix for that lint is to box the `Err`. Here it shrinks nothing:

```
ConnectionStatus                                184
Ok  variant (Vec<GpuInfo>, ConnectionStatus)    208
Err variant (ConnectionStatus,)                 184
Result                                          208
Result with a boxed Err                         208
```

The `Ok` variant carries the same `ConnectionStatus` plus a `Vec`, so it dominates the enum and the `Result` is 208 bytes either way. Boxing would add a heap allocation on every failed host and remove zero bytes.

So both functions take `#[allow(clippy::result_large_err)]` carrying that measurement, and a new test pins it. Suppressing a lint on a claim about sizes is only honest if the claim is checked. If someone later shrinks the `Ok` side, `boxing_the_error_would_not_shrink_the_result` fails and the suppression stops being justified, which is exactly when boxing would become worth its allocation.

These two findings are pre-existing on `main` (confirmed by stashing this branch and re-running clippy) and were recorded once before in PR #382 as a rustc 1.98 difference the CI toolchain did not reproduce.

## Validation

On Linux, rustc 1.98.0:

- `cargo clippy --all-targets -- -D warnings` passes with **no findings at all**, where `main` has two
- `cargo fmt --check` clean
- `cargo test --lib` 1684 passed, `cargo test --bin all-smi` 1861 passed

Not verified on Windows from here. The `windows-checks` job has no `pull_request` trigger by design, so verification is either `gh workflow run "Windows Self-Hosted" --ref fix/issue-367-windows-clippy` or a build on a Windows host.

Note for whoever runs it: the job's Report step fails while #366 is open, because `cargo test --lib` still has its seven failures. The signal to read is the per-step line, which should now say `cargo clippy  success`.

### DIFF
--- a/src/api/shutdown.rs
+++ b/src/api/shutdown.rs
@@ -102,7 +102,13 @@ pub async fn wait_until_serving() {
 }
 
 /// Whether at least one listener has been bound.
-#[cfg_attr(not(windows), allow(dead_code))]
+///
+/// Unlike [`wait_until_serving`], which the Windows service host awaits,
+/// this has no caller outside the tests. The attribute below therefore
+/// keys on `test` rather than on the platform: it was copied from the
+/// sibling above, and `not(windows)` silenced every target except the one
+/// where the item is genuinely dead.
+#[cfg_attr(not(test), allow(dead_code))]
 pub fn is_serving() -> bool {
     serving_latch().is_triggered()
 }
--- a/src/device/cpu_windows.rs
+++ b/src/device/cpu_windows.rs
@@ -46,11 +46,12 @@ thread_local! {
 fn with_cimv2_connection<T, F: FnOnce(&WMIConnection) -> T>(f: F) -> Option<T> {
     WMI_CIMV2_CONNECTION.with(|cell| {
         let mut conn_ref = cell.borrow_mut();
-        if conn_ref.is_none() {
-            if let Ok(wmi_con) = WMIConnection::new() {
-                *conn_ref = Some(wmi_con);
-            }
-            // Silently fail if connection cannot be created
+        // A failed connection is left as `None` on purpose: the caller
+        // gets `None` back and degrades, and the next call retries.
+        if conn_ref.is_none()
+            && let Ok(wmi_con) = WMIConnection::new()
+        {
+            *conn_ref = Some(wmi_con);
         }
         conn_ref.as_ref().map(f)
     })
@@ -60,11 +61,11 @@ fn with_cimv2_connection<T, F: FnOnce(&WMIConnection) -> T>(f: F) -> Option<T> {
 fn with_root_wmi_connection<T, F: FnOnce(&WMIConnection) -> T>(f: F) -> Option<T> {
     WMI_ROOT_WMI_CONNECTION.with(|cell| {
         let mut conn_ref = cell.borrow_mut();
-        if conn_ref.is_none() {
-            if let Ok(wmi_con) = WMIConnection::with_namespace_path("root\\WMI") {
-                *conn_ref = Some(wmi_con);
-            }
-            // Silently fail if connection cannot be created
+        // Same as above: a failed connection stays `None` and is retried.
+        if conn_ref.is_none()
+            && let Ok(wmi_con) = WMIConnection::with_namespace_path("root\\WMI")
+        {
+            *conn_ref = Some(wmi_con);
         }
         conn_ref.as_ref().map(f)
     })
--- a/src/device/readers/windows_gpu_perf.rs
+++ b/src/device/readers/windows_gpu_perf.rs
@@ -125,6 +125,12 @@ pub struct Snapshot {
 }
 
 impl Snapshot {
+    /// Test-only today. This module is gated `cfg(any(target_os =
+    /// "windows", test))`, so off Windows it exists only under `test`,
+    /// where this is used and nothing is reported. A non-test Windows
+    /// build compiles it with no caller, which is the one configuration
+    /// that sees it as dead.
+    #[cfg_attr(not(test), allow(dead_code))]
     pub fn is_empty(&self) -> bool {
         self.adapters.is_empty() && self.processes.is_empty()
     }
--- a/src/device/readers/windows_gpu_perf/ids.rs
+++ b/src/device/readers/windows_gpu_perf/ids.rs
@@ -49,6 +49,11 @@ pub struct AdapterLuid {
 }
 
 impl AdapterLuid {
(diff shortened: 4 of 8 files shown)


inureyes:
## Verified on the real Windows runner: `cargo clippy  success`

Run [32723212182](https://github.com/lablup/all-smi/actions/runs/32723212182) against this branch:

```
cargo fmt --check  success
cargo clippy       success
cargo test --lib   failure
```

The job is red, and that is expected while #366 is open: `cargo test --lib` still has its seven failures on this branch, and the Report step fails if any of the three does. The line to read is `cargo clippy  success`, which is what this PR is for.

## A third commit was needed, and the reason is the same one that made #366 undercount

The first verification run ([32720773764](https://github.com/lablup/all-smi/actions/runs/32720773764)) still failed clippy. Clearing the lib let clippy reach the **bin for the first time**, where it found five more Windows-only denials:

```
unused import: `SocketSetting`            src\main.rs:39
unused variable: `interval`               src\main.rs:324
function `is_serving` is never used       src\api\shutdown.rs:106
method `is_empty` is never used           src\device\readers\windows_gpu_perf.rs:128
associated function `new` is never used   src\device\readers\windows_gpu_perf\ids.rs:52
```

So this issue undercounted for exactly the reason #366 did: nothing could see past the first thing that stopped the build. The real total is nine, four in the lib and five in the bin.

Three of the five are one shape, and **none of them should be deleted**. `Snapshot::is_empty` is asserted on in its own module's tests and `AdapterLuid::new` has ten call sites, all assertions. Their module is gated `cfg(any(target_os = "windows", test))`, so off Windows it exists only under `test`, where the items are used; a non-test Windows build is the single configuration that compiles them with no caller. `is_serving` is the same conclusion by a different route: its `#[cfg_attr(not(windows), allow(dead_code))]` was copied from `wait_until_serving`, which the Windows service host really does await, so keying on the platform silenced every target except the one where the item is actually dead. All three now key on `not(test)`.

The other two are ordinary cfg drift in `main.rs`, both fixed by gating rather than by allowing, so a future consumer on another target fails to compile instead of quietly reintroducing dead code.

## Merge order

This and #395 are independent (zero file overlap) and can merge in either order. Whichever lands second is the one whose Windows run finally goes fully green, and that is the moment the three `continue-on-error` flags in `windows-selfhosted.yml` can come off.

PR #412, opened 2026-09-10
fix: correct Apple Silicon power over-count and 0 W ticks

### LINKED ISSUE #410: fix: macOS power readings alternate between 0 W and inflated values on M5 Max (IOReport energy batch cadence + substring channel matching)
## Problem / Background
On Apple M5 Max (macOS 27.0 build 26A428, all-smi main at 720b8c2), `all-smi view` under load shows CPU power in the System panel alternating every tick between a plausible value and zero, e.g. 40.0 W, 0 W, 40.0 W. The GPU panel's Pkg Power (fed by `combined_power_mw` = cpu + gpu + ane from `NativeMetricsData`, `src/device/macos_native/metrics.rs:85`) swings like 39.2 W, 0.1 W, 32.1 W, 167 W, including values that are impossible for the chip. The same numbers feed the Prometheus exporter (`src/api/metrics/gpu.rs:132-136`, `src/api/metrics/gpu.rs:194-198`, `src/api/metrics/chassis.rs`) and the chassis `cpu_power_watts` / `gpu_power_watts` details (`src/device/readers/chassis/apple_silicon_native.rs:64-74`), so every power surface is wrong, not only the TUI.

## Current Behavior
`NativeMetricsManager::collect_once` (`src/device/macos_native/manager.rs:381-441`) takes one IOReport sample per collection cycle and deltas it against the previous cycle's sample via `IOReport::get_sample_since_last` (`src/device/macos_native/ioreport.rs:992-1030`). `IOReportMetrics::from_sample` (`ioreport.rs:1099`) then converts every `Energy Model` channel delta to watts via `IOReportChannelItem::calculate_watts` (`ioreport.rs:780-798`, energy / elapsed) and sums them into cpu/gpu/ane/dram in `process_energy_channel` (`ioreport.rs:1152-1172`). powermetrics is not involved anywhere in this path. Two independent defects combine.

### Root cause 1: energy counters are published in batches about every 2.1 s, but the delta window is the 1 s poll interval
A probe taking raw IOReport samples every ~117 ms under an 8-thread `yes` load showed that every mJ-unit channel in the `Energy Model` group (`CPU Energy`, `MCPU0`, `MCPU1`, `PCPU`, every per-core channel, `_SRAM`, `DTL`, `GPU0`, `ANE0`, `DRAM0`, ...) advances only once per roughly 18 to 19 ticks, i.e. about every 2.1 s; all other ticks show a delta of exactly 0 for all of those channels. Batch ticks landed at #17, #35, #54, #72 in one run and #03, #21, #40 in another. Only `GPU Energy` (unit nJ, a separately driven counter) advances on its own cadence of roughly 0.6 s. The `CPU Stats/CPU Core Performance States` and `GPU Stats/GPU Performance States` residency channels advance on every 100 ms tick, so residency-derived utilization and frequency are unaffected.

With a 1 s window, consecutive deltas alternate between one whole ~2.1 s batch of energy divided by ~1.0 s (roughly double the true average) and zero energy (0 W). Measured at 1 s cadence, `CPU Energy` gave 0.00, 36.17, 0.00, 33.25, 0.00, 34.24, 0.00, 34.01 W on successive ticks. At a 2 s cadence the same load gave 33.87, 31.06, 31.50, 32.31, 27.20, 30.21, 24.61, 23.86 W with no zeros, still subject to beat effects as the 2.1 s batch period and the 2.0 s window drift apart. `calculate_watts` returns 0.0 for a zero delta and the manager publishes it as a real reading, so the TUI shows 0 W instead of "no new data".

### Root cause 2: substring channel matching sums an entire sub-channel hierarchy together with the roll-up
`process_energy_channel` classifies channels by `channel.contains("CPU") && !channel.contains("GPU")`, `contains("GPU") && !contains("CPU")`, `contains("ANE")`, `contains("DRAM")`. On M5 Max the `Energy Model` group has 364 channels. In one loaded batch tick the CPU match summed:

| family | channels | watts |
|---|---|---|
| `CPU Energy` (roll-up) | 1 | 36.17 |
| `MCPU0`, `MCPU1`, `PCPU` (clusters; their sum equals the roll-up exactly) | 3 | 36.17 |
| `MCPU*DTL*`, `PCPUDTL*` (per-core telemetry sub-channels) | 300 | 26.01 |
| `MCPUx_y` (per-core) | 12 | 5.20 |
| `*_SRAM` | 18 | 3.13 |
| total reported as `cpu_power` | 334 | 106.68 |

That is 2.95x the roll-up. GPU double counts `GPU0` (mJ, batched) plus `GPU Energy` (nJ, continuous). ANE matches only `ANE0` and DRAM only `DRAM0`, which is correct by luck. Combined with root cause 1 (up to 2x on batch ticks) the package figure reaches about 6x the true value, which is where 167 W comes from. `package_power` (`ioreport.rs:1168-1171`) is also recomputed inside the loop only when a CPU-prefixed channel is seen, so its value depends on channel iteration order.

For reference, macmon matches exactly: `"GPU Energy"`, `c.ends_with("CPU Energy")` (covers `DIE_n_CPU Energy` on Ultra), `c.starts_with("ANE")` (`ANE`, `ANE0`, `ANE0_n`), `c.starts_with("DRAM")`. macmon has no handling for the batched cadence (its changelog mentions only M5 voltage-state and core-label fixes), so root cause 1 is ours to solve.

## Proposed Solution
1. Anchored matching in `process_energy_channel`: `GPU Energy` exact for GPU (fall back to `GPU0` only when the sample contains no `GPU Energy` channel), `ends_with("CPU Energy")` for CPU, `starts_with("ANE")` restricted to top-level names (`ANE`, `ANEn`, `ANEn_m`) for ANE, `starts_with("DRAM")` for DRAM. Never sum per-core, cluster, `_SRAM`, or `DTL` sub-channels into the roll-ups. Compute `package_power` once after the loop in `from_sample`, not inside the per-channel function. Because the `GPU0` fallback depends on the whole sample, classification should collect matched channel values first and resolve GPU after iteration.
2. Make the energy delta robust to batched publication: keep a separate energy baseline (sample plus `Instant`) in `IOReport` next to `prev_sample` (`ioreport.rs:884`) that only advances when the roll-up energy counters actually moved. Each collection still takes one sample; residency deltas keep using the previous-tick baseline (they advance every tick), but power is computed as `(energy now - energy at last movement) / (t_now - t_last_movement)`. On a tick where the roll-up delta is zero, keep the previous power reading (do not publish 0 W) and leave the energy baseline in place. On M1 to M4, where counters appear to advance every tick, this degrades to today's behavior. Document the measured ~2.1 s batch cadence in the module docs of `ioreport.rs` so nobody "fixes" it back.
3. Optional guard: clamp obviously impossible per-chip totals only as a last resort; the two fixes above should remove the need.
4. Verify on M1 Ultra and M1 to M4 hardware that channel names and update cadence match these assumptions (an M1 Ultra is available to the maintainer). Record each chip family's `Energy Model` channel inventory as a test fixture.

Rejected alternative: raising the poll interval to 2 s or more hides the zeros but keeps the beat artifacts shown above and does nothing for the 2.95x over-count, so it is not a fix.

## Scope
**In scope:** `src/device/macos_native/ioreport.rs` (`process_energy_channel`, `from_sample`, `get_sample_since_last`, new energy baseline, module docs, unit tests), `src/device/macos_native/manager.rs` (`collect_once` carrying the previous power reading across zero-delta ticks if that state lives in the manager rather than in `IOReport`).

**Out of scope:** powermetrics fallback paths, residency/frequency handling (`process_cpu_channel`, `process_gpu_channel`), Intel Mac chassis power (`src/device/readers/chassis/intel_mac.rs`), UI rendering changes.

## Implementation Notes
- **Reuse**: keep the existing `prev_sample` / `MIN_DELTA_WINDOW` structure in `IOReport` (`ioreport.rs:878-884`, `992-1030`) and add the energy baseline alongside it rather than introducing a second sampler. `calculate_watts` unit handling (mJ/uJ/nJ) stays as is.
- **Testability**: `IOReportChannelItem` wraps a `CFDictionaryRef` (`ioreport.rs:743-749`), so pull the classification into a pure function over `(channel name, unit, energy delta)` that unit tests can drive without IOReport. Add tests with the M5 Max inventory (roll-up + clusters + per-core + SRAM + DTL + `GPU0` + `GPU Energy`) asserting cpu equals the roll-up and gpu is counted once, an Ultra-style `DIE_0_CPU Energy` / `DIE_1_CPU Energy` case, and a zero-delta tick that must retain the previous power without moving the energy baseline.
- **Edge cases**: first collection of a session (no energy baseline yet: publish whatever `get_sample` returns, as today); a batch tick that arrives inside `MIN_DELTA_WINDOW`; a sample where `GPU Energy` is absent (use `GPU0`); very long gaps between movements (still divide by the real elapsed time, never by the poll interval).
- **Error handling**: no new failure paths. A zero-delta tick is not an error and must not be logged at warn level; a debug-level trace of "energy counters unchanged, holding previous reading" is enough.

## Acceptance Criteria
- [ ] With `all-smi view --interval 1` under sustained CPU load on M5 Max, CPU power and Pkg Power show no 0 W ticks and do not exceed the values a 2 s window reports by more than sampling jitter.
- [x] Reported CPU power equals the `CPU Energy` roll-up within rounding, not the sum of sub-channels; GPU power equals `GPU Energy` once.
- [x] Unit tests cover the M5 Max channel inventory, the Ultra `DIE_n_CPU Energy` naming, and the zero-delta tick hold behavior, and pass in `cargo test`.
- [ ] `all_smi_gpu_power_consumption_watts`, `all_smi_ane_power_watts`, the chassis power gauges in `src/api/metrics/chassis.rs`, and the chassis `cpu_power_watts` / `gpu_power_watts` details show the same corrected values, since they all read `NativeMetricsData`.
- [ ] Behavior on M1 to M4 unchanged or explicitly verified on hardware, with the channel inventory recorded as a fixture.
- [x] Integrated into the real code flow: the fix lands in the `IOReport` / `collect_once` path that view mode, API mode, and the chassis reader already use, not in a parallel helper.

## Verification
```bash
cargo build --release
cargo fmt --check
cargo clippy --all-targets
cargo test macos_native
# Manual, on M5 Max under load:
yes > /dev/null & yes > /dev/null & yes > /dev/null & yes > /dev/null &
sudo ./target/release/all-smi view --interval 1
# Pass: CPU power and Pkg Power stay in a stable band (roughly 25 to 40 W for this load) with no 0 W ticks and no values above ~60 W.
./target/release/all-smi api --port 9090 & curl -s localhost:9090/metrics | grep -E 'power'
# Pass: the exported watts match what view mode shows at the same moment.
kill %1 %2 %3 %4 %5
```

## Technical Considerations
Related code: `src/device/macos_native/ioreport.rs` (`process_energy_channel`, `calculate_watts`, `get_sample_since_last`), `src/device/macos_native/manager.rs` (`collect_once`), `src/ui/gpu_sparkline_panel.rs:632-640` (`package_power` reads `combined_power_mw`). Prior M5 work: #167 (MCPU channels / Super cores). Related: #87 (native macOS metrics), #325 (Apple Silicon GPU metrics reporting 0).

inureyes:
Apple Silicon power readings on M5 Max alternated between 0 W and inflated values. Both causes are fixed in the `IOReport` / `collect_once` path that local, view, API, and the chassis reader share, so the Prometheus and chassis gauges pick up the corrected values without edits of their own.

- Exact channel matching (`energy.rs`): `GPU Energy`, names ending in `CPU Energy`, `GPU<n>` only when a sample has no `GPU Energy`, and top-level `ANE`/`DRAM` names. Substring matching summed 334 channels into CPU power (106.68 W against a 36.17 W roll-up) and counted the GPU twice.
- Publication-timed power: the mJ counters publish in batches 2.03 to 2.23 s apart, so each tracked channel is timed by the driver's own timestamp in its raw sample's `RawElements` (IOReportElement offset 24). Readings hold between publications, split tails under 50 ms fold into the next span, and a reading is dropped after 10 s without a publication. A channel whose timestamps are missing, frozen while the value moves, or ahead of the observation clock falls back to observation time for good, which is the old poll-window behavior. The first frame after start reads 0 W until the first publication span closes (up to about 2.2 s).

The issue proposed an `Instant` baseline that advances when the roll-up moves. Observation-time windows stay quantized to the poll grid (1.05x or 0.70x at a 1 s poll, 0.7x to 1.4x beats at the 3 s API default, where no zero tick ever triggers a hold), and split publications need per-channel timing, so the driver timestamp replaces it.

Measured on M5 Max (macOS 27), release build, four `yes` processes:

| check | before | after |
|---|---|---|
| `all-smi local --interval 1`, Pwr/Pkg | 0.0 and 113 to 126 W alternating | 19.3 to 23.9 W after the first frame |
| API `--interval 1`, CPU | 0 and ~116 W alternating | 19 to 38 W as load shifted, no 0 W |
| API `--interval 3` plus a Metal load, CPU / GPU | 29/66 W and 89/142 W beating | 14 to 26 W / 49 to 70 W, smooth |

Chip, combined, and chassis gauges report the same value. Full suite (3535 passed), `clippy --all-targets -D warnings`, and fmt pass.

Not verified: M1 to M4 and M1 Ultra hardware. `cargo test --lib ioreport_energy_diagnostics -- --ignored --nocapture` prints the inventory and cadence needed to add their fixtures. The new tests are macOS-only, and CI runs `cargo test` on Linux only.

Closes #410


inureyes:

### Intent
> Fix M5 Max power readings (substring over-count, 0 W and 2x alternation) with exact channel classification and per-channel publication timestamps from `RawElements`, per #410 and its consolidated requirements.

### Findings Addressed
- [x] The observation-clock fallback measured its first span from the last driver timestamp to the observation time, so a driver with a frozen or zero stamp read about 0 W on the switching sample. The switch now restarts the span from the previous observation, with tests for a zero stamp and for timestamps disappearing mid-stream (MEDIUM, c53026d)
- [x] `// SAFETY:` comments on the new unsafe blocks in `ioreport.rs` (LOW, c53026d)
- [x] Test for per-sample GPU source resolution in `EnergyTracker::readings()` (LOW, c53026d)

### Remaining Items
- [ ] The `energy` tests compile only on macOS and CI runs `cargo test` only on ubuntu-latest, so they never run in CI (LOW), reason: moving the pure module out of the macOS gate is a structural change outside this PR.
- [ ] Ultra names other than `DIE_<n>_CPU Energy` are unverified (LOW, open question). Stripping `DIE_<n>_` for GPU/ANE/DRAM was not applied: macmon's Ultra names are `GPU Energy` and `ANE0_<n>`, and stripping would double count if a roll-up sits next to per-die channels. Running `ioreport_energy_diagnostics` on an M1 Ultra settles it.
- [ ] M1 to M4 and Ultra not verified on hardware.

### Verification
- [x] Tests pass (`cargo test --lib device::macos_native`: 82 passed, 1 ignored; clippy and fmt clean)

inureyes:
## Security and Performance Review

Driver data (RawElements bytes, counters, timestamps) treated as untrusted.

### Findings Addressed
- [x] MEDIUM: a driver timestamp later than the observation time was trusted, so a stalled channel stamped on another clock (or with non-timestamp bytes at offset 24) held its reading forever and was exported as live, because the 10 s staleness check saturated to 0. Stamps more than 1 s ahead now take the existing sticky observation-clock fallback (1259a26, with a stall test that fails on the old code).
- [x] LOW: `scale_mach_ticks` wrapped on overflow via `as u64`; it now saturates and reads a zero denom as 1 (f0e2766).
- [x] LOW: `get_io_channels` cast `IOReportChannels` and its elements with only null checks; both are now CF type-checked (f0e2766).
- [x] LOW: SAFETY comments on two `CFRelease` calls in `ioreport_energy_diagnostics` (f0e2766).

### Checked, no change needed
- CF ownership in `raw_element_timestamp` is sound (get rule, null and CFData type checks, null-safe `CFData::bytes` in core-foundation 0.10.1); `get_sample` releases `sample1` when the second sample fails.
- Energy arithmetic cannot panic or yield NaN/inf: i128 deltas, a span divisor of at least 50 ms, and rebaselining on backwards values or timestamps.
- The tracker map holds only rail channels from the fixed subscription (5 on M5 Max) and is mutated only through `&mut self` under the manager mutex.
- Hot path, M5 Max debug build: `energy_observations` plus the tracker take 114 µs per raw sample (383 channels) against 5.0 ms for `IOReportCreateSamples`. Proportionate at 1 Hz.

### Verification
- `cargo test --lib device::macos_native`: 84 passed, 1 ignored; clippy `-D warnings` and fmt clean.
- `ioreport_energy_diagnostics` on the M5 Max after the fix: all stamps are in the past (GPU Energy 0.6 to 0.8 ms old) and CPU readings hold between batches, so the bound rejects no real timestamps.

No CRITICAL or HIGH findings remain.

PR #413, opened 2026-09-11
update: cut per-tick collection cost on macOS

### LINKED ISSUE #411: perf: cut per-tick collection cost on macOS (disk re-enumeration, per-PID ps spawn, IOReport/SMC overhead)
## Problem / Background

On an idle Apple M5 Max (macOS 27.0, all-smi 0.26.3 at main 720b8c2) a release build of `all-smi local --interval 1` holds 3.4 to 4.0 % of one core at steady state and burns 1.30 s of CPU time in its first ~5 s; `all-smi api --port 19099 --interval 1` shows 3.3 to 3.6 % and 1.31 s. Sampled with `top -l 4 -s 8 -pid`; child `ps` processes spawned at startup are not included in the steady-state figure.

| mode | steady-state CPU (one core) | RSS | threads | CPU time at startup |
|---|---|---|---|---|
| `all-smi local --interval 1` | 3.4 to 4.0 % | 12 to 13 MB | 13 to 14 | 1.30 s in the first ~5 s |
| `all-smi api --port 19099 --interval 1` | 3.3 to 3.6 % | 8.7 MB | 6 to 7 | 1.31 s in the first ~5 s |

Per-stage cost of one local-mode tick (release, first tick excluded, 11 ticks, measured with a temporary ignored integration test that builds the readers via `get_gpu_readers` / `get_cpu_readers` / `get_memory_readers` / `create_chassis_reader` and replays `LocalCollector::collect_steady_state` stage by stage):

| stage | avg | min | max | note |
|---|---|---|---|---|
| `Disks::new_with_refreshed_list()` (storage) | 37.2 ms | 19.3 ms | 51.3 ms | debug build spiked to 309 ms idle and 419 ms under load; host has 16 mounts including an SMB Time Machine share and local snapshots |
| `gpu.get_gpu_info()` (Apple Silicon native) | 15.9 ms | 13.1 ms | 19.2 ms | this is where `NativeMetricsManager::collect_once` runs; CPU and chassis readers then hit its 500 ms cache |
| process refresh, full (every 5th tick) | 13.5 ms | 13.1 ms | 14.0 ms | `refresh_processes_specifics(All)` |
| process refresh, selective (top 500 PIDs) | 5.4 ms | 4.3 ms | 6.9 ms | |
| `update_process_cache` on full-refresh ticks | 56 to 94 ms | | | every newly seen PID spawns `ps -p <pid> -o pri,nice` |
| `update_process_cache` on selective ticks | 0.4 ms | 0.37 ms | 0.57 ms | |
| `cpu.get_cpu_info()` | 34 us | | | native data comes from the cache |
| memory, chassis, gpu processes, vgpu, mig, merge+sort | under 0.4 ms each | | | |

First tick (cold): `update_process_cache` took 2.69 s in release (1070 processes, one `ps` spawn each), `cpu.get_cpu_info` 143 ms (100 ms warm-up sleep in `ensure_cpu_refreshed` plus sysctl), `gpu.get_gpu_info` 90 ms, storage 92 ms. The debug build's first tick was 1.87 s idle and 5.0 s under an 8-thread load.

The steady-state stages sum to roughly 60 to 75 ms per second, matching the top reading. Storage enumeration is about half, the native metrics manager about a quarter, process refresh plus cache the rest.

Inside `NativeMetricsManager::collect_once` (debug build, per call):

| step | avg |
|---|---|
| `IOReportCreateSamples` on the merged subscription (383 items) | 5.5 ms |
| same, Energy Model group alone (364 channels, of which we use about 6) | 5.2 ms |
| same, CPU Stats alone (130 channels) | 4.5 ms |
| same, GPU Stats alone (140 channels) | 0.35 ms |
| `IOReportCreateSamplesDelta` + iterating 383 items with 4 `CFString` conversions each | 1.1 ms |
| `SMCMetrics::collect` (opens a new SMC connection via `SMC::new()` every call, reads 8 static temp keys, GPU temp keys, PSTR, fans) | 5.8 ms (max 8.8 ms) |
| `get_thermal_state` | 16 us |

## Current Behavior

1. `src/view/data_collection/local_collector.rs:675-677` `collect_storage_info` calls `Disks::new_with_refreshed_list()` every tick, re-enumerating the mount table and running statfs on every mount, network shares included, on the collection critical path. `src/api/collection_loop.rs:64` already keeps one `Disks` instance and only refreshes it (`collect_storage_info_from` at line 216 takes `&Disks`), so the two modes differ for no reason. `src/storage/reader.rs:98` does the same fresh enumeration.
2. `src/device/process_list.rs:199` `get_process_priority_nice` on macOS (lines 233-250) runs `ps -p <pid> -o pri,nice` through `new_command` for every PID not yet in the cache: about 1070 spawns at startup (2.7 s of wall time on the first tick) and one spawn per newly discovered PID on every full-refresh tick (`FULL_REFRESH_INTERVAL = 5`, `local_collector.rs:60`). The Linux branch reads `/proc/<pid>/stat`.
3. `src/device/macos_native/ioreport.rs:166` `build_merged_channels` subscribes to the whole Energy Model group (364 channels on M5 Max, 300 of them `DTL` per-core telemetry we never read) and the whole CPU Stats group. Sample cost scales with channel count. The per-item loop around line 1109 converts 4 `CFString`s per item per tick.
4. `src/device/macos_native/smc.rs:980-983` `SMCMetrics::collect` calls `SMC::new()` (IOServiceMatching + IOServiceOpen) on every collection and drops it afterwards; `manager.rs:279-290` and `manager.rs:436` call it once per `collect_once`. `SmcConnection` (`smc.rs:905`) already caches an open handle for the Intel reader.
5. `src/device/cpu_macos.rs:783-787` `ensure_cpu_refreshed` sleeps 100 ms on the first call, and the first `collect_once` pays another 100 ms `get_sample` window; with the `ps` storm this makes the first frame appear late.

## Proposed Solution

- Storage: add a `disks: Disks` field to `LocalCollector`, call `disks.refresh_specifics(false, DiskRefreshKind::nothing().with_storage())` per tick, and call `refresh_list()` only every 30 s (tick counter, same pattern as `FULL_REFRESH_INTERVAL`) or when the mount table changes. Extract one shared helper (for example `src/storage/disks.rs` with a `DiskCache` that owns the `Disks` and the list-refresh cadence) and use it from `local_collector.rs`, `collection_loop.rs`, and `storage/reader.rs:98` so the three call sites stop diverging. Skip network filesystems (`disk.file_system()` in `smbfs`, `nfs`, `afpfs`, `webdav`) from the per-tick statfs, or move them to a `spawn_blocking` task with a timeout, so an SMB share cannot stall the tick; a stale value for those mounts is acceptable.
- Process priority/nice: replace the `ps` spawn in `process_list.rs:233-250` with `libc::getpriority(libc::PRIO_PROCESS, pid)` for nice (reset `errno` before the call because -1 is a valid return) and `libc::proc_pidinfo(pid, PROC_PIDTASKINFO, 0, &mut proc_taskinfo, size)` reading `pti_priority` for the scheduler priority, behind the existing `fn get_process_priority_nice(pid: u32) -> (i32, i32)` signature and the same `(20, 0)` fallback. `libc` is already a dependency. Both calls take microseconds and spawn nothing. Deferring the lookup to display time for the top N rows only is optional and secondary.
- IOReport: in `build_merged_channels` filter the Energy Model and CPU Stats subscription dictionaries to the channels we consume (`CPU Energy` family, `GPU Energy`, `ANE*`, `DRAM*`, the per-cluster performance-state channels, `GPUPH`) before merging; cache channel group/subgroup/name/unit strings per channel index on first sample instead of converting 4 `CFString`s per item per tick. This is the same code the exact-match fix in #410 touches, so land it after or together with #410 and keep the exact-match names as the filter source.
- SMC: give `NativeMetricsManager` one long-lived `SmcConnection` (reusing `smc.rs:905`) and make `SMCMetrics::collect` take `&mut SmcConnection`; read fans and PSTR every few ticks (5 s) and reuse the last values in between, temperatures every tick.
- Startup: leave the 100 ms CPU warm-up in place for now; the first-frame latency target is met by removing the `ps` storm. Note in the PR if the warm-up can be overlapped with the first IOReport window.
- Add a permanent `#[ignore]` benchmark test `tests/perf_tick_stages.rs` that constructs the readers the same way the probe did and prints per-stage timings, runnable with `cargo test --release --test perf_tick_stages -- --ignored --nocapture`, so regressions are measurable on any machine.

## Scope

**In scope:** `src/view/data_collection/local_collector.rs`, `src/api/collection_loop.rs`, `src/storage/reader.rs` (plus a new shared disk helper), `src/device/process_list.rs`, `src/device/macos_native/{ioreport,smc,manager}.rs`, `tests/perf_tick_stages.rs`.

**Out of scope:** IOReport energy correctness on M5 Max (#410); changing process refresh cadence or `MAX_DISPLAY_PROCESSES`; Linux or Windows priority paths; any change to displayed values.

## Implementation Notes

- **Reuse**: `collect_storage_info_from(&Disks, hostname)` in `collection_loop.rs:216` is already the shape the shared helper should have; `filter_docker_aware_disks` must keep running on the cached list. `SmcConnection` already handles reconnect on failure. `new_command` stays for other callers.
- **Constraints**: no change in displayed data or Prometheus output; `api` mode must pick up the same helper; the IOReport filter must not drop any channel that #410's exact-match table reads.
- **Edge cases**: `getpriority` on a PID that exited returns -1 with `errno = ESRCH`, fall back to `(20, 0)`; `proc_pidinfo` on a protected PID returns 0 bytes, fall back the same way; a mount that appears between list refreshes shows up within 30 s; a hung network mount must not block the other panels; SMC open failure leaves temperatures and fans empty as today.
- **Error handling**: no new logging on the hot path; a failed `refresh_list` keeps the previous list; a failed SMC reconnect is retried on the next tick.

## Acceptance Criteria

- [x] Steady-state CPU at `--interval 1` on the M5 Max, measured with `top -l 4 -s 8 -pid` alternating with the previous build in the same 160x50 pty: `all-smi api` drops from 3.3 to 3.8 % to at most 1.5 % of one core (measured 1.0 to 1.4 %), and `all-smi local` to about half of its baseline (5.5 to 6.6 % to 2.6 to 3.5 %; most of the rest is TUI rendering). No change in displayed data except the documented PRI semantics and missing SMC keys reading as absent instead of 0.
  - Corrected after measurement (#413). The original target was 1.5 % for `local`. The per-tick cost left is `IOReportCreateSamples` (about 9 ms per call regardless of channel count), SMC temperature reads (about 5 ms for 24 sensors) and process refresh (about 8 ms selective, 19 ms full), all outside this issue's changes.
- [x] No child processes are spawned during steady-state collection on macOS (verify with `sudo fs_usage -w -f exec` or by counting `ps` in `ps -ef` during a run); first-tick process collection for about 1000 processes completes in under 300 ms.
- [x] Storage stage under 5 ms per tick with the disk list refreshed at most every 30 s; a network mount that hangs does not delay the other panels.
- [x] `collect_once` at the real 1 s cadence in a release build on the M5 Max drops from about 22 ms to about 15 ms (measured 14.7 ms average over 11 ticks), with the floor broken out: `IOReportCreateSamples` 9.1 ms, SMC 5.0 ms, delta and parsing 0.6 ms.
  - Corrected after measurement (#413). The original target was under 5 ms. Filtering the subscription from 383 to 24 channels only moved `IOReportCreateSamples` from 10.2 to 8.7 ms per call, so 5 ms is not reachable without sampling IOReport less often.
- [x] A unit test asserts the macOS `get_process_priority_nice` path never reaches `new_command` (for example by returning a real value for the current PID without any `ps` on `PATH`).
- [x] `tests/perf_tick_stages.rs` exists as an `#[ignore]` test and its output is pasted in the PR.
- [x] `api` mode uses the shared disk helper and the same reader changes, so it benefits equally.
- [x] Integrated into the real code flow: both `local` and `api` modes run the new paths, not a standalone module.

## Verification

```bash
cargo build --release
cargo fmt --check && cargo clippy --all-targets
cargo test
cargo test --release --test perf_tick_stages -- --ignored --nocapture
sudo ./target/release/all-smi local --interval 1 &  # then, in another shell:
top -l 4 -s 8 -pid "$(pgrep -n all-smi)"            # expect at most 1.5 % steady state
sudo fs_usage -w -f exec | grep -c ' ps '            # expect 0 during steady state
```

## Related

- #410 (IOReport energy correctness on M5 Max; same code area for the subscription filter)
- `src/view/data_collection/local_collector.rs`, `src/api/collection_loop.rs`, `src/storage/reader.rs`, `src/device/process_list.rs`, `src/device/macos_native/{ioreport,smc,manager}.rs`, `src/device/cpu_macos.rs`

inureyes:
Collection on macOS stops re-enumerating the mount table every tick, stops spawning `ps` per new PID, and does less IOReport and SMC work per collection. Nothing is spawned after startup.

- **Storage**: `storage::DiskCache` backs the view collector, the API loop and `LocalStorageReader`. The mount table is enumerated on a background thread at most every 30 s; each tick only refreshes capacities. On macOS sysinfo's per-disk refresh returns a CFURL-cached value, so each enumeration records an anchor (exact available space plus `statfs` free space) and each tick reports the anchor moved by the `statfs` change, clamped to the total. Writes and deletes show on the next tick; purgeable-space changes and new mounts within 30 s. Network and cluster filesystems keep the last list's values. The collectors wait at most 2 s for the first list; `LocalStorageReader` still blocks for a complete one.
- **Priority (behavior change)**: nice from `getpriority`; PRI from `proc_pidinfo` `pti_priority` (the task's base priority) instead of the highest current thread priority `ps` reports, which the per-PID cache froze at first sight anyway. PRI is 20 when unreadable (other users' processes unless root).
- **IOReport**: the subscription keeps only the channels the parsers read (383 to 24 on an M5 Max).
- **SMC**: one kept connection, reopened on the next tick after a failure; system power and fans every 5 s; key info cached per key. A missing key now reads as absent instead of 0, so a Mac without `PSTR` omits chassis system power (Apple Silicon and Intel) instead of reporting 0 W. `SMCMetrics::collect()` is unchanged; `collect_from` takes an open connection.
- `tests/perf_tick_stages.rs` and `NativeMetricsManager::last_collection_timings()`.

Whole process at `--interval 1` on an M5 Max, `top -l 4 -s 8 -pid`, alternating with `c5bf759` in a 160x50 pty:

| mode | c5bf759 | this PR |
|---|---|---|
| `api` | 3.3 to 3.8 % | 1.0 to 1.4 % |
| `local` | 5.5 to 6.6 % | 2.6 to 3.5 % |

Collection only (benchmark below, before in parentheses): tick 26.5 ms (55 ms), first tick 270 ms (2.0 s), storage 0.013 ms (26 ms), `collect_once` 14.7 ms (22 ms). The remaining floor is `IOReportCreateSamples` at about 9 ms per call regardless of channel count, SMC temperature reads at about 5 ms, and process refresh. The #411 criteria for `local` CPU and `collect_once` were corrected to these measurements; follow-up options are sampling IOReport every 2 s, cheaper process refresh, and fewer SMC temperature reads.

Full suite (3595 passed), `cargo clippy --all-targets -- -D warnings` and fmt pass. A child-process watch (`proc_listchildpids` every 2 ms) saw 0 children after startup in `local` and `api`, against 616 at startup and 9 in steady state before. Linux and Windows are compiled only by CI.

`cargo test --release --test perf_tick_stages -- --ignored --nocapture`:

```text
reader construction: 203.103 ms
first tick: 1075 processes, 1 GPU rows, 1 storage rows

| first tick | time |
|---|---|
| gpu.get_gpu_info | 80.788 ms |
| cpu.get_cpu_info | 132.275 ms |
| storage (initial list) | 46.040 ms |
| process refresh (full) | 8.655 ms |
| update_process_cache | 1.889 ms |
| whole tick | 269.987 ms |

| steady-state stage (1 s cadence) | avg | min | max | ticks |
|---|---|---|---|---|
| gpu.get_gpu_info (runs collect_once) | 14.810 ms | 13.871 ms | 16.807 ms | 11 |
| gpu processes + vgpu + mig | 0.002 ms | 0.001 ms | 0.003 ms | 11 |
| cpu.get_cpu_info | 0.070 ms | 0.030 ms | 0.084 ms | 11 |
| memory.get_memory_info | 0.010 ms | 0.008 ms | 0.011 ms | 11 |
| chassis.get_chassis_info | 0.005 ms | 0.004 ms | 0.006 ms | 11 |
| storage (DiskCache::storage_info) | 0.013 ms | 0.006 ms | 0.015 ms | 11 |
| process refresh (full, every 5th) | 18.608 ms | 17.608 ms | 19.608 ms | 2 |
| process refresh (selective) | 8.286 ms | 6.730 ms | 8.765 ms | 9 |
| update_process_cache (full ticks) | 0.619 ms | 0.604 ms | 0.634 ms | 2 |
| update_process_cache (selective ticks) | 0.782 ms | 0.725 ms | 0.805 ms | 9 |
| merge + sort + truncate | 0.673 ms | 0.547 ms | 0.728 ms | 11 |
| whole tick | 26.507 ms | 22.099 ms | 37.729 ms | 11 |

| collect_once breakdown | avg | min | max | ticks |
|---|---|---|---|---|
| IOReportCreateSamples (floor) | 9.068 ms | 8.449 ms | 9.897 ms | 11 |
| IOReport energy + delta + parse | 0.586 ms | 0.516 ms | 0.688 ms | 11 |
| SMC | 5.036 ms | 4.564 ms | 6.275 ms | 11 |
| thermal + assembly | 0.016 ms | 0.009 ms | 0.024 ms | 11 |
| collect_once total | 14.706 ms | 13.817 ms | 16.703 ms | 11 |

process CPU during steady-state ticks: 244.389 ms over 291.576 ms of tick work, 2.22 % of one core at 1 s

| mount table enumeration | avg | min | max | runs |
|---|---|---|---|---|
| DiskCache list refresh (background, every 30 s) | 32.649 ms | 30.742 ms | 33.655 ms | 5 |
| Disks::new_with_refreshed_list (old per-tick path) | 35.564 ms | 29.333 ms | 51.302 ms | 5 |
```

Closes #411


inureyes:

### Intent
> Cut the per-tick collection cost on macOS with no child processes in steady state, per #411 and the decisions recorded after measurement.

### Findings Addressed
- [x] `SMCMetrics::collect()` is public library API and had become `collect(&mut SMC)`. The no-argument form is back with its old behavior; the kept-connection variant is `collect_from`, used by `SmcSampler` (MEDIUM, bfea1a5)
- [x] `LocalStorageReader::get_storage_info()` could return an empty first list after a 2 s wait. It now builds its first list on the calling thread (`DiskCache::with_blocking_first_list`); the view and API loops keep the 2 s bound. Documented on both and in the docs (MEDIUM, dbb8a72)
- [x] The SMC key-info cache latched any result other than 0x84. Only result 0 and 0x84 are cached now (LOW, bfea1a5)
- [x] After a failed SMC read, `SmcSampler` repeated a partial power and fan read for 5 s. The reopened connection now does a full read (LOW, bfea1a5)
- [x] On macOS, `statfs` on a mount point whose volume was unmounted reads the parent filesystem. The read now requires `f_mntonname` to match (LOW, dbb8a72)

### Remaining Items
- [ ] On Linux the same unmounted-directory case goes through sysinfo's per-disk `statvfs` and can show the parent filesystem for up to 30 s (LOW), left as is because sysinfo owns that call

### Verification
- [x] All stated requirements implemented (the two restated criteria are reported as measured in the PR body)
- [x] Tests pass: `storage` 30, `device::macos_native` 92, `snapshot_test` 13, `view::data_collection` 11, clippy on lib and tests, fmt. The full suite was not rerun after these two commits.

inureyes:

No critical or high findings. Fixed in 256f609:

- [x] **Medium**, `src/storage/disk_cache.rs` `NETWORK_FILE_SYSTEMS`: on Linux, sysinfo only drops cifs, nfs, nfs4 and autofs, so GPFS, BeeGFS, Weka and similar mounts still got a live `statvfs` every tick, and a hung one stalled both collection loops. Added gpfs, beegfs, wekafs, panfs, pvfs2, afs, fuse.ceph-fuse, fuse.juicefs, fuse.rclone, fuse.s3fs and fuse.gcsfuse. These now keep the capacity from the last list.
- [x] **Low**, `statfs_free_bytes`: the `f_mntonname` check no longer uses `unsafe CStr::from_ptr`. It compares bytes up to the first NUL, bounded by the array.
- [x] **Low**, `DiskCache::update_listing`: the 30 s interval now counts from when the list arrived, not when enumeration started, so a slow enumeration (a dead SMB share) is not followed straight away by another.

Checked, no change needed: `getpriority` errno handling and PID 0, the `proc_pidinfo` size check against the kernel's 96-byte `proc_taskinfo`, CF ownership on every path of `retain_channels`, the SMC key cache (bounded, dropped with the connection), `convert_value` bounds, lock order, and at most one enumeration thread per cache.

Not fixed (low, existed before this PR):

- On Linux, a volume unmounted between list refreshes reads its parent filesystem for up to 30 s.
- The view collector joins the storage task without a timeout, so a hung filesystem type missing from the list still stalls the tick.
- When the SMC cannot be opened, every collection retries the open. The issue asks for this.

Verification: fmt, `cargo clippy --lib --tests -- -D warnings`, `cargo test --lib storage` (31 passed), `cargo test --bin all-smi view::data_collection` (11 passed).

inureyes:
## PR Finalization Complete

### Summary
- **Tests**: verified adequate, no additions needed. `storage::disk_cache` covers the anchor arithmetic (write/delete/clamp/anchor-replace), network-filesystem classification, list-refresh cadence, hung/failed enumeration, and both first-list modes; `device::process_list::priority_macos` structurally pins that the macOS lookup never reaches `new_command` plus behavioral tests for a reniced child, an exited PID, and PID 0; `device::macos_native::smc` covers the key-info cache (hit, absent-key, non-definitive-result), `collect()` vs `collect_from`, and `get_or_retry` vs `get`; `device::macos_native::ioreport::channel_filter` checks the energy/CPU/GPU keep-predicates against a recorded 383-channel M5 Max inventory.
- **Documentation**: fixed one inconsistency. `docs/LIB_mode.md`'s `ProcessInfo` table documented `priority`'s macOS source and 20-fallback but left `nice_value` as a bare "Nice value" even though this PR moved it to `getpriority`; added the same treatment (source plus the 0 fallback for an already-exited process). `docs/ARCHITECTURE.md` and the rest of `docs/LIB_mode.md` already described the disk cache's anchor-plus-statfs-delta semantics, the IOReport channel filter, and the kept SMC connection accurately; `README.md`, `API.md`, and `docs/man/all-smi.1` do not describe the PRI/NI columns or the SMC absent-key/0 W distinction at all, so there was nothing stale to fix there. No Korean counterparts of these docs exist in the repository.
- **Lint/Format**: `cargo fmt --check` and `cargo clippy --lib --tests -- -D warnings` were already clean; no fixes needed.

### Verification
- `cargo fmt --check`: clean
- `cargo clippy --lib --tests -- -D warnings`: clean
- `cargo test --lib storage`: 31 passed
- `cargo test --lib device::macos_native`: 92 passed, 1 ignored (hardware diagnostic)
- `cargo test --lib device::process_list`: 9 passed
- `cargo test --lib api::`: 125 passed
- `cargo test --bin all-smi view::data_collection`: 11 passed, 1 ignored (release-only measurement harness)
- `cargo test --test perf_tick_stages -- --list`: compiles, lists the ignored benchmark

PR #416, opened 2026-09-12
fix: Rebellions NPU devices vanish under load, plus four silent parsing/routing defects

hoyajigi:
# fix: Rebellions NPU — devices vanish under load, plus four silent parsing/routing defects

## The headline: NPUs disappear from all-smi the moment a workload runs

`RblnContext` declared four fields that do not match what `rbln-stat --json` actually emits. Captured verbatim from an ATOM Plus node running vLLM:

```json
{"ctx_id": "10001", "npu": 0, "process": "VLLM::EngineCore",
 "pid": "2733390", "priority": "normal", "ptid": "0",
 "memalloc": "3.5GiB", "status": "idle", "util_info": "1.1"}
```

| struct declared | tool emits | |
|---|---|---|
| `npu: String` | `0` — integer | type mismatch |
| `pid: u32` | `"2733390"` — string | type mismatch |
| `cmd: String` | no such key; it is `process` | missing field |
| `memory: String` | no such key; it is `memalloc` | missing field |

`#[serde(default)]` on the `contexts` field only covers an **absent key** — it does not rescue an element that fails to deserialize. So the entire `RblnResponse` parse fails, `get_npu_info_internal` takes its `Err(_) => return Vec::new()` path, and **every NPU on the node vanishes from all-smi**. Not just the process list — the devices themselves.

This is why it was never caught: an **idle** node parses fine and looks perfectly healthy. The tool only breaks once the hardware is actually being used.

Verified empirically — the first new regression test fails on the previous code with:

```
Error("invalid type: integer `0`, expected a string", line: 34, column: 14)
```

Also fixed here: per-context `memalloc` is a human-readable string (`"3.5GiB"`, `"72.0MiB"`) unlike the device-level bare-byte fields, so the memory parser learned the IEC/SI suffixes (and now rejects out-of-range values explicitly, since `as u64` saturates rather than wrapping). Each context is joined back to its owning device so `device_uuid` carries the real UUID instead of the bare index.

> Note on ATOM Max: `npu` repeats once per card there (32 device entries, four distinct `npu` values), and `rbln-stat`'s JSON exposes no per-context card key, so the context→device join is inherently ambiguous on that platform. Documented in a code comment rather than papered over.

---

*The four defects below were the original scope of this PR and are unchanged.*

## Summary

all-smi's Rebellions (RBLN) support has four defects that each fail *silently*:
nothing errors, nothing logs at the user's level, and the output looks plausible.
On a real 8-card ATOM Plus node the tool today reports **15.72 PiB of memory per
card**, **0.0 W of power on every card**, and **zero `all_smi_rebellions_*`
metrics on the Prometheus endpoint**. Device detection additionally carries the
wrong PCI vendor ID, which is currently masked by other checks.

All four were diagnosed against captured output from a live node
(8 x RBLN-CA22, KMD/driver 3.0.0) and each is now pinned by a regression test
built from that verbatim output.

### 1. Memory read as mebibytes when the tool reports bytes

`rbln-stat --json` reports memory as a **bare byte count with no unit suffix**:

```json
"memory": { "used": "0", "total": "16877879296" }
```

The reader passed that through the shared `parse_memory_mb_to_bytes()` helper,
which strips `MB`/`MiB` and multiplies by 1 MiB. `16877879296` therefore became
`17,697,739,160,682,496` bytes.

**User-visible symptom:** every Rebellions card reports **15.72 PiB** of total
memory (true value: 15.72 GiB) in the TUI, the API and the Prometheus endpoint.
The number is large but does not overflow `u64`, so nothing warns; utilisation
percentages computed against it are meaningless.

**Fix:** a Rebellions-local `parse_rbln_memory_bytes()` that reads the value as
bytes. `parse_memory_mb_to_bytes()` is deliberately **not** changed — other
vendor readers depend on its MB semantics. An explicit `MB`/`MiB` suffix is
still honoured, so an SDK release that starts labelling the unit will not be
misread as bytes.

### 2. Power read as 0.0 W because the value is in microwatts

`card_power` is reported in microwatts: `"card_power": "17521800uW"`.
`parse_power()` did `trim_end_matches(['W', ' '])`, leaving `"17521800u"`, which
fails `parse::<f64>()`. The caller's `unwrap_or_else` fallback then substituted
`0.0`.

**User-visible symptom:** **every** Rebellions power reading all-smi has ever
produced is `0.0 W` (true idle value: ~17.5 W). Power panels, the
`all_smi_gpu_power_consumption` series and any alerting built on them are flat
zero, indistinguishable from a genuinely idle card.

**Fix:** `parse_power()` now understands SI prefixes (`u` / `µ` / `μ`, `m`,
`k`/`K`). This is the shared helper, so the change is deliberately
behaviour-preserving: **no input that parsed successfully before changes
result**, because no `f64` literal ends in one of those characters. The existing
`"123W"`, `"123 W"`, `"150/250W"` and `"invalid"` cases are pinned by tests. One
input that previously returned `None` now parses: `"150W/250W"` (unit on both
halves of a current/limit pair).

### 3. No `all_smi_rebellions_*` metric was ever emitted

Both the fast path in `npu/mod.rs` and `RebellionsExporter::can_handle()` tested
`info.name.contains("Rebellions")`. Real hardware reports
`"name": "RBLN-CA22"`. Since the fast path and the `can_handle` fallback used
the same predicate, the fallback could not rescue it.

**User-visible symptom:** the Prometheus endpoint on a Rebellions node exposes
the generic NPU series and **not one** `all_smi_rebellions_*` series. Any
dashboard or alert keyed on firmware, KMD version, P-state or device status is
permanently empty, with no error anywhere.

**Fix:** routing now goes through `rebellions::is_rebellions_device()`, shared by
the fast path and `can_handle`. It prefers the vendor tag the reader already
stamps into `detail["lib_name"] == "RBLN-SDK"` — independent of marketing
names — and falls back to the `RBLN` product prefix (not a specific SKU) for
devices that arrive without that tag, such as remote nodes parsed from
Prometheus text and the mock server. The mock server's `"Rebellions ATOM"` name
keeps working.

The exporter pool in `npu/mod.rs` is indexed by hardcoded positions, so the
`vec!` order was left untouched, and a test now pins the resolved vendor for
Gaudi, Rebellions, Furiosa, Google TPU and (on Linux) Tenstorrent.

### 4. Wrong PCI vendor ID in device detection

`platform_detection.rs` looked for vendor ID `1f3f`. Rebellions' real PCI vendor
ID is **`1eff`**, confirmed two ways on the captured node:

```
03:00.0 Processing accelerators [1200]: Rebellions Inc. RBLN-CA22 (PF) [1eff:1220] (rev 03)
```

and all 26 PCI aliases in the `rebellions` KMD (`modinfo`), every one of them
`pci:v00001EFF...`.

**User-visible symptom:** latent today — `/dev/rbln0` is checked first, and the
same `lspci` line also matches the literal `"Rebellions"`. It would bite on a
host with the driver loaded but an old `pci.ids` that cannot name the vendor,
where the line reads `Device [1eff:1220]`.

**Fix:** the constant and its comment now say `1eff`, the vendor-name check is
kept as a second signal, and the probe is invoked as `lspci -nn` so the numeric
ID is actually present in the output it searches (with bare `lspci` the ID is
never printed, so the vendor-ID branch could not fire at all).

### Also fixed: exporter detail keys (bug 3 would otherwise deliver nothing)

With routing fixed, the exporter still emitted nothing, because it looked up
detail keys the reader never writes: `firmware_version`, `kmd_version`,
`device_name`, `serial_id`, `performance_state`, `status`. The Rebellions reader
writes the project's Title Case convention (`Firmware Version`, `KMD Version`,
`Serial ID`, `Performance State`, `Status`) — the same convention used
throughout `readers/detail_keys.rs`. The lookups now match the producer.

The `all_smi_rebellions_device_info` `location` label also carried the constant
`"5"` with the comment *"Default location from mock server"*. The driver-reported
`location` field is now surfaced by the reader and used for that label, so the
value is real rather than a copied placeholder.

## How it was verified

Diagnosis and every test fixture come from read-only captures of a live
Rebellions node (8 x RBLN-CA22 / ATOM Plus, KMD 3.0.0): `rbln-stat --json`,
`rbln-stat -g -j`, `lspci -nn`, `lspci -vv`, `modinfo rebellions` and the sysfs
tree. Test inputs are inlined verbatim string literals, following the house
pattern in `src/device/hlsmi/parser.rs`; no test reads a file at runtime.

Each defect was reintroduced one at a time against the new test suite to confirm
the tests actually catch it. With all four (plus the detail-key fix) reverted,
exactly these nine tests fail:

```
api::metrics::npu::rebellions::tests::real_hardware_is_recognised
api::metrics::npu::rebellions::tests::vendor_metrics_are_emitted_for_a_real_device
api::metrics::npu::tests::a_real_rebellions_card_routes_to_the_rebellions_exporter
api::metrics::npu::tests::a_rebellions_scrape_contains_vendor_metrics
device::common::parsers::tests::test_parse_power_si_prefixes
device::platform_detection::tests::rebellions_is_detected_by_its_real_pci_vendor_id
device::readers::rebellions::tests::card_power_microwatts_are_read_as_watts
device::readers::rebellions::tests::memory_is_read_as_bytes_not_mebibytes
device::readers::rebellions::tests::memory_parser_accepts_bare_bytes_and_explicit_suffixes
```

## What the tests now pin

`src/device/readers/rebellions.rs` had **no tests at all**; it now has ten.
20 tests added in total.

| Test | Pins |
|---|---|
| `memory_is_read_as_bytes_not_mebibytes` | `total` parses to exactly `16 877 879 296` bytes, and asserts the value stays under 1 TiB |
| `card_power_microwatts_are_read_as_watts` | 17.5218 W and 18.307255 W for the two captured cards, and explicitly that power is not 0.0 |
| `memory_parser_accepts_bare_bytes_and_explicit_suffixes` | bare bytes, `MB`/`MiB` suffixes, garbage, and no panic on a value that would overflow when scaled |
| `temperature_and_utilization_are_read` | 31 C / 33 C, 0.0 % util, `device_type == "NPU"`, `name == "RBLN-CA22"` |
| `real_output_deserializes`, `grouped_output_deserializes_despite_the_extra_group_id` | the serde struct still accepts real output, including the extra `group_id` from `rbln-stat -g -j` |
| `detail_carries_the_keys_the_exporter_reads`, `cached_static_info_matches_the_uncached_path` | the reader/exporter detail-key contract, on both the cached and uncached paths |
| `a_real_rebellions_card_routes_to_the_rebellions_exporter` | an `RBLN-CA22` device resolves to the Rebellions exporter, with and without the `lib_name` tag, plus a hypothetical later SKU and the mock server's name |
| `the_other_vendors_still_route_to_their_own_exporters` | the hardcoded exporter-pool indices, for every vendor incl. Tenstorrent on Linux |
| `vendor_metrics_are_emitted_for_a_real_device`, `a_rebellions_scrape_contains_vendor_metrics` | all five `all_smi_rebellions_*` metrics are present, with real `firmware`, `pstate`, `sid`, `location` and `status` label values |
| `test_parse_power`, `test_parse_power_si_prefixes`, `test_parse_power_unit_on_both_halves` | existing `parse_power` inputs are unchanged, and `uW`/`µW`/`μW`/`mW`/`kW` now convert correctly |
| `rebellions_is_detected_by_its_real_pci_vendor_id` | `1eff` matches (including with the vendor name stripped), `1f3f` does not, other vendors do not |

## Validation

```
cargo fmt --all                          # clean
cargo fmt --all -- --check               # exit 0
cargo clippy --all-targets -- -D warnings # exit 0, no warnings
cargo test                               # 3635 passed, 0 failed (27 suites)
```

Baseline on `main` before this change was also fully green, so there are no
pre-existing failures to separate out. Built and tested on macOS
(aarch64-apple-darwin); the Linux-only Tenstorrent routing assertion is behind
`#[cfg(target_os = "linux")]` and was not executed locally.

## Not changed (deliberately)

- **`parse_memory_mb_to_bytes()` semantics** — other vendor readers depend on
  the MB interpretation. The Rebellions byte parsing lives in the Rebellions
  reader.
- **The exporter-pool `vec!` order and its hardcoded indices** — inserting into
  the middle would silently mis-route other vendors. Left as-is and now covered
  by tests.
- **`API.md`'s Rebellions metrics table** — it already documents metrics that do
  not exist in the code (`all_smi_rebellions_device_status`,
  `..._performance_state`, `..._pcie_generation`, `..._memory_bandwidth_gbps`,
  `..._compute_tops`) and labels that do not match the implementation. This
  drift predates the change; now that the metrics are actually emitted it
  becomes visible to users, and deserves its own PR that decides per row whether
  to implement or delete.
- **The `model` label heuristic** (`ATOM Max` / `ATOM+` / else `ATOM`) — it
  matches on marketing names that real hardware never reports, so an RBLN-CA22
  is labelled `ATOM`. Mapping SKU codes to product families needs a source of
  truth for the full SKU list, which this PR does not have.
- **Process/context memory unit** — `contexts[]` was empty on the captured node,
  so there is no ground truth for its unit. It now goes through the same parser
  (bare number = bytes, consistent with the device memory in the same document)
  while still honouring an explicit `MB`/`MiB` suffix, so the previous
  interpretation survives if that is what the field actually uses.


inureyes:
## Hardening review

Target: #416. No linked issue references were present.

- Correctness: malformed individual `contexts[]` entries could still invalidate the full response; contexts now deserialize independently, preserving devices and valid process rows.
- Correctness/security: PATH discovery returned a relative command that the absolute-path guard rejected; it now uses the safe absolute path resolved by `which`.
- Correctness: suffixed memory values at the `u64` boundary now fail instead of saturating.
- Reliability: fallback routing now requires the RBLN prefix or vendor name, and KMD samples carry device identity so multi-device nodes do not emit duplicate label sets.
- Security review: command execution remains argument-vector based and validated; no new authorization, network, or secret-handling surface was introduced.
- Performance review: context processing remains O(n), with no new unbounded cache or repeated external command.

Validation: `cargo test rebellions --lib` (23 passed), `cargo clippy --all-targets -- -D warnings`, `cargo test` (all suites passed), and `cargo fmt --all -- --check`.

Reviewed PR: https://github.com/lablup/all-smi/pull/416

PR #423, opened 2026-09-19
fix: count Rebellions ATOM Max card power once per card

### LINKED ISSUE #418: Rebellions ATOM Max: total NPU power is overcounted 4x because per-card power is summed once per die
## Summary

On Rebellions ATOM Max, `all-smi` reports total NPU power at roughly 4x the real draw (measured 1369.7 W on a node whose NPUs actually draw 341.5 W). ATOM Plus is unaffected. The cause is that the Rebellions reader emits one device row per die while the power value it reads is a per-card figure, so every consumer that sums power counts each card four times.

## Background

`rbln-stat --json` enumerates dies, not cards. An 8-card ATOM Max node reports 32 devices (4 dies per card), and each die repeats its card's `card_power` value. ATOM Plus is one die per card, so summing per-die power happens to be correct there and the bug went unnoticed.

The reader confirms this. In `src/device/readers/rebellions.rs`, `get_npu_info_internal` iterates over `response.devices` (one entry per die) and calls `create_gpu_info_from_device` for each, which sets `power_consumption: parse_power_safe(&device.card_power)` (around line 396). One `GpuInfo` per die, each carrying the full per-card power.

That inflated value then flows into two surfaces:

1. TUI cluster summary: `total_power_watts` in `src/metrics/gpu_readings.rs` sums `power_consumption_reading` over every row unconditionally, and the result feeds `GpuClusterMetrics.total_power_watts` in `src/metrics/aggregator.rs`.
2. Prometheus: `all_smi_npu_power_watts` is emitted as one series per device, so `sum by (instance)` inflates by the same factor.

A side effect makes the average look right: `avg_power = total / reporting_count` = 1369.7 / 32 = 42.8 W, which coincidentally equals the true per-card power. So the average reads correctly while the total is 4x wrong.

Measured evidence from real captured output:

| | devices reported | physical cards | dies/card | total_power_watts | real NPU draw | overcount |
|---|---|---|---|---|---|---|
| ATOM Plus (RBLN-CA22) | 8 | 8 | 1 | 143.3 W | 143.3 W | 1.0x |
| ATOM Max (RBLN-CA25) | 32 | 8 | 4 | 1369.7 W | 341.5 W | 4.0x |

Per-die memory is not affected: it sums correctly today (32 x 15.72 GiB = 503 GiB, matching the hardware).

## Proposed Solution

A design decision is needed on how to model a multi-die card. Three options, from the report:

(a) Divide `card_power` by dies-per-card so each die carries a derived share. Total becomes correct, but it places a computed value in `power_consumption`, whose "absence" contract otherwise treats the field as directly measured.

(b) Report power on the first die of each card and leave it absent on the other three. `total_power_watts` already skips absent rows via `filter_map`, so the total becomes correct with all values measured. The cost is that 24 of 32 rows show no power, which can read as broken; the per-card value could still be carried in the `detail` map on every row for display.

(c) Collapse rows to cards (8 rows, memory summed per card). This is arguably the most faithful model, since the card is the unit the vendor's `rbln-stat` table prints power for and the unit Backend.AI allocates. The cost is that it changes the device count users see and requires summing per-die memory (which is correct as-is today).

Grouping key: `sid` (board serial) is the only portable grouping key that yields 8 groups on both Plus and Max. `group_id` collapses ATOM Plus to a single card, and `location` is a die-position-within-board code, not a card index, so neither works.

Maintainer input is requested on which option to adopt before implementation.

## Implementation Notes

- The per-die emission is in `create_gpu_info_from_device` and its caller `get_npu_info_internal` in `src/device/readers/rebellions.rs`. `card_power`, `sid`, and `location` are already parsed into `RblnDevice` there.
- Whichever option is chosen, the unconditional sum in `total_power_watts` (`src/metrics/gpu_readings.rs`) and its consumer `GpuClusterMetrics.total_power_watts` (`src/metrics/aggregator.rs`) should end up with a correct total without special-casing the vendor.
- Prometheus `all_smi_npu_power_watts` must not double-count once the fix lands.
- Option (b) relies on the existing absence encoding (`GPU_METRIC_UNAVAILABLE`) so that `filter_map` skips non-reporting rows.
- A captured `rbln-stat --json` from a real ATOM Max node is available to use as a test fixture.

## Acceptance Criteria

- On an 8-card ATOM Max node, the reported total NPU power matches the real draw (about 341.5 W in the captured sample), not 4x it.
- ATOM Plus behavior is unchanged (total stays 1.0x correct).
- Prometheus `all_smi_npu_power_watts` no longer double-counts per-card power when summed per instance.
- Per-die memory totals remain correct (no regression from any row collapsing).
- A regression test built from the ATOM Max fixture fails before the fix and passes after it.

---

## Summary

On Rebellions **ATOM Max**, `all-smi` reports NPU power at **4× the real draw** — 1369.7 W on a node whose NPUs actually draw 341.5 W.

`rbln-stat --json` enumerates **dies** (32 on an 8-card node), and `card_power` is a **per-card** value that all four dies of a card repeat. The reader emits one `GpuInfo` per die, so every consumer that sums `power_consumption` counts each card's power four times.

Measured from real captured output of an 8-card ATOM Max node (`RBLN-CA25`, KMD 3.0.0) and an 8-card ATOM Plus node (`RBLN-CA22`) for comparison:

| | devices reported | physical cards | dies/card | `total_power_watts` | real NPU draw | overcount |
|---|---|---|---|---|---|---|
| ATOM Plus | 8 | 8 | 1 | 143.3 W | 143.3 W | 1.0× |
| **ATOM Max** | **32** | **8** | **4** | **1369.7 W** | **341.5 W** | **4.0×** |

ATOM Plus is unaffected because it is one die per card, which is why this has gone unnoticed.

## Affected surfaces

1. **TUI cluster summary** — `src/metrics/gpu_readings.rs` sums unconditionally over every row:
   ```rust
   pub fn total_power_watts(gpus: &[GpuInfo]) -> f64 {
       gpus.iter().filter_map(GpuInfo::power_consumption_reading).sum()
   }
   ```
   consumed by `GpuClusterMetrics.total_power_watts` in `src/metrics/aggregator.rs`.
2. **Prometheus** — `all_smi_npu_power_watts` is emitted as one series per device, so `sum by (instance)` is inflated by the same factor.

Note a side effect: `avg_power = total / reporting_count` = 1369.7 / 32 = 42.8 W, which *coincidentally* equals the true per-card power. So the average reads correctly while the total is 4× wrong.

## How to identify a card

`sid` (board serial) is the only portable grouping key — it yields 8 groups on both Plus and Max. Two alternatives do **not** work:

- `group_id` (from `rbln-stat -g -j`) is `"1"` for all 8 devices on ATOM Plus, so grouping by it collapses Plus to a single card.
- `location` is a die-position-within-board code, not a card index — it is `5` for every device on Plus and cycles `1,2,3,4` within each card on Max.

Corroborating evidence that 8 is the card count: 8 distinct `sid` values (4 devices each), `rsd1..rsd8` in `/sys/class/rebellions/` each listing exactly its 4 `rbln` members, and 8 PCIe switches in `lspci -tv` each fronting 4 × `1eff:1250` endpoints.

## Possible fixes — I don't think this is mine to pick

| | total | per-row value | cost |
|---|---|---|---|
| **(a)** divide `card_power` by dies-per-card | correct | derived (~10.7 W) | puts a computed value in a field the absence contract treats as measured |
| **(b)** report power on the first die of each card, absent on the rest | correct (`filter_map` skips `None`) | all measured | 24 of 32 rows show no power, which reads as broken; the `detail` map could carry the card value on every row for display |
| **(c)** collapse rows to cards (8 rows, memory summed per card) | correct | all measured | loses per-die utilization/memory granularity |

**(c)** is arguably the most faithful model: the card is the coherent domain (the `topology` sysfs attribute is a 4×4 intra-card matrix), it is the unit Backend.AI allocates (`atom-max.device: 8`), and the vendor's own human-readable `rbln-stat` table prints Power once per 4-row block. But it changes the device count users see, and memory would need summing — per-die memory currently sums *correctly* (32 × 15.72 GiB = 503 GiB, matching the hardware), so that part is not broken today.

Happy to implement whichever you prefer, with an ATOM Max fixture and a test that fails before the fix. I have the captured `rbln-stat --json` from a real Max node to use as the fixture.

## Relationship to #416

Deliberately **not** folded into #416. That PR is a separate set of five silent-failure fixes, already green, and its tests are built from ATOM Plus output where this bug does not reproduce. This one needs a design decision plus a Max fixture.

inureyes:
## Summary

`rbln-stat` repeats each ATOM Max card's `card_power` on all four dies, and the reader counted it on every die, so an 8-card node (32 dies) read 4x its real draw. Per the maintainer's choice of option (b), one die per card reports the card's power and its siblings read absent. Rows stay one per die (option (c) rejected) and no per-die share is derived (option (a) rejected). ATOM Plus, one die per card, is unchanged.

## Behaviour

- Cards are grouped by `sid` only. Reporting die: lowest kernel index parsed from `device` (`rbln12` is 12); unparsed names last; ties by rbln-stat order. Position alone is not stable (the ATOM Plus capture lists devices out of index order), and the reporter must not hop between polls because its power series and energy-counter key are per uuid.
- Siblings read `GPU_METRIC_UNAVAILABLE`. An empty or missing `sid` makes the die its own card; a missing `sid` no longer fails the whole response.
- New vendor-neutral detail key `card_power_watts` (reporter's value, two decimals) on every die of a multi-die card. Interpretation the maintainer can flip: single-die cards, i.e. all of ATOM Plus, do not get it, since their `power_consumption` already shows it and a per-poll label would churn ATOM Plus `all_smi_gpu_info` series for users with no bug. Cost on ATOM Max: each die's `all_smi_gpu_info` series now churns every poll, as Tenstorrent's already do with live current and clock readings in `detail`.
- Local-view chassis power sums readings only (a raw sum adds 24 x -1 W). Dashboard Avg. Power and the aggregator divide by reporting rows via `gpu_readings::mean_power_watts`, N/A when none; Apple Silicon keeps its combined CPU+GPU+ANE figure. The TUI shows a sibling die's card value in parentheses, e.g. `(43W)`. The remote parser keeps `card_power_watts`.

## Notes

- Correction to the issue: Rebellions never emits `all_smi_npu_power_watts`. The per-device series is `all_smi_gpu_power_consumption_watts`, which already omits absent readings, so the exporter is unchanged. The key is not named `power`/`power_draw`, which would feed `all_smi_npu_power_watts`.
- Snapshot JSON/CSV and record/replay serialize non-reporting dies as `power_consumption: -1.0`, the existing sentinel contract; snapshot Prometheus omits them.
- Users-tab attribution clamps absent power to 0 W, so a process on a non-reporting die gets no card power.
- Untouched: `parse_power_safe`'s 0.0 fallback (an unparsable reporter value publishes no `card_power_watts`), and the ambiguous context-to-device join on ATOM Max.
- `tests/fixtures/rbln-stat-atom-max-synthetic.json` is synthetic, not a hardware capture (8 `sid`s x 4 dies, card powers summing to 341.5 W, in the ATOM Plus capture's exact shape). Replace it with a real ATOM Max capture when one is available.

## Fails before

Saved run of `cargo test --lib device::readers::rebellions` with the fixture and regression test, before the reader fix (excerpt):

```
test device::readers::rebellions::atom_max_tests::atom_max_counts_each_card_power_once ... FAILED
thread 'device::readers::rebellions::atom_max_tests::atom_max_counts_each_card_power_once' (9540989) panicked at src/device/readers/rebellions_atom_max_tests.rs:101:5:
total NPU power 1366 W, expected 341.5 W
test result: FAILED. 15 passed; 1 failed; 0 ignored; 0 measured; 1639 filtered out; finished in 0.01s
```

## Validation (local, macOS)

- Reader tests: 25 passed, 10 new (regression, reordering, ATOM Plus unchanged, per-die memory, missing/empty `sid`, `/metrics` render with 8 power series summing to 341.5 and no sentinel or `all_smi_npu_power_watts`, remote round trip at 341.5 W).
- On the committed tree, `cargo test --lib` filtered to the reader and every touched module: 154 passed; `cargo test --bin all-smi` filtered likewise plus the local collector test: 93 passed.
- `cargo test --test user_facing_text_test`: 5 passed. `cargo clippy --lib --tests -- -D warnings` and `cargo fmt --check` clean; the pre-commit hook's `cargo clippy --all-targets --all-features -- -D warnings` passed on both commits.
- Unfiltered unit runs on the final tree: `cargo test --bin all-smi` 1849 passed; `cargo test --lib` 1667 passed and 2 failed, both `storage::disk_cache` timing tests in code this PR does not touch, which pass when rerun alone and in the bin target.
- CI on 9e68dbd: Build Check, Test Suite, launchd and systemd smoke tests, Packaging Sync Check and CLA all passed.
- Not verified: no ATOM Max or ATOM Plus hardware here, and the ATOM Max fixture is synthetic.

Closes #418


inureyes:

### Intent
> Count each Rebellions ATOM Max card's power once (341.5 W on the 8-card reference, not 1366 W) by grouping dies on `sid`, leaving ATOM Plus unchanged.

### Findings Addressed
No CRITICAL or HIGH findings; no fix commits were needed.

### Remaining Items
- [ ] Users-tab attribution clamps absent power to 0 W (`src/ui/aggregation/user.rs:222`), so a process on a non-reporting die gets no card power. Inherent to one-reporter-per-card and already stated in the PR body; follow-up candidate. (LOW)
- [ ] Sort by power compares raw `power_consumption` (`src/app_state.rs:799`), so the 24 non-reporting dies sort below the reporters. Existing behavior for every unavailable row, not introduced here. (LOW)
- [ ] Dashboard Avg. Power keeps the Apple Silicon branch instead of `mean_power_watts`, because that total is the combined CPU+GPU+ANE figure; switching would silently change the cell to GPU-only power. Deliberate and noted in the PR body. (LOW)
- [ ] `card_power_watts` on `all_smi_gpu_info` creates a new info series per die each time the card value changes. This is the maintainer's chosen trade-off, and the PR body states the cost. (accepted)

### Verification
- [x] Tests pass

Checked locally: filtered lib tests for the reader, detail_keys, gpu_readings, aggregator, metrics_parser, gpu_renderer and dashboard (154 passed); `cargo test --bin all-smi view::data_collection::local_collector` (6 passed, including `inject_gpu_power_skips_unavailable_rows`); `cargo test --test user_facing_text_test` (5 passed); `cargo clippy --lib --tests -- -D warnings` and `cargo fmt --all -- --check` clean. The fixture matches the specified shape (32 dies, 8 sids, 4 dies each, card values summing to 341.5 W), and the saved pre-fix run shows 1366 W. CI Test Suite passed; Build Check was still pending when I checked. Not verified: no ATOM Max or ATOM Plus hardware here, and the ATOM Max fixture is synthetic.

inureyes:

No CRITICAL or HIGH findings, so no fixes were pushed. Remote `card_power_watts` is ANSI-stripped and capped at 1024 chars by the label parser, then accepted for display only when finite and non-negative. `card_power_roles` is O(n) per poll with no panicking index. The missing-`sid` default parses. The power series drop from 32 to 8 per ATOM Max node. The `all_smi_gpu_info` label churn is the accepted trade-off from the brief.

Remaining items, none fixed (below the auto-fix threshold):

- LOW: the Users tab attributes a card's power only to processes on the reporting die (`src/app_state.rs:695`, `src/ui/aggregation/user.rs:222`, clamped with `.max(0.0)`, no sentinel leak). The process-to-die join on ATOM Max is already ambiguous by `npu` (#416).
- LOW: `parse_power_safe` passes `+inf` through (for example `1e400uW`) into the reporter's `power_consumption` and the totals. This predates the PR and the brief puts it out of scope. `card_power_roles` rejects it for the detail key.
- LOW: a remote `card_power_watts` such as `1e300` renders a roughly 300-char Pwr cell (`src/ui/renderers/gpu_renderer.rs:341`). The existing `power_limit_max` cell has the same problem. A plausibility cap could be a follow-up.
- LOW: a non-empty placeholder `sid` shared by every device would merge them all into one card. There is no evidence of one, and sid-only grouping is the maintainer's decision.

Validation: 154 targeted lib tests and the `local_collector` bin tests pass, and clippy on lib and tests plus fmt are clean.

inureyes:
## PR Finalization Complete

### Summary
- **Tests**: verified adequate, no additions needed. The reader's regression suite (`src/device/readers/rebellions_atom_max_tests.rs`) already covers the fix directly: 341.5 W total on 32 rows/8 cards, reorder stability, ATOM Plus unchanged, per-die memory, missing/empty `sid`, the reporting-die tie-break rule, an unparsable reporter value, cache refresh on both the cached and uncached paths, the `/metrics` render (8 power series, no sentinel, no `all_smi_npu_power_watts`, `card_power_watts` on all 32 `all_smi_gpu_info` rows), and a remote round trip. Every consumer touched by the fix (`gpu_readings::mean_power_watts`, the aggregator, `local_collector::inject_gpu_power`, the dashboard Avg. Power cell, the TUI power cell, `metrics_parser`) has its own unit test for the absent-power case.
- **Documentation**: complete, no changes needed. `API.md` and `docs/man/all-smi.1` already document the per-card grouping, the reporting-die rule, and the `card_power_watts` label. No Korean translation of `API.md`, `README.md`, or the man page exists in the tracked tree (`git ls-files | grep -i -e _ko -e ko/ -e korean` only matches per-issue `TECHNICAL_REPORTS/*.ko.md`, a different doc set). `TESTING.md`, `DEVELOPERS.md`, and `docs/LIB_mode.md` were swept too; none enumerates fixtures or per-device power detail that needs updating.
- **Lint/Format**: clean, no fixes needed. `cargo check --lib --tests`, `cargo clippy --lib --tests -- -D warnings`, and `cargo fmt --all -- --check` all pass with no output.

### Validation run
- `CARGO_BUILD_JOBS=9 RUST_TEST_THREADS=9 cargo test --lib -- device::readers::rebellions device::readers::detail_keys metrics::gpu_readings metrics::aggregator network::metrics_parser ui::renderers::gpu_renderer ui::dashboard`: 154 passed.
- `CARGO_BUILD_JOBS=9 RUST_TEST_THREADS=9 cargo test --bin all-smi view::data_collection::local_collector`: 6 passed, 1 ignored (hardware measurement harness).
- `CARGO_BUILD_JOBS=9 RUST_TEST_THREADS=9 cargo test --test user_facing_text_test`: 5 passed.
- `cargo check --lib --tests`, `cargo clippy --lib --tests -- -D warnings`, `cargo fmt --all -- --check`: all clean.
- CI on this PR: Build Check, Test Suite, launchd Service Smoke Test, systemd Service Smoke Test, Packaging Sync Check, and license/cla all pass.

No source, test, or doc changes were needed beyond what is already committed, so no new commit was made. Ready for merge.

